PR #71, opened 2017-06-29
add agent.rs Java wrapper

peacekeeper:
This adds the current `agent.rs` Java wrapper plus some other minor improvements.

Not yet happy with how the various agent callbacks (e.g. `connection_cb`, `message_cb`) are expressed here in Java, will investigate better design.


wrappers/java/src/main/java/org/hyperledger/indy/sdk/agent/Agent.java
```diff
@@ -0,0 +1,329 @@
...
+
+	private Agent() {
+
+	}
+
+	/*
+	 * STATIC METHODS
+	 */
+
+	public static Future<AgentConnectResult> agentConnect(
```
vimmerru: I see that all return types have Result wrapper. For me, it looks like just additional boilerplate. I suggest:

1. if there are no fields in Result then just return Future<Void>
2. if we return an intermediate object like AgentConnection then return it directly without Result wrapper. For example,

`public static Future<AgentConnection> connect`

3. If the result is a simple type like String then use this type directly. For example, `Result<String>`
4. And only If the result is a tuple then define a custom Result type.

wrappers/java/src/main/java/org/hyperledger/indy/sdk/agent/Agent.java
```diff
@@ -0,0 +1,329 @@
...
+				receiverDid,
+				connectionCb,
+				messageCb);
+
+		checkResult(result);
+
+		return future;
+	}
+
+	public static Future<AgentListenResult> agentListen(
```
vimmerru: > Callback connectionCb
> Callback messageCb

These callbacks are low-level c callbacks that use low-level ids, handles and error codes as params. It would be nice to hide this details from library user. I suggest using observer-like approach. We can define one interface 

```
interface ConnectionObserver {
  void onConnection(AgentConnection connection);
  void onMessage(AgentConnection connection, String message);
}
```

And ask a user to pass implementation of this observer as the latest parameter of agentConnect. This parameter will replace both callbacks.

wrappers/java/src/main/java/org/hyperledger/indy/sdk/agent/Agent.java
```diff
@@ -0,0 +1,329 @@
...
+	public static Future<AgentConnectResult> agentConnect(
+			Pool pool,
+			Wallet wallet,
+			String senderDid,
+			String receiverDid,
+			Callback messageCb) throws SovrinException {
+
+		final CompletableFuture<AgentConnectResult> future = new CompletableFuture<> ();
+
+		Callback connectionCb = new Callback() {
```
vimmerru: > connectionCb = new Callback()

We are creating a new instance of Callback for each call and perform passing of this callback to c function. JVM can't manage reference counting in C code. As result, GC will free this object on the next GC iteration and it will cause bad behavior like crashes or missed callback calls. 

To avoid this I suggest the following:

1. For each call create one static callback instance and `static Map<i32 /* command handle */, Future>`
2. Use global atomic counter to generate new command handles
3. When user calls our method generate new command handles and put corresponded Future to static Map
4. When libsovrin calls our callback resolve corresponded future with result

wrappers/java/src/main/java/org/hyperledger/indy/sdk/agent/Agent.java
```diff
@@ -0,0 +1,329 @@
...
+
+		return future;
+	}
+
+	public static Future<AgentAddIdentityResult> agentAddIdentity(
+			Agent.Listener listener,
+			Pool pool,
+			Wallet wallet,
+			String did,
+			Callback connectionCb,
```
vimmerru: > Callback connectionCb,
> Callback messageCb

I don't understand why agentAddIdentity and agentRemoveIdentity require these params. It looks like copy-paste issue.

wrappers/java/src/main/java/org/hyperledger/indy/sdk/agent/Agent.java
vimmerru: You can check https://github.com/java-native-access/jna/blob/master/www/CallbacksAndClosures.md

"If your callback needs to live beyond the method invocation where it is used, make sure you keep a reference to it or the native code will call back to an empty stub after the callback object is garbage collected."

wrappers/java/src/main/java/org/hyperledger/indy/sdk/agent/Agent.java
peacekeeper: agree, fixed in a7b0eaaf4d6d9fe62f3dd3e503d61e75498d7d6f

wrappers/java/src/main/java/org/hyperledger/indy/sdk/agent/Agent.java
peacekeeper: agree, fixed in 2c8fc025ae72f0b1d1737f13dee81710b84d0ba5

peacekeeper: agree, fixed in 14326b46f870c2ce5d84ce6b480b1a4249e596d0

wrappers/java/src/main/java/org/hyperledger/indy/sdk/agent/Agent.java
```diff
@@ -0,0 +1,329 @@
...
+	/*
+	 * STATIC METHODS
+	 */
+
+	public static Future<AgentConnectResult> agentConnect(
+			Pool pool,
+			Wallet wallet,
+			String senderDid,
+			String receiverDid,
+			Callback messageCb) throws SovrinException {
```
peacekeeper: see dcbd87f19300dfbf1795c4ee2c5762caf6914449 for initial commit to use observer pattern

wrappers/java/src/main/java/org/hyperledger/indy/sdk/agent/Agent.java
peacekeeper: see dcbd87f19300dfbf1795c4ee2c5762caf6914449 for initial commit to use observer pattern

wrappers/java/src/main/java/org/hyperledger/indy/sdk/agent/Agent.java
peacekeeper: this is now WIP in 29813c7874f18551e8b174d424a114071259da23

PR #98, opened 2017-07-13
Deleted wallet_seq_no Api call from Rust and Java wrapper.

Artemkaaas:


vimmerru:
What about removing of this call from iOS wrapper?

Artemkaaas:
I can not check for iOS what after deleting all tests work.

Artemkaaas:
It is hard. I can not check for iOS what all tests work after deleting seqNo.

PR #119, opened 2017-07-19
Python Wrapper: Agent API

vimmerru:
* Agent API wallet implementation


wrappers/python/indy/agent.py
```diff
@@ -0,0 +1,308 @@
...
+
+class Event:
+    handle: int
+    error: IndexError
+
+    def is_success(self):
+        return self.error is None
+
+
+class ConnectionEvent(Event):
```
mzk-vct: Add docstring with descriptions of these classes, public methods, functions and their params

PR #127, opened 2017-07-21
Feature/java register wallet type

peacekeeper:
Java wrapper for indy_register_wallet_type()


srottem:
Just stylistic - it might be more intention revealing to rename WalletType to CustomWalletBase.

PR #128, opened 2017-07-21
New .NET wrapper for the SDK

srottem:
This pull request is to add a new .NET wrapper for the SDK.  The approach has been closely modeled on the Java wrapper, although there some differences:

- Namespaces/package names have been consolidated
- Classes that generate JSON for opening pools, wallets, etc. have been skipped
- Method naming uses CamelCase and an Async suffix in line with .NET conventions. 
- Instance methods don't mention the types they refer to (e.g. CloseAsync rather than CloseWalletAsync)
- Minor differences due to language features. 

The test suite has also been replicated although several are producing anomalous results, some possibly due to IS-228.  These will require investigation and additional eyes would be appreciated.

Note that the port numbers used in Agent tests have been changed due to conflicts with the ports used by the pool nodes in my environment, but otherwise the tests are functionally the same.


srottem:
Closed in favor of  https://github.com/hyperledger/indy-sdk/pull/147

PR #132, opened 2017-07-24
Added demo tests

Artemkaaas:


wrappers/python/tests/demo/agent.py
```diff
@@ -0,0 +1,94 @@
...
+import logging
+import json
+
+from tests.utils.pool import create_genesis_txn_file
+
+logging.basicConfig(level=logging.DEBUG)
+
+
+@pytest.fixture(autouse=True)
+def before_after_each():
```
vimmerru: Use config.py and shared fixtures

PR #133, opened 2017-07-24
Python Wrapper: Agent API tests

vimmerru:
* New tests for agent 2 agent API
* Refactoring of tests with massive fixtures usage


wrappers/python/tests/agent/conftest.py
```diff
@@ -0,0 +1,84 @@
...
+
+@pytest.fixture
+async def endpoint():
+    return "127.0.0.1:9701"
+
+
+@pytest.fixture
+async def wallet_with_identity(wallet_handle, endpoint):
+    did, verkey, pk = await signus.create_and_store_my_did(wallet_handle, "{}")
+    await signus.store_their_did(wallet_handle,
```
ashcherbakov: I think it would be better to have two separate wallets and endpoints

wrappers/python/tests/agent/test_agent_send.py
```diff
@@ -0,0 +1,25 @@
+import json
+
+import pytest
+
+from indy import signus, ledger, agent
+from tests.utils import wallet
+
+
+@pytest.mark.asyncio
+async def test_agent_send_works_for_all_data_in_wallet_present(connection):
```
ashcherbakov: Do we have a test to check a pairwise connection, that is communication in both directions (with replies)?

PR #145, opened 2017-07-28
Added new Error - PoolLedgerAlreadyExistsError.

Artemkaaas:


src/services/pool/mod.rs
```diff
@@ -578,7 +578,8 @@ impl PoolService {
         };
 
         if path.as_path().exists() {
-            return Err(PoolError::NotCreated("Already created".to_string()));
+            return Err(PoolError::AlreadyExists(
```
FarooqKhan: At this point I am not sure from the error message what path the library is seeing verses what path I am passing in the config.

It would help to have the path that the rust library is seeing in the error message

vimmerru: > It would help to have the path that the rust library is seeing in the error message

It is internal information about libindy persistent storage. All you need to know as the developer is the pool config name, but you just passed this name to this call. 

Actually, it is a good idea to add pool config name to this error message.

include/indy_mod.h
```diff
@@ -93,6 +93,9 @@ typedef enum
     // Attempt to send transaction without the necessary privileges
     LedgerSecurityError = 305,
 
+    // Attempt to create pool with name used for another exists pool
+    PoolLedgerAlreadyExistsError = 306,
```
vimmerru: I am not sure that it is the correct terminology. Actually we are creating pool ledger config, but not just pool ledger. I suggest to change it in all places.

src/errors/pool.rs
```diff
@@ -24,6 +25,7 @@ impl fmt::Display for PoolError {
             PoolError::InvalidHandle(ref description) => write!(f, "Invalid Handle: {}", description),
             PoolError::Rejected(ref description) => write!(f, "Rejected by pool: {}", description),
             PoolError::Terminate => write!(f, "Pool work terminated"),
+            PoolError::AlreadyExists(ref description) => write!(f, "Pool already exists {}", description),
```
vimmerru: > Pool already exists

should be changed to

> Pool ledger config already exists

Also, we need pool config name in the description

src/services/pool/mod.rs
```diff
@@ -578,7 +578,8 @@ impl PoolService {
...
+                "Pool config file with same name already exists".to_string()));
```
vimmerru: Need to correct English in some messages. Discuss this with me.

wrappers/python/indy/error.py
```diff
@@ -92,6 +92,9 @@ class ErrorCode(IntEnum):
     # Attempt to send transaction without the necessary privileges
     LedgerSecurityError = 305,
 
+    # Attempt to create pool with name used for another exists pool
+    PoolLedgerAlreadyExistsError = 306,
```
vimmerru: May be it is better to rename this to PoolLedgerConfigAlreadyExistsError. Let's discuss offline.

src/services/pool/mod.rs
FarooqKhan: I meant the path of the ledger transaction file

PR #152, opened 2017-08-01
Updated error list for iOS. Fixed getTxnRequest tests.

anastasia-tarasova:


tests/utils/pool.rs
```diff
@@ -0,0 +1,202 @@
+extern crate time;
```
jovfer: This file seems like garbage after merge, please re-check and clean-up.

libindy/build-libindy-core-ios.sh
```diff
@@ -22,6 +22,7 @@ fi
 echo "Packing...\n\n"
 
 cp include/*.h $WORK_DIR
+cd ../
```
jovfer: Please re-check the next line: should be swapped this current?

anastasia-tarasova: this line is deleted

tests/utils/pool.rs
anastasia-tarasova: This file is deleted.

PR #157, opened 2017-08-03
Python Wrapper: Windows support and tests refactoring

vimmerru:
* Refactoring of python tests to follow libindy tests changes related to windows support
* Solved windows tests execution problems introduced by merge issues


wrappers/python/tests/pool/test_create_pool_ledger_config.py
```diff
@@ -1,31 +1,45 @@
...
+from tests.utils import pool as pool_utils, storage as storage_utils
 from indy.error import ErrorCode, IndyError
 
 import pytest
 
 
 @pytest.mark.asyncio
 async def test_create_pool_ledger_config_works(cleanup_storage):
-    await pool.create_pool_ledger_config("pool_1")
+    await pool.create_pool_ledger_config(
```
ashcherbakov: Can we check that genesis txns are really applied/processed?

wrappers/python/tests/pool/test_delete_pool_ledger_config.py
```diff
@@ -1,21 +1,29 @@
...
 async def test_delete_pool_ledger_config_works(cleanup_storage):
-    await pool.create_pool_ledger_config("pool_1")
-    await delete_pool_ledger_config("pool_1")
+    await pool.create_pool_ledger_config(
+        "pool_1",
+        json.dumps({
+            "genesis_txn": str(pool_utils.create_genesis_txn_file_for_test_pool("pool_1"))
+        }))
+
+    await pool.delete_pool_ledger_config("pool_1")
```
ashcherbakov: Can we check that it's really deleted?

wrappers/python/tests/pool/test_open_pool_ledger.py
```diff
@@ -1,58 +1,81 @@
...
 from indy.error import ErrorCode, IndyError
 
 import pytest
 
 
 @pytest.mark.asyncio
 async def test_open_pool_ledger_works(cleanup_storage):
-    await pool.create_pool_ledger_config("pool_1")
-    pool_handle = await open_pool_ledger("pool_1", "")
+    await pool.create_pool_ledger_config(
```
ashcherbakov: Should we move this code into a fixture as it's used in multiple tests?

wrappers/python/tests/pool/test_open_pool_ledger.py
```diff
@@ -1,58 +1,81 @@
...
-    ]
-    await pool.create_pool_ledger_config("pool_1", nodes)
-    pool_handle = await open_pool_ledger("pool_1", None)
+    await pool.create_pool_ledger_config(
+        "pool_1",
+        json.dumps({
+            "genesis_txn": str(pool_utils.create_genesis_txn_file_for_test_pool("pool_1", 2))
+        }))
+
+    pool_handle = await pool.open_pool_ledger("pool_1", None)
```
ashcherbakov: How do we check that it's really opened? is `assert pool_handle is not None` enough?

wrappers/python/tests/pool/test_create_pool_ledger_config.py
vimmerru: Looks like we need to start a discussion about test procedure design.

My vision is the following. This method provides only ok/exception for checking as a public interface. Integration tests should work on this public interface level only.

Sure we can perform additional checks:

1. Try to create config twice and get corresponded exception, but it is a dedicated test case. In rust we have this test case, but it is in the medium cases bunch. So we didn't port it to python.
2. We can try to open pool with this configuration, but it will actually test case for open_pool method. We have this test case in a dedicated file.
3. We can check that files on file system are actually created, but it is testing of non-documented side effects. It can be good unit tests for libindy, but we don't need this test for python wrapper.

wrappers/python/tests/pool/test_delete_pool_ledger_config.py
vimmerru: See my comment about create_pool_ledger_config. We have exactly same situation here.

wrappers/python/tests/pool/test_open_pool_ledger.py
vimmerru: See my comment about create_pool_ledger_config. We have an exactly same situation here.

wrappers/python/tests/anoncreds/test_prover_create_proof.py
```diff
@@ -42,29 +44,33 @@
...
     }
 
     claim_defs = {
-        claim_for_attr: json.loads(init_common_wallet[1])
+        claim_for_attr: json.loads(claim_def_json)
     }
 
-    await prover_create_proof(init_common_wallet[0], json.dumps(proof_req), json.dumps(requested_claims),
-                              json.dumps(schemas), anoncreds.COMMON_MASTER_SECRET_NAME,
+    await prover_create_proof(wallet_handle, json.dumps(proof_req), json.dumps(requested_claims),
```
ashcherbakov: Should we check that returned proof satisfies the expected result (as defined by the API)?

wrappers/python/tests/anoncreds/test_prover_get_claim_offers.py
```diff
@@ -1,57 +1,74 @@
...
 
+# noinspection PyUnusedLocal
 @pytest.mark.asyncio
-async def test_prover_get_claim_offers_works_for_empty_filter(init_common_wallet):
-    claim_offers = json.loads(await prover_get_claim_offers(init_common_wallet[0], "{}"))
+async def test_prover_get_claim_offers_works_for_empty_filter(wallet_handle, prepopulated_wallet):
+    claim_offers = json.loads(
+        await prover_get_claim_offers(wallet_handle, "{}"))
+
     assert len(claim_offers) == 3
```
ashcherbakov: Can we check that the returned claim offers match the expected result (as defined by API)?

wrappers/python/tests/anoncreds/test_prover_get_claims.py
```diff
@@ -1,40 +1,54 @@
...
 
+# noinspection PyUnusedLocal
 @pytest.mark.asyncio
-async def test_prover_get_claims_works_for_empty_filter(init_common_wallet):
-    claims = json.loads(await prover_get_claims(init_common_wallet[0], "{}"))
+async def test_prover_get_claims_works_for_empty_filter(wallet_handle, prepopulated_wallet):
+    claims = json.loads(
+        await prover_get_claims(wallet_handle, "{}"))
+
     assert len(claims) == 1
```
ashcherbakov: Can we check that the returned claims match the expected result (as defined by API)?

PR #158, opened 2017-08-03
Disabled Java tests in jenkins pipeline

Artemkaaas:


vimmerru:
Looks like we fixed the problem with tests freezes and don't need this workaround now.

PR #160, opened 2017-08-04
[WIP] Feature/wallet type marshalling

srottem:
Modified WalletType to abstract marshalling of values to unmanaged memory away from implementers.  As part of this a CustomWallet interface has been introduced and instances WalletType subclasses can be passed to registerWalletType().

The in memory custom wallet implementation has been moved to the test code as it isn't part of the API and it has been restructured to use the new WalletType and CustomWallet.  

A temporary test that exercises custom WalletType implementations to ensure marshalling is working correctly has been added to the RegisterWalletTypeTest unit, however this test is not passing correctly at present due to an issue with the JSON data returned which I have not yet been able to diagnose.


srottem:
It looks like the failing build is because the CloseWalletTest.testCloseWalletWorksForPlugged is actually doing what it's supposed to - it looks like it's uncovered an SDK issue.  Bug reported: https://jira.hyperledger.org/browse/INDY-762

jovfer:
https://jira.hyperledger.org/browse/IS-310 has been fixed

PR #169, opened 2017-08-09
Java Wrapper: Platform-independent way to init libindy

peacekeeper:
I think 9ad2cb6eff85589666664e6717456cadca7ffb42 added Windows support in `wrappers/java/src/test/java/org/hyperledger/indy/sdk/utils/InitHelper.java` but broke Linux. I propose this change to add a more platform-independent way to load the library.

### DIFF
--- a/wrappers/java/src/test/java/org/hyperledger/indy/sdk/utils/InitHelper.java
+++ b/wrappers/java/src/test/java/org/hyperledger/indy/sdk/utils/InitHelper.java
@@ -3,12 +3,10 @@ package org.hyperledger.indy.sdk.utils;
 
 import org.hyperledger.indy.sdk.LibIndy;
 
-import java.io.File;
-
 public class InitHelper {
 	public static void init() {
 
-		if (!LibIndy.isInitialized()) LibIndy.init(new File("./lib/libindy"));
+		if (!LibIndy.isInitialized()) LibIndy.init("./lib/");
 
 	}
 }
--- a/wrappers/java/src/main/java/org/hyperledger/indy/sdk/LibIndy.java
+++ b/wrappers/java/src/main/java/org/hyperledger/indy/sdk/LibIndy.java
@@ -5,6 +5,7 @@ import java.io.File;
 import com.sun.jna.Callback;
 import com.sun.jna.Library;
 import com.sun.jna.Native;
+import com.sun.jna.NativeLibrary;
 
 public abstract class LibIndy {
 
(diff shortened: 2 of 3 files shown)


wrappers/java/src/test/java/org/hyperledger/indy/sdk/utils/InitHelper.java
```diff
@@ -3,12 +3,10 @@
 
 import org.hyperledger.indy.sdk.LibIndy;
 
-import java.io.File;
-
 public class InitHelper {
 	public static void init() {
 
-		if (!LibIndy.isInitialized()) LibIndy.init(new File("./lib/libindy"));
+		if (!LibIndy.isInitialized()) LibIndy.init("./lib/");
```
vimmerru: I am not sure that looking for libindy inside jar is a good idea at all. libindy should be installed with corresponded installation package and loaded from system LD path.

peacekeeper: It doesn't look for libindy inside the jar.

Calling `LibIndy.init()` will load libindy from system LD path. <-- default
Calling `LibIndy.init(new File("./lib/libindy.so"))` will load libindy from a given file path. <-- not platform independent.
Calling `LibIndy.init("./lib/")` will load libindy from a given directory path. <-- useful for testing, this is new.

vimmerru: Personally, i prefer lazy loading without providing explicit LibIndy.init() API at all. If a user (or tester) needs to load the library from the specific place he can just modify LD path for the current process that can be easily done with environment variables on the most platforms.

What do you think about this?

peacekeeper: I agree. Actually, if libindy is in the LD path, it is already loaded automatically today, and the user doesn't have to manually call the LibIndy.init() API at all.

But for example the current unit tests (see [InitHelper.java](https://github.com/hyperledger/indy-sdk/blob/master/wrappers/java/src/test/java/org/hyperledger/indy/sdk/utils/InitHelper.java)) are loading libindy from a local directory as follows:

`LibIndy.init(new File("./lib/libindy"));
`

I agree it's preferable to load libindy from the LD path, but just in case someone doesn't want to do it, this PR provides a better, platform-independent way of doing it.

PR #185, opened 2017-08-13
Feature/dotnet agent events restructure

srottem:
Added WaitForConnection() to the Agent Listener and WaitForMessage() to the Agent Connection.  These methods return async tasks that resolve to events.


wrappers/dotnet/indy-sdk-dotnet-test/Wrapper/AgentTests/AgentSendTest.cs
```diff
@@ -1,69 +1,45 @@
...
+            var connectionEvent = await listener.WaitForConnection();
+            var serverConnection = connectionEvent.Connection;
 
             var clientToServerMessage = "msg_from_client";
             var serverToClientMessage = "msg_from_server";
 
-            clientToServerConnection.SendAsync(clientToServerMessage).Wait();
+            await clientConnection.SendAsync(clientToServerMessage);
+
+            var serverMessageEvent = await serverConnection.WaitForMessage();
```
vimmerru: You created 2 dedicated methods WaitForConnection and WaitForMessage. The problem with this approach is that during awaiting of Connection events you can't receive Message events and vice versa. I see 2 solutions:

1. It can be possible to combine 2 Tasks in the way that awaiting will be finished on the finishing of first task. So we need to check .Net asyncio docs how and provide an example if it is possible.
2. Create a single waitForEvent method that will return generic event. Python wrapper follows this way.

srottem:
.NET does have Task.WaitForAll() and Task.WaitForAny() as well as
Task.WhenAll() and Task.WhenAny() which take an array of async tasks to
wait on.  Is that what you're looking for?

wrappers/dotnet/indy-sdk-dotnet-test/Wrapper/AgentTests/AgentSendTest.cs
srottem: .NET does have Task.WaitForAll() and Task.WaitForAny() as well as
Task.WhenAll() and Task.WhenAny() which take an array of async tasks to
wait on.  Is that what you're looking for?

vimmerru: Yes, I suggest trying rewriting of this tests with WaitForAny usage and check how usable it will be.

PR #186, opened 2017-08-14
Sign/Verify work with array of byte

Artemkaaas:


libindy/src/utils/byte_array.rs
```diff
@@ -0,0 +1,15 @@
...
+        }
+
+        let $x =  unsafe { slice::from_raw_parts($x, $l as usize) };
+        let $x = $x.to_vec();
+    }
+}
+
+pub fn vec_to_pointer(v: Vec<u8>) -> (*const u8, u32) {
+    let len = v.len() as u32;
+    (v.as_ptr() as *const u8, len)
```
jovfer: Invalid pointer: vector will be freed at the end of this function

libindy/tests/demo.rs
```diff
@@ -1049,16 +1049,20 @@ fn signus_demo_works() {
             "dest":"4efZu2SXufS556yss7W5k6Po37jt4371RM4whbPKBKdB"
         }
     }"#;
+
+    let message_ptr = message.as_bytes().as_ptr() as *const u8;
```
jovfer: Why not just message.as_ptr()?

PR #204, opened 2017-08-21
[WIP] Fix IS-287 Failure to sync with validator pool

jovfer:
Update GenTransaction struct to be compatible with pool JSON format. TODO: use AST instead of fixed structure.


jovfer:
this hotfix can be closed: generic solution in #223

PR #215, opened 2017-08-23
Feature/ios updating tests

anastasia-tarasova:


libindy/build-libindy-core-ios.sh
```diff
@@ -33,16 +33,16 @@ cd $CUR_DIR
 
 echo "Uploading...."
```
jovfer: Logging should be commented too.

wrappers/ios/libindy-pod/libindy-demoTests/Test Utils/PoolUtils.m
```diff
@@ -35,10 +35,12 @@ + (PoolUtils *)sharedInstance
 
 + (NSString *) nodeIp
```
jovfer: Remove redundant function.

libindy/build-libindy-core-ios.sh
anastasia-tarasova: Done

PR #245, opened 2017-09-03
Feature/dotnet documentation

srottem:
Many files have changed in this branch but no actual functionality has been changed - all modifications have been about tidying the code for documentation.  The following changes have been made:

- Reorganized code and namespaces to improve navigability of code and documentation
- Changed visibility of several members to hide them from API users and documentation
- Significantly expanded XML code docs to improve clarity of method usage
- Added a Sandcastle project to auto generate HTML documentation from XML code docs

Please see the [sample of the code docs generated](http://www.symbiotic-development.com/indy-sdk-dotnet-help/) with Sandcastle from this pull request are available temporarily online at  for review.


wrappers/dotnet/indy-sdk-dotnet-test/AgentTests/AgentAddIdentityTest.cs
```diff
@@ -1,8 +1,9 @@
-﻿using Indy.Sdk.Dotnet.Wrapper;
+﻿using Hyperledger.Indy.Sdk.AgentApi;
+using Hyperledger.Indy.Sdk.SignUsApi;
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System.Threading.Tasks;
 
-namespace Indy.Sdk.Dotnet.Test.Wrapper.AgentTests
+namespace Hyperledger.Indy.Sdk.Test.AgentTests
```
jovfer: Can we remove `.Sdk.` part of packages names? In our current vision SDK is complex of wrappers, lib, docs, samples etc. So, in wrapper it seems like redundant.

wrappers/dotnet/indy-sdk-dotnet-test/AgentTests/AgentAddIdentityTest.cs
```diff
@@ -1,8 +1,9 @@
-﻿using Indy.Sdk.Dotnet.Wrapper;
+﻿using Hyperledger.Indy.Sdk.AgentApi;
```
jovfer: We discuss with @vimmerru and our vision is `Api` suffix is redundant (here and below: Signus, AnonCreds ...)

wrappers/dotnet/indy-sdk-dotnet-test/AgentTests/AgentCloseConnectionTest.cs
```diff
@@ -12,12 +13,12 @@ public class AgentCloseConnectionTest : AgentIntegrationTestBase
         {
             var endpoint = "127.0.0.1:9603";
 
-            var myDid = await Signus.CreateAndStoreMyDidAsync(_wallet, "{}");
+            var myDid = await SignUs.CreateAndStoreMyDidAsync(_wallet, "{}");
```
jovfer: Initially `us` is just a suffix for `sign`, let's change back `SignUs` to `Signus`.

wrappers/dotnet/indy-sdk-dotnet-test/AgentTests/AgentAddIdentityTest.cs
srottem: The issue with this is that it's bad form to have a namespace that conflicts with a class name so an Agent namespace and an Agent class will cause problems, hence the selection of the Api suffix.  The same isn't true of package names in Java.

wrappers/dotnet/indy-sdk-dotnet-test/AgentTests/AgentAddIdentityTest.cs
srottem: Sounds reasonable, so sure.

wrappers/dotnet/indy-sdk-dotnet-test/AgentTests/AgentCloseConnectionTest.cs
srottem: OK.  The only reason I changed it is that it's unclear what the pronunciation of "Signus" is; it could be with a hard 'g'.  Slava mentioned that the correct pronunciation is "sign us" so I change the case to reflect that.  If after reviewing this the preference is still to roll back to the original name I don't see a problem.

jovfer: > If after reviewing this the preference is still to roll back to the original name I don't see a problem.

yes, please roll back

wrappers/dotnet/indy-sdk-dotnet-test/AgentTests/AgentAddIdentityTest.cs
jovfer: OK, we haven't any another proposal for the suffix, let's use `API`

PR #270, opened 2017-09-20
Changed libindy root folder

vimmerru:
* Use .indy_client folder as root of libindy instead of .indy
* Increased timeout for anoncreds tests


libindy/src/utils/environment.rs
```diff
@@ -37,7 +37,7 @@ impl EnvironmentUtils {
 
     pub fn tmp_path() -> PathBuf {
         let mut path = env::temp_dir();
-        path.push("indy");
```
ashcherbakov: Can we have a constant, or config option for it instead of hard-coded value?

PR #274, opened 2017-09-21
Feature/dotnet granular exceptions

srottem:
Implemented separate .NET exception classes for each SDK error code to allow error cases to be handled using the normal .NET exception management features.  Also implemented guard clauses on public method parameters.

Note: the exception error messages should be reviewed for accuracy as the SDK documentation for some of the error codes is not very descriptive or is missing.

The samples project has been updated to use some of the new exception types.


jovfer:
@srottem @vimmerru 
this approach differ against Java wrapper. Is granular exceptions really useful in this case?

srottem:
I think so, yes, and the Java wrapper would benefit from this as well (I'm happy to handle implementation).  It allows developers to handle exceptions in a way that is more in line with normal error handling in those languages and the errors convey more information directly to the developer.  I've already provided a PR to add guard clauses to the Java wrapper as parameter errors from the SDK actually even convey incorrect information to the developer (e.g. param1 from a Java/.NET wrapper call does not correspond to the 1st parameter of an SDK call).

jovfer:
I agree almost full with @srottem (may be this one insignificant exception: order of parameters between a wrapper and libindy - it should be same).
@vimmerru ?

srottem:
It's not really possible for the parameter indicies to be the same; for example, when calling the SendAsync method on an AgentConnection instance the only parameter passed by the caller is the message string.  Under the covers the SendAsync method calls indy_agent_send due to its signature the message string is passed as the 3rd parameter.

srottem:
@dhh1128 Actually, we're maintaining a map of errors anyway; in the .NET and Java wrappers have an enumeration that matches the list of existing error numbers returned by the SDK.  For granular exceptions I've gone with the approach that all Indy generated errors derive from IndyException which exposes the error number reported by the SDK - I propose that we just throw the base IndyException for any error that doesn't match a known error number.

jovfer:
@srottem please resolve conflict and we will merge this PR

srottem:
Conflicts resolved and IndyException or one of its subclasses with always be thrown regardless of whether or not we have a specific mapping for an SDK error number.

PR #277, opened 2017-09-25
Corrected function for finding of four squares

Artemkaaas:


sklump:
Unit tests should include a much larger test number, such as an EPOCH date. I've tried 1506099439 often, and it hangs the machine for hours. I have a less elegant fix, attached, that returns in a few seconds, tops.

I've never written anything in Rust before, and it will show. Surely in such a pointer-heavy language, for example, there must be a memset to zeroes. But the basic algorithm runs much faster than the admittedly beautiful fix under review.

[helpers-4sq.rs.txt](https://github.com/hyperledger/indy-sdk/files/1332869/helpers-4sq.rs.txt)

PR #288, opened 2017-10-02
Keys rotation

Artemkaaas:


libindy/include/indy_signus.h
```diff
@@ -65,16 +65,37 @@ extern "C" {
     /// Wallet*
     /// Crypto*
 
-    extern indy_error_t indy_replace_keys(indy_handle_t command_handle,
-                                          indy_handle_t wallet_handle,
-                                          const char *  did,
-                                          const char *  identity_json,
+    extern indy_error_t indy_replace_keys_start(indy_handle_t command_handle,
```
jovfer: Please update comment too.

libindy/src/commands/signus.rs
```diff
@@ -205,11 +215,28 @@ impl SignusCommandExecutor {
...
     }
 
+    fn replace_keys_apply(&self,
+                          wallet_handle: i32,
+                          did: &str,
+                          cb: Box<Fn(Result<(), IndyError>) + Send>) {
+        cb(self._replace_keys_apply(wallet_handle, did));
+    }
+
+    fn _replace_keys_apply(&self,
```
jovfer: Should we delete temporary entry?
@vimmerru

vimmerru: It is better to remove from my point of view

PR #319, opened 2017-10-13
Feature/issuer revoke claim tests

srottem:
Ported high cases testing indy_issuer_revoke_claims from Rust to Java and .NET wrappers.


jovfer:
@srottem I've tried to update the PR from master and resolved some simple conflicts. But PR still failed - seems like some auto-merged files is incorrect (missed granular exception usage?)
Please let me know, if you want to see error logs from Jenkins to fix this PR.

srottem:
Yes, it looks like there is an expected exception in a test that's still
using the error code matcher rather than the hamcrest isA() matcher.  I've
fixed the problem and all tests pass.

PR #342, opened 2017-10-24
Changed Default Wallet to use SQLCipher

mikelodder7:
Default Wallet uses SQLCipher. Will need to update licensing to reflect community edition license


libindy/Cargo.toml
```diff
@@ -44,7 +44,7 @@ log = "0.3.7"
 openssl = { version = "0.9.11", optional = true }
 rand = "0.3"
 rlp = "0.2.0"
-rusqlite = "0.10.1"
+rusqlite = { git = "https://github.com/mikelodder7/rusqlcipher", features=["bundled"] }
```
jovfer: Link to repository doesn't acceptable, we need for separate Rust crates like `rusqlcipher`
@mikelodder7 could you create it, please?

mikelodder7:
Okay I made the change.

libindy/src/services/wallet/default.rs
```diff
@@ -403,4 +430,30 @@ mod tests {
 
         TestUtils::cleanup_indy_home();
     }
-}
\ No newline at end of file
+
+    #[test]
+    fn default_wallet_create_encrypted() {
```
vimmerru: It would be nice to provide some additional test cases:

1. Trying to open unencrypted database with some key
2. Trying to open encrypted database with invalid key
3. Trying to use key without Key Derivation (in format "x'2DD29CA851E7B56E4697B0E1F08507293D761A05CE4D1B628663F411A8086D99'")

libindy/src/services/wallet/default.rs
```diff
@@ -30,10 +30,18 @@ impl Default for DefaultWalletRuntimeConfig {
...
 
 #[derive(Deserialize)]
-struct DefaultWalletCredentials {}
+struct DefaultWalletCredentials {
+    key: String
+}
 
 impl<'a> JsonDecodable<'a> for DefaultWalletCredentials {}
 
+impl Default for DefaultWalletCredentials {
```
vimmerru: We need to update documentation for indy_create_wallet/indy_open_wallet functions and describe format of credentials param. We need it in:

- api/wallet.rs comments
- include/wallet.h comments
- python wrapper
- java wrapper
- iOS wrapper
- .Net wrapper too.

libindy/src/services/wallet/default.rs
vimmerru: Ideally, we need to add these tests to wrappers too, but as it backward compatible change we can merge it without wrappers update, but for Rust code, we need these tests.

jovfer:
@mikelodder7 please do not forget to fix DCO check

mikelodder7:
Going to do another one this one is getting long

jovfer:
moved to #380

PR #354, opened 2017-10-30
Feature/endpoint for did

Artemkaaas:


libindy/src/commands/signus.rs
```diff
@@ -477,14 +488,27 @@ impl SignusCommandExecutor {
...
-        // TODO: FIXME: It should support resolving of endpoint from ledget!!!
-        let endpoint = self._wallet_get_did_endpoint(wallet_handle, &did)?;
+                            pool_handle: i32,
+                            did: String,
+                            cb: Box<Fn(Result<(String, String), IndyError>) + Send>) {
+        try_cb!(self.signus_service.validate_did(&did), cb);
 
-        let res = (endpoint.ha, endpoint.verkey);
-        Ok(res)
+        check_wallet_and_pool_handles_consistency!(self.wallet_service, self.pool_service,
```
jovfer: pool_handle should be optional. This check should be performed only in case of missed local value.

jovfer: also it will simplify some tests

PR #364, opened 2017-11-06
Must use valid, tested version for indy-* deps

ckochenower:
Signed-off-by: Corin Kochenower <corin.kochenower@evernym.com>


ckochenower:
@jovfer - DCO sign-off is now in my commit message and I have set the indy-* dep versions to the versions you say have been tested.

PR #424, opened 2017-12-13
Feature/cli

vimmerru:
Basic command/params completion was implemented


cli/src/command_executor.rs
```diff
@@ -295,6 +295,99 @@ impl CommandExecutor {
...
 
+    pub fn complete(&self, line: &str, word: &str, _start: usize, _end: usize) -> Vec<(String, char)> {
+        let mut completes: Vec<(String, char)> = vec![];
+
+        let (cmd, params) = CommandExecutor::_split_first_word(line);
+
+        if let Some(ref command) = self.commands.get(cmd) {
+            // Complete command params
+
+            if "help".starts_with(word) {
```
jovfer: I suggest to pop up help compiltion from this and next blocks.
Something like
```
if !line.contains("help") {
  completes.push("help");
}
```

jovfer: Will be fixed at the next iteration.

PR #428, opened 2017-12-14
Added tests for Cli Commands

Artemkaaas:


cli/src/commands/pool.rs
```diff
@@ -27,7 +27,7 @@ pub mod create_command {
 
         let name = get_str_param("name", params).map_err(error_err!())?;
         let gen_txn_file = get_opt_str_param("gen_txn_file", params).map_err(error_err!())?
-            .unwrap_or("pool_transactions_genesis");
+            .unwrap_or("docker_pool_transactions_genesis");
```
jovfer: I suggest to extract string constant for this.
In non-test configuration it should be "pool_transactions_genesis" like in nodes
In test configuration it may be same with non-test or may be "docker_pool_transactions_genesis"

PR #432, opened 2017-12-15
Add test cases for functional test

trongnhan1312400:
Signed-off-by: Nhan Nguyen - 6187 <nhan.trong.nguyen@logigear.com>


vimmerru:
@trongnhan1312400 Could you provide a bit more context about this PR? Do we have a jira tickets? What components you test? and etc???

nghia47:
@vimmerru This PR is for the functional test of Python wrapper (libindy_1.0.1_amd64.deb). There are 31 Wallet, Pool, Signus and Ledger test cases. The tests are not driven by any unit test framework but are standalone python scripts and currently requires python 3.6. And there is no Jira card for this.

ashcherbakov:
We use Python 3.5 for indy-node and indy-sdk wrapper. Can we use Python 3.5 for these tests as well?

vimmerru:
But what is the purpose for these tests? Python Wrapper already covered by similar functional tests placed inside of indy-sdk\wrappers\python\tests folder.

Only reason i see is acceptance testing of final artifacts after packaging. For acceptance tests i don't like proposed location. For me it looks better to put acceptance tests inside of top level "acceptance" folder.

nghia47:
@ashcherbakov: These tests run well with python3.5. Will update the python requirement.

@vimmerru The existed tests are the unit-test and uses all hardcode data. They do not covered the end to end workflow but the new functional tests do. Currently, we are automating the positive tests, the next steps are negative, boundary ones. 
About acceptance tests, you are absolutely right. It should be under "acceptance" folder.

vimmerru:
> The existed tests are the unit-test and uses all hardcode data. They do not covered the end to end workflow but the new functional tests do. Currently, we are automating the positive tests, the next steps are negative, boundary ones.

In this case i don't understand why we need additional unti tests for python wrapper as we already have acceptable coverage by unit tests based on pytest and placed inside the same folder. 

You just add unnecessary complexity. We have list of unit tests that designed similar way in libindy and each wrapper. We support these tests and execute on each CI run. With this tests without CI integration we will get regressions in test code soon and just break our development process. 

@nghia47 @ashcherbakov What do you think?

nghia47:
The goals of our new tests are:
- Simulator the real scenarios that end-user performs with the system. By covering the end-to-end workflow, the tests use the runtime data to verify the product features, prevent the leaks of using hardcode data.
- Any user should be able to run the python scripts without having outside dependencies (other than the libindy library). Any user should be able to download and run any or all of the scripts to quickly asses the functionality of libindy and/or the Sovrin client (by using libindy). 

Moreover, these tests are used for regression test purpose, a type of testing process so will not affect current development process. It will help the team have a clearer quality report of the product, make sure that a change or addition hasn't broken any existing product features.

So, we should keep both of tests, the current unit tests is used for CI run and our tests will be stored in a separated folder and will be executed via an other CI task.

@ashcherbakov Please advise.

wrappers/python/Automation-Tests/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+This Python wrapper functional test currently requires python 3.6, base58.
+
+### How to run
+
+After building successfully the Indy SDK for Python, you need to run the commands below so that could run the tests:
+
+- Install base58 dependency with pip install: 
+```
+     python3.6 -m pip install base58
```
ashcherbakov: Please use Python 3.5

vimmerru:
@nghia47 @ashcherbakov 

> Simulator the real scenarios that end-user performs with the system. By covering the end-to-end workflow, the tests use the runtime data to verify the product features, prevent the leaks of using hardcode data.

With complete scenarios you perform testing not just python wrapper, but the whole multi-component system (python wrapper, indy node, libindy). So it isn't python wrapper tests. As result i don't like the current place. 

Also current tests ins't based on predefined values only. There are complete integration scenarios too and we will have duplication. 

So can we implement this tests as acceptance? It can be much more valuable from my point of view. The only difference is that for acceptance case you need to install libindy as binary artifact instead of usage of git-managed source code.

> Any user should be able to run the python scripts without having outside dependencies (other than the libindy library). Any user should be able to download and run any or all of the scripts to quickly asses the functionality of libindy and/or the Sovrin client (by using libindy).

Your tests also require near the same preparation as current tests:

1. You have to deploy pool (most tricky part)
2. You have to install dependencies: python3.6 -m pip install base58

vimmerru:
> and will be executed via an other CI task.

I am really don't like any additional CI tasks introduction as it breaks our continuous delivery process. For now all releases are performed in completely automated way. 

See our current release process and testing approach documentation:

https://github.com/hyperledger/indy-sdk/blob/master/doc/release-workflow.md
https://github.com/hyperledger/indy-sdk/blob/master/doc/test-design.md
https://github.com/hyperledger/indy-sdk/blob/master/doc/cd-pipeline.svg

Any PRs with breaking changes can be merged only after updating of this documentation and approval.

wrappers/python/Automation-Tests/README.md
nghia47: We will update this on next commit.

nghia47:
@vimmerru @ashcherbakov
I will work with Kelly Wilson <kelly.wilson@evernym.com> and Steve Lafranca <steve.lafranca@evernym.com> on this new tests to provide more information.

nghia47:
@vimmerru @ashcherbakov 
Please help to cancel this pull request. We had a new repo for our test.
Thank you for your support and review.

PR #467, opened 2018-01-10
Corrected prompt for Indy-Cli

Artemkaaas:

### DIFF
--- a/cli/src/commands/wallet.rs
+++ b/cli/src/commands/wallet.rs
@@ -117,10 +117,15 @@ pub mod open_command {
                         set_opened_wallet(ctx, Some((handle, name.to_owned())));
                         Ok(println_succ!("Wallet \"{}\" has been opened", name))
                     }
-                    Err(ErrorCode::CommonInvalidStructure) => Err(println_err!("Invalid wallet config")),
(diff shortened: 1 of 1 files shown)


cli/src/commands/wallet.rs
```diff
@@ -117,10 +117,15 @@ pub mod open_command {
                         set_opened_wallet(ctx, Some((handle, name.to_owned())));
                         Ok(println_succ!("Wallet \"{}\" has been opened", name))
                     }
-                    Err(ErrorCode::CommonInvalidStructure) => Err(println_err!("Invalid wallet config")),
-                    Err(ErrorCode::WalletAlreadyOpenedError) => Err(println_err!("Wallet \"{}\" already opened", name)),
-                    Err(ErrorCode::CommonIOError) => Err(println_err!("Wallet \"{}\" not found or unavailable", name)),
-                    Err(err) => Err(println_err!("Indy SDK error occurred {:?}", err)),
+                    Err(err) => {
+                        set_opened_wallet(ctx, None);
```
jovfer: `set_opened_wallet` should be performed in one place: after final `res` will be determined.

jovfer: UPD: one call above seems redundant, so we can leave 2 call of `set_opened_wallet`

PR #476, opened 2018-01-12
[WIP] Temporary disable iOS testing: there is unexpected issue with installing pods on jenkins agent.

jovfer:


jovfer:
iOS agent is fixed

PR #477, opened 2018-01-13
Update indy-sdk/doc/rhel-build.md

hyperfocus:
fixed ci/amazon.dockerfile link with the only amazon.dockerfile i could find.


vimmerru:
@hyperfocus 

Hi, Thanks for contribution. We have a rule that each git commit should have DCO sign-off (use -s option with each git commit that will add author and email information to commit message).

Could you fix this? To fix the latest commit in the branch you can use:

```
git commit --amend -s --no-edit
```

And after this push with force to replace this commit with a new one:

```
git push -f
```

PR #481, opened 2018-01-15
Fix invalid F in RBFT implementation

jovfer:


wrappers/java/src/test/java/org/hyperledger/indy/sdk/did/GetEndpointForDidTest.java
```diff
@@ -32,6 +32,8 @@ public void testGetEndpointForDidWorksFromLedger() throws Exception {
 		String attribRequest = Ledger.buildAttribRequest(trusteeDid, trusteeDid, null, endpoint, null).get();
 		Ledger.signAndSubmitRequest(pool, wallet, trusteeDid, attribRequest).get();
 
+		Thread.sleep(5_000);
```
vimmerru: Do we need switching to ensurePreviousRequestApplied ?

libindy/tests/utils/ledger.rs
```diff
@@ -108,6 +115,38 @@ impl LedgerUtils {
         Ok(request_result_json)
     }
 
+    pub fn extract_timestamp_from_reply(reply: &str) -> Result<u64, &'static str> {
```
vimmerru: Could we switch to ensurePreviousRequestApplied concept here too?

wrappers/java/src/test/java/org/hyperledger/indy/sdk/did/GetEndpointForDidTest.java
jovfer: Only via one more GET_ATTRIB request to ledger and still without guaranties

PR #487, opened 2018-01-17
[TODO] Increment pool version to 1.2.273.

jovfer:
TODO new pool require unique schema+version for DID, so some tests now failed and should be updated.


jovfer:
Moved to #523

PR #494, opened 2018-01-19
Corrected opening of encrypted wallet

Artemkaaas:


libindy/src/services/wallet/default.rs
```diff
@@ -280,7 +289,7 @@ impl From<rusqlcipher::Error> for WalletError {
     fn from(err: rusqlcipher::Error) -> WalletError {
         match err {
             rusqlcipher::Error::QueryReturnedNoRows => WalletError::NotFound(format!("Wallet record is not found: {}", err.description())),
-            _ => WalletError::CommonError(CommonError::InvalidState(format!("Unexpected SQLite error: {}", err.description())))
+            _ => WalletError::SecurityError(format!("Unexpected SQLite error: {}", err.description()))
```
jovfer: Can we stay InvalidState error for some unexpected errors and catch specific error from rusqlciphter::Error as Security error?

cli/src/libindy/mod.rs
```diff
@@ -83,6 +83,9 @@ pub enum ErrorCode
     // Trying to open wallet that was opened already
     WalletAlreadyOpenedError = 206,
 
+    // Attempt to open encrypted wallet with invalid credentials
+    WalletSecurityError = 207,
```
jovfer: @vimmerru `rusqlcipher::ErrorCode::NotADatabase` is returned by `rusqlcipher` in case of incorrect key. May be we should introduce slightly different error code, like `WalletFailAccessToDatabase` ?

vimmerru: For this case we can't authenticate. I more like WalletSecurityError approach or WalletInvalidCredentials.

jovfer: But `rusqlcipher` doesn't provide ability to distinct security error and invalid database file

jovfer: @Artemkaaas after discussion we suggest to use `WalletAccessFailed`

PR #501, opened 2018-01-23
[TODO] Try run cli on macos

Artemkaaas:
Signed-off-by: Artem Ivanov <artem.ivanov@dsr-company.com>


vimmerru:
Seems like outdated. I suggest to close for now,

PR #511, opened 2018-01-31
Align comments with recent data structure updates.

sklump:
Signed-off-by: Stephen Klump <srklump@becker-carroll.com>


sklump:
ah ffs

PR #512, opened 2018-01-31
Align comments with recent data structure updates.

sklump:
Signed-off-by: Stephen Klump <stephen.klump@becker-carroll.com>


sklump:
This is absurd. If the bot knows who I am and knows what to expect, I shouldn't have to work for it - it should work for me.

PR #519, opened 2018-02-02
Rotate-key available on read-only pool

Artemkaaas:
Corrected verifier attributes checking

### DIFF
--- a/cli/src/utils/table.rs
+++ b/cli/src/utils/table.rs
@@ -6,7 +6,11 @@ use prettytable::Table;
 use prettytable::row::Row;
 use prettytable::cell::Cell;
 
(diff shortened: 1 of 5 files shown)


cli/src/utils/table.rs
```diff
@@ -6,7 +6,11 @@ use prettytable::Table;
 use prettytable::row::Row;
 use prettytable::cell::Cell;
 
-pub fn print_list_table(rows: &Vec<serde_json::Value>, headers: &[(&str, &str)]) {
+pub fn print_list_table(rows: &Vec<serde_json::Value>, headers: &[(&str, &str)], empty_msg: &str) {
+    if rows.is_empty() {
+        println_succ!("{}", empty_msg);
```
jovfer: seems like `return` is missed

PR #527, opened 2018-02-08
Fixed IS-545 and IS-546

Artemkaaas:


cli/src/commands/did.rs
```diff
@@ -96,10 +96,11 @@ pub mod import_command {
...
         File format:
         {
-            \"version\": 1
-            \"dids\": [
-                { \"did\": \"did1\", \"seed\": \"UTF-8 or base64 seed string\" },
-            ]
+            \"version\": 1,
+            \"dids\": [{
+                \"did\": \"AAAAAAAAAAAAAAAAAAAAAA\",
+                \"seed\": \"00000000000000000000000000000000\"
```
jovfer: `UTF-8 or base64 seed string` is important comment about format of seed and should not be removed. If we want to have correct json, may be it's better to move this below?

PR #538, opened 2018-02-16
Hotfix for CLI pool_upgrade_works test.

jovfer:


cli/src/commands/ledger.rs
```diff
@@ -1791,6 +1791,7 @@ pub mod tests {
                 params.insert("force", "true".to_string()); // because node_works test added fifth Node
                 cmd.execute(&ctx, &params).unwrap();
             }
+            ::std::thread::sleep(::std::time::Duration::from_secs(1));
```
vimmerru: Could you add comment with motivation for this line?

PR #550, opened 2018-02-26
[WIP] Revocation

jovfer:


libindy/src/api/anoncreds.rs
```diff
@@ -525,19 +597,22 @@ pub extern fn indy_prover_create_and_store_claim_req(command_handle: i32,
 pub extern fn indy_prover_store_claim(command_handle: i32,
                                       wallet_handle: i32,
                                       claims_json: *const c_char,
-                                      rev_reg_json: *const c_char,
+                                      rev_reg_def_json: *const c_char,
+                                      rev_reg_entry_json: *const c_char,
```
vimmerru: What does rev_reg_entry_json mean?

libindy/src/api/anoncreds.rs
```diff
@@ -926,3 +1010,118 @@ pub extern fn indy_verifier_verify_proof(command_handle: i32,
...
+            })
+        ))));
+
+    result_to_err_code!(result)
+}
+
+#[no_mangle]
+pub extern fn indy_store_witness(command_handle: i32,
+                                 wallet_handle: i32,
+                                 id: *const c_char,
```
vimmerru: What is id here?

libindy/src/api/blob_storage.rs
```diff
@@ -0,0 +1,38 @@
...
+use api::ErrorCode;
+use errors::ToErrorCode;
+use commands::{Command, CommandExecutor};
+use commands::blob_storage::BlobStorageCommand;
+use utils::cstring::CStringUtils;
+
+use self::libc::c_char;
+
+#[no_mangle]
+pub extern fn indy_open_reader(command_handle: i32,
```
vimmerru: May be rename to indy_open_blob_reader ???

libindy/src/commands/anoncreds/tails.rs
```diff
@@ -0,0 +1,64 @@
...
+use self::indy_crypto::cl::{Tail, RevocationTailsAccessor, RevocationTailsGenerator};
+use self::indy_crypto::errors::IndyCryptoError;
+use self::digest::Input;
+
+use std::rc::Rc;
+
+const TAILS_BLOB_TAG_SZ: usize = 2;
+const TAIL_SIZE: usize = Tail::BYTES_REPR_SIZE;
+
+pub struct SDKTailsAccessor {
```
vimmerru: May be rename to DefaultTailsAcessor?

libindy/src/api/anoncreds.rs
```diff
@@ -83,46 +85,54 @@ pub extern fn indy_issuer_create_and_store_claim_def(command_handle: i32,
 #[no_mangle]
 pub extern fn indy_issuer_create_and_store_revoc_reg(command_handle: i32,
                                                      wallet_handle: i32,
+                                                     tails_writer_config_handle: i32,
```
vimmerru: I can't find public API to find this handle. Could you point me?

libindy/src/api/anoncreds.rs
Artemkaaas: It is transaction that contains accumulator value

libindy/src/api/anoncreds.rs
Artemkaaas: Witness and Claim must be associated. We decided to pass the same id as a parameter in store_witness and store_credential

libindy/src/api/anoncreds.rs
Artemkaaas: I have already changed it.
Now it look like: 
pub extern fn indy_issuer_create_and_store_revoc_reg(command_handle: i32,
                                                                                         wallet_handle: i32,
                                                                                         tails_writer_type: *const c_char,
                                                                                         tails_writer_config: *const c_char,
                                                                                         issuer_did: *const c_char,
                                                                                         schema_json: *const c_char,
                                                                                         max_claim_num: u32,
                                                                                         issuance_by_default: bool,

libindy/src/api/blob_storage.rs
jovfer: or `indy_blob_storage_open_reader` ?

libindy/src/api/anoncreds.rs
vimmerru: But in theory we can have multiple witnesses for one claim for different timestamps.

PR #556, opened 2018-02-28
Various fixes in getting started guide and docstrings [skip ci]

skhoroshavin:
Signed-off-by: Sergey Khoroshavin <sergey.khoroshavin@dsr-corporation.com>


skhoroshavin:
Concerning commit https://github.com/hyperledger/indy-sdk/pull/556/commits/7c9f5b870050032630a4e1e1f4cd79af56b9f5d6: after looking through the rest of notebook I'm not sure if this change is valid - it might just be better to rename few _verkey variables to _keys. The main reason for this change is that I felt it misleading to have the very same data (although from different sources) in variables those naming is quite different semantically. Feedback is very welcome.

Also it might be better to refactor onboarding function into different parts to separate actors better, but this is out of scope of this PR, I'm going to open an issue to discuss this.

PR #558, opened 2018-02-28
Added default wallet documentation

mikelodder7:


doc/indy-sdk-defaultwallet.md
```diff
@@ -0,0 +1,86 @@
+# Indy-SDK Default Wallet Implementation
```
vimmerru: Can we just rename it to "default-wallet.md"? ind-sdk prefix seems excess.

PR #564, opened 2018-03-06
[WIP] Design for wallet with pluggable storage approach

vimmerru:
In current state libindy allows to plug different wallet implementations. Plugged wallet now handles both security
and storage layers. This design proposes to restrict our plugged interface by handling only storage layer.
All encryption will be performed in libindy. It will simplify plugged wallets and warranty good security level
for 3d party wallets implementations.

Also proposals enhances our API to efficient and flexible search with paging support.

### DIFF
--- a/doc/wallet-storage.md
+++ b/doc/wallet-storage.md
@@ -1,0 +1,206 @@
+# Wallet Storage Design
+
+In current state libindy allows to plug different wallet implementations. Plugged wallet now handles both security
+and storage layers. This design proposes to restrict our plugged interface by handling only storage layer.
+All encryption will be performed in libindy. It will simplify plugged wallets and warranty good security level
+for 3d party wallets implementations.
+
+Also proposals enhances our API to efficient and flexible search with paging support.
+
+## Goals and ideas
+
+* Simplify plugged wallets and warranty security level for 3d party wallets implementation
+* Perform all encryption on libindy level:
+  * Entity ids will be always encrypted
+  * Entity values will be always encrypted
+  * Entity metadata will be always encrypted
+  * Tag names will be always encrypted
+  * Tag values will be **optionally** encrypted. If user wants to perform some complex searches it can be possible
+    to include some additional un-encrypted tags
+* Allow plugging of different storages
+* Allow efficient and flexible search for entities
+* Allow pagination
+* Native OpenSSL style object-oriented C interface
+* Try to avoid unnecessary json and re-allocation
+
+## Wallet Storage interface
+
+To plug wallet storage use should implement and register the following callbacks:
+
+### Wallet Storage Type API
+
+```Rust
+/// Create the wallet storage (For example, database creation)
+///
+/// name: wallet storage name (the same as wallet name)
+/// config: wallet storage config (For example, database config)
+/// credentials: wallet storage credentials (For example, database credentials)
+extern fn create_wallet_storage(name: *const c_char,
+                                config: *const c_char,
+                                credentials: *const c_char) -> ErrorCode
+```
+
+```Rust
+/// Delete the wallet storage (For example, database deletion)
+///
+/// name: wallet storage name (the same as wallet name)
+/// config: wallet storage config (For example, database config)
+/// credentials: wallet storage credentials (For example, database credentials)
+extern fn delete_wallet_storage(name: *const c_char,
+                                config: *const c_char,
+                                credentials: *const c_char) -> ErrorCode
+```
+
+```Rust
+/// Open the wallet storage (For example, opening database connection)
+///
+/// name: wallet storage name (the same as wallet name)
+/// config: wallet storage config (For example, database config)
+/// runtime_config: wallet storage runtime config (For example, connection config)
+/// credentials: wallet storage credentials (For example, database credentials)
+/// storage_handle_p: pointer to store opened storage handle
+extern fn open_wallet_storage(name: *const c_char,
+                              config: *const c_char,
+                              runtime_config: *const c_char,
+                              credentials: *const c_char,
+                              storage_handle_p: *mut u32) -> ErrorCode
+```
+
+### Wallet Storage API
+
+```Rust
+/// Create a new entity in the wallet storage
+///
+/// storage_handle: opened storage handle (See open_wallet_storage)
+/// type_: allows to separate different entity types collections
+/// id: the id of entity
+/// value: the value of entity
+/// metadata: (optional) the entity metadata
+/// tags: (optional) the entity tags used for search as json: 
+///  {
+///    "tagName1": "tag value 1", // string value
+///    "tagName2": 123, // numeric value
+///  }
+extern fn wallet_storage_create_wallet_entity(storage_handle: u32,
+                                              type_: *const c_char,
+                                              id: *const c_char,
+                                              value: *const c_char,
+                                              metadata: *const c_char,
+                                              tags: *const c_char) -> ErrorCode
+```
+
+```Rust
+/// Update an existing entity in the wallet storage
+///
+/// storage_handle: opened storage handle (See open_wallet_storage)
+/// type_: allows to separate different entity types collections
+/// id: the id of entity
+/// value: (optional) the new value of entity
+/// metadata: (optional) the new entity metadata
+/// tags: (optional) the new entity tags (see wallet_storage_create_wallet_entity)
+extern fn wallet_storage_update_wallet_entity(storage_handle: u32,
+                                              type_: *const c_char,
+                                              id: *const c_char,
+                                              value: *const c_char,
+                                              metadata: *const c_char,
+                                              tags: *const c_char) -> ErrorCode
+```
+
+```Rust
+/// Delete an existing entity in the wallet storage
+///
+/// storage_handle: opened storage handle (See open_wallet_storage)
+/// type_: allows to separate different entity types collections
+/// id: the id of entity
+extern fn wallet_storage_delete_wallet_entity(storage_handle: u32,
+                                              type_: *const c_char,
+                                              id: *const c_char) -> ErrorCode
+```
+
+```Rust
+/// Get an wallet entity by id
+///
+/// storage_handle: opened storage handle (See open_wallet_storage)
+/// type_: allows to separate different entity types collections
+/// id: the id of entity
+/// options_json: //TODO: FIXME: Think about replacing by bitmaks
+///  {
+///    retrieveValue: (optional, true by default) Retrieve entity value,
+///    retrieveMetadata: (optional, true by default) Retrieve entity metadata
+///  }
+/// entity_handle_p: pointer to store retrieved entity handle
+extern fn wallet_storage_get_wallet_entity(storage_handle: u32,
+                                           type_: *const c_char,
+                                           id: *const c_char,
+                                           options_json: *const c_char,
+                                           entity_handle_p: *mut u32) -> ErrorCode
+```
+
+```Rust
+/// Search for wallet entities
+///
+/// storage_handle: opened storage handle (See open_wallet_storage)
+/// type_: allows to separate different entity types collections
+/// id: the id of entity
+/// query_json: MongoDB style query to wallet entity tags:
+///  {
+///    "tagName": "tagValue",
+///    $or: {
+///      "tagName2": { $regex: 'pattern' },
+///      "tagName3": { $gte: 123 },
+///    },
+///  }
+/// options_json: //TODO: FIXME: Think about replacing by bitmaks
+///  {
+///    retrieveEntities: (optional, true by default) If false only "counts" will be calculated,
+///    retrieveValue: (optional, true by default) Retrieve entity value,
+///    retrieveMetadata: (optional, true by default) Retrieve entity metadata
+///  }
+/// search_handle_p: pointer to store wallet search handle
+extern fn wallet_storage_search_entities(storage_handle: u32,
+                                         type_: *const c_char,
+                                         query_json: *const c_char,
+                                         options_json: *const c_char,
+                                         search_handle_p: *mut u32) -> ErrorCode
+```
+
+```Rust
+/// Close the wallet storage (For example, closing database connection)
+///
+/// storage_handle: opened storage handle (See open_wallet_storage)
+extern fn wallet_storage_close(storage_handle: u32) -> ErrorCode
+```
+
+### Wallet Entity API
+
+```Rust
+/// Get an id for retrieved wallet entity
+///
+/// entity_handle: retrieved entity handle (See wallet_storage_get_wallet_entity)
+wallet_entity_get_id(storage_handle: u32,
+                     entity_handle: u32) -> *const c_char
+```
+
+```Rust
+/// Get an value for retrieved wallet entity
+///
+/// entity_handle: retrieved entity handle (See wallet_storage_get_wallet_entity)
+wallet_entity_get_value(storage_handle: u32,
+                        entity_handle: u32) -> *const c_char
+```
+
+```Rust
+/// Get an metadata for retrieved wallet entity
+///
+/// entity_handle: retrieved entity handle (See wallet_storage_get_wallet_entity)
+wallet_entity_get_metadata(storage_handle: u32,
+                           entity_handle: u32) -> *const c_char
+```
+
+```Rust
+/// Free retrieved wallet entity (make retrieved entity handle invalid)
+///
+/// entity_handle: retrieved entity handle (See wallet_storage_get_wallet_entity)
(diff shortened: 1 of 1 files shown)


doc/wallet-storage.md
```diff
@@ -0,0 +1,255 @@
+# Wallet Storage Design
+
+In current state libindy allows to plug different wallet implementations. Plugged wallet now handles both security
```
ashcherbakov: What about what entities should be stored in wallet vs. multiple wallets vs. some storage?
Is it up to the app where to store, for example, claims?

I think there are 3 main options possible:
1) Only secrets (private, master secret) are stored in the wallet on libindy level. All other entities are assumed to be stored on app level.
2) Everything is stored in the wallet on libindy level. Here we assume that each libindy call works with **one** wallet only, so that by providing a proper wallet_handle, the app can decide what kind of wallet/implementation needs to be used. Here the wallet may mean something more general (like just a storage with optional security properties).

doc/wallet-storage.md
```diff
@@ -0,0 +1,255 @@
...
+and storage layers. This design proposes to restrict our plugged interface by handling only storage layer.
+All encryption will be performed in libindy. It will simplify plugged wallets and warranty good security level
+for 3d party wallets implementations.
+
+Also proposals enhances our API for efficient and flexible search with paging support.
+
+## Goals and ideas
+
+* Simplify plugged wallets and warranty security level for 3d party wallets implementation
+* Perform all encryption on libindy level:
```
ashcherbakov: What keys will be used for encryption? Where are they persisted? How libindy has access to them? Are there different keys, or one fey for the whole wallet?

doc/wallet-storage.md
```diff
@@ -0,0 +1,255 @@
...
+  * Entity ids will be always encrypted
+  * Entity values will be always encrypted
+  * Entity metadata will be always encrypted
+  * Tag names will be always encrypted
+  * Tag values will be **optionally** encrypted. If user wants to perform some complex searches it can be possible
+    to include some additional un-encrypted tags
+* Allow plugging of different storages
```
ashcherbakov: Do we consider that Wallet interface is public, so that the App can set/get/addTags to the wallet?

doc/wallet-storage.md
```diff
@@ -0,0 +1,255 @@
+# Wallet Storage Design
+
+In current state libindy allows to plug different wallet implementations. Plugged wallet now handles both security
+and storage layers. This design proposes to restrict our plugged interface by handling only storage layer.
+All encryption will be performed in libindy. It will simplify plugged wallets and warranty good security level
+for 3d party wallets implementations.
+
+Also proposals enhances our API for efficient and flexible search with paging support.
+
+## Goals and ideas
```
ashcherbakov: I think we need to have a diagram/doc that lists all libindy calls and how they work with the Wallet.
I think it confuses community a lot, where some methods doesn't work with wallet at all, some work with them quite expectedly, and some may have access to the wallet implicitly and surprisingly (like caches, etc.).
I think it will not only provide a better documentation and help community to understand the API, but also can check whether we always work with just ONE wallet at a time (so, we can support Case 2 from the comment above).

vimmerru:
> I don't think you can search tag values if you encrypt them.

In my vision the most of tags will be encrypted/hashed. It will allow to compare exact values in queries. For queries with "like" or "gte" predicates we can't use encryption. We can allow user to store some unencrypted tags with entities in the wallet for complex search purposes.

> "wallet entity" is too verbose; use "item" (we already have "wallet" as a prefix for every function)

API is object-oriented and prefix here means "class name". wallet_entity_get_id means WalleteEntity::get_id(), but i understand your point. 

> I'm not happy about using a string as an id for a wallet. For a few secrets, this is fine--but for a large wallet with tens of millions of keys, wouldn't it be more efficient to make the key an integer? (There is no reason to identify credentials with a string, BTW, since such identifiers have no independent meaning, unlike the naming convention we've chosen for secrets)

For the most of entities like Key, DID, ClaimDefinition we have natural id that libindy API accepts. We need the way to convert this id to wallet id. So i vote for string id.

> I'm dubious about the wisdom of doing mongo-db-style queries.

It is the point for discussion. I will vote for any query-style that is easy to parse and analyze. I see "tags" as an object in document-oriented database and mongo-style query is very natural way to query documents. I really don't like SQL because we will need to write our own parser and it is really non-obvious how to map tags to some tables.

> It's not clear to me how creating and opening wallet storage is intended to relate to creating and opening a wallet.

In my vision it is one-to-one mapping:

wallet_create calls storage_create 
wallet_open calls storage_open
and etc...

.vscode/settings.json
```diff
@@ -0,0 +1,15 @@
+{
```
jovfer: Please remove IDE settings from the repo

doc/design/wallet/storage-interface.puml
```diff
@@ -0,0 +1,35 @@
...
+    + add_record_tags(type, id, tags)
+    + delete_record_tags(type, id, tag_names)
+    + delete_record(type, id)    
+    + get_record(type, id, options) -> Record
+    + search_records(type, query, options) -> Search
+    +close()
+}
+
+interface Search {
+    + get_total_count() -> u32
```
dkulic: Counts can be expensive to get in some situations.

doc/design/wallet/wallet-design.md
```diff
@@ -0,0 +1,724 @@
...
+                                           /// name: wallet storage name (the same as wallet name)
+                                           /// config: wallet storage config (For example, database config)
+                                           /// runtime_config: wallet storage runtime config (For example, connection config)
+                                           /// credentials: wallet storage credentials (For example, database credentials)
+                                           /// storage_handle_p: pointer to store opened storage handle
+                                           open: Option<extern fn(name: *const c_char,
+                                                                  config: *const c_char,
+                                                                  runtime_config: *const c_char,
+                                                                  credentials: *const c_char,
+                                                                  storage_handle_p: *mut i32) -> ErrorCode>,
```
dkulic: May we make this storage handle *mut c_void. I would like to give storage implementer ability to put there pointer to whatever object it is the best for his implementation (for eg. pointer to the storage object or struct holding the internal state).

doc/design/wallet/wallet-design.md
```diff
@@ -320,7 +356,8 @@ pub extern fn indy_register_wallet_storage(command_handle: i32,
                                            add_record: Option<extern fn(storage_handle: i32,
                                                                         type_: *const c_char,
                                                                         id: *const c_char,
-                                                                        value: *const c_char,
+                                                                        value: *const u8,
```
dkulic: For value, it is used sometimes *const u8 and sometimes *const c_char. I would guess both will work but we should chose some.

doc/design/wallet/wallet-design.md
```diff
@@ -320,7 +356,8 @@ pub extern fn indy_register_wallet_storage(command_handle: i32,
                                            add_record: Option<extern fn(storage_handle: i32,
                                                                         type_: *const c_char,
                                                                         id: *const c_char,
```
dkulic: I would like it more if id and type_ values are passed as pointer to buffer and len (not using base64 or similar). I agree there is small performance impact because these values are small enough (less than 100 bytes), but it still to me makes not a sense to do that.

doc/design/wallet/wallet-design.md
```diff
@@ -0,0 +1,762 @@
...
+                                           ///     "tagName1": "tag value 1", // string value
+                                           ///     "tagName2": 123, // numeric value
+                                           ///   }
+                                           ///   Note that null means no tags
+                                           add_record: Option<extern fn(storage_handle: i32,
+                                                                        type_: *const c_char,
+                                                                        id: *const c_char,
+                                                                        value: *const u8,
+                                                                        value_len: usize,
+                                                                        tags_json: *const c_char) -> ErrorCode>,
```
dkulic: We also need to transfer value_key (encrypted key which is used for encrypting the value). Every value have its own key.
It is the same for all functions transferring the value.

doc/design/wallet/wallet-design.md
```diff
@@ -0,0 +1,762 @@
...
+                                           /// storage_handle: opened storage handle (See open handler)
+                                           /// type_: allows to separate different record types collections
+                                           /// id: the id of record
+                                           /// options_json: //TODO: FIXME: Think about replacing by bitmaks
+                                           ///  {
+                                           ///    retrieveValue: (optional, true by default) Retrieve record value,
+                                           ///    retrieveTags: (optional, true by default) Retrieve record tags
+                                           ///  }
+                                           /// record_handle_p: pointer to store retrieved record handle
+                                           get_record: Option<extern fn(storage_handle: u32,
```
dkulic: I think we do not need functions with record_handle. In get_record we can return the data immediately (if requested) or set the returned pointer to NULL if not requested.
I had in mind function like this: 
```
get_record: Option<extern fn(handle: i32,
                                class_p: *const c_char,
                                class_len: c_int,
                                name_p: *const c_char,
                                name_len: c_int,
                                options: *const c_char,
                                value_p: *mut *const c_char,
                                value_len: *mut c_int,
                                value_key_p: *mut *const c_char,
                                value_key_len: *mut c_int,
                                tags_p: *mut *const c_char) -> ErrorCode>;
```

doc/design/wallet/wallet-design.md
```diff
@@ -0,0 +1,762 @@
...
+                                          ///
+                                          /// #Params
+                                          /// storage_handle: opened storage handle (See open handler)
+                                          /// search_handle: wallet search handle (See search_records handler)
+                                          ///
+                                          /// returns: record handle (the same as for get_record handler)
+                                          ///          Note if no more records WalletNoRecords error will be returned
+                                          fetch_search_next_record: Option<extern fn(storage_handle: u32,
+                                                                                     search_handle: u32,
+                                                                                     record_handle_p: *mut i32) -> ErrorCode>,
```
dkulic: Here I would again avoid record handle and return the record immediately.

doc/design/wallet/wallet-design.md
```diff
@@ -0,0 +1,724 @@
...
+                                           /// type_: allows to separate different record types collections
+                                           /// id: the id of record
+                                           /// value: the value of record
+                                           /// tags_json: the record tags used for search and storing meta information as json:
+                                           ///   {
+                                           ///     "tagName1": "tag value 1", // string value
+                                           ///     "tagName2": 123, // numeric value
+                                           ///   }
+                                           ///   Note that null means no tags
+                                           add_record: Option<extern fn(storage_handle: i32,
```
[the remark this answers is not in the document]
vimmerru: As we discussed we will use record value as a byte array. type_, id, tags will be encoded as base64

doc/design/wallet/storage-interface.puml
```diff
@@ -0,0 +1,35 @@
+@startuml storage-interface
+
+interface WalletStorage {
+    + {static} create(name, config, cred)
```
dkulic: I am not sure we should skip WalletStorageType struct :-/

doc/design/wallet/storage-interface.puml
vimmerru: Count will be calculated optionally. It can be fast enough for a lot of use cases.

doc/design/wallet/wallet-design.md
vimmerru: I have big concern about this proposal. When you provide the pointer you transfer memory ownership and should provide the way to free memory. We have mix of types here and will require a lot of handlers to free memory. Providing record_handle solve this problem. We have just one free_record function that consumes this handle.

doc/design/wallet/wallet-design.md
```diff
@@ -0,0 +1,762 @@
...
+                      credentials: &str) -> Result<i32, WalletError> {}
+
+   pub fn close_wallet(&self,
+                       wallet_handle: i32) -> Result<(), WalletError> {}
+
+   pub fn list_wallets(&self) -> Result<Vec<WalletMetadata>, WalletError> {}
+
+   pub fn add_record(wallet_handle: i32,
+                     type_: &str,
+                     id: &str,
```
dkulic: missing value.

doc/design/wallet/wallet-design.md
```diff
@@ -0,0 +1,768 @@
...
+/// open: Wallet storage "open" operation handler
+/// close: Wallet storage "close" operation handler
+/// add_record: Wallet storage "add_record" operation handler
+/// update_record_value: Wallet storage "update_record_value" operation handler
+/// update_record_tags: Wallet storage "update_record_tags" operation handler
+/// add_record_tags: Wallet storage "add_record_tags" operation handler
+/// delete_record_tags: Wallet storage "delete_record_tags" operation handler
+/// delete_record: Wallet storage "delete_record" operation handler
+/// get_record: Wallet storage "get_record" operation handler
+/// get_record_id: Wallet storage "get_record_id" operation handler
```
dkulic: get_record_type would also be needed for export.

doc/design/wallet/wallet-design.md
```diff
@@ -0,0 +1,768 @@
...
+///  }
+/// #Returns
+/// wallet record json:
+/// {
+///   id: "Some id",
+///   value: "Some value", // present only if retrieveValue set to true
+///   tags: <tags json>, // present only if retrieveTags set to true
+/// }
+extern pub fn indy_get_wallet_record(command_handle: u32,
+                                     wallet_handle: i32,
```
dkulic: handles are sometimes i32, sometimes u32, which one is correct?

doc/design/wallet/wallet-design.md
dkulic: OK, this is acceptable.

doc/design/wallet/wallet-design.md
```diff
@@ -0,0 +1,768 @@
...
+                                          get_record_tags: Option<extern fn(storage_handle: u32,
+                                                                            record_handle: u32,
+                                                                            record_tags_p: *mut *const c_char) -> ErrorCode>,
+
+                                          /// Free retrieved wallet record (make retrieved record handle invalid)
+                                          ///
+                                          /// #Params
+                                          /// storage_handle: opened storage handle (See open_wallet_storage)
+                                          /// record_handle: retrieved record handle (See wallet_storage_get_wallet_record)
+                                          free_record: Option<extern fn(storage_handle: u32,
```
dkulic: There is no use for ErrorCode in free functions. You cannot do anything. It is probably the best if those functions do not have return value.

doc/design/wallet/wallet-design.md
```diff
@@ -0,0 +1,768 @@
...
+/// search_records: Wallet storage "search_records" operation handler
+/// get_search_total_count: Wallet storage "get_search_total_count" operation handler
+/// get_search_count: Wallet storage "get_search_count" operation handler
+/// get_search_next_record: Wallet storage "get_search_next_record" operation handler
+/// free_search: Wallet storage "free_search" operation handler
+///
+/// #Returns
+/// Error code
+#[no_mangle]
+pub extern fn indy_register_wallet_storage(command_handle: i32,
```
dkulic: Order of extern functions seams totally random. We should create some logical order.

doc/design/wallet/wallet-design.md
```diff
@@ -0,0 +1,768 @@
...
+/// get_record: Wallet storage "get_record" operation handler
+/// get_record_id: Wallet storage "get_record_id" operation handler
+/// get_record_value: Wallet storage "get_record_value" operation handler
+/// get_record_tags: Wallet storage "get_record_tags" operation handler
+/// free_record: Wallet storage "free_record" operation handler
+/// search_records: Wallet storage "search_records" operation handler
+/// get_search_total_count: Wallet storage "get_search_total_count" operation handler
+/// get_search_count: Wallet storage "get_search_count" operation handler
+/// get_search_next_record: Wallet storage "get_search_next_record" operation handler
+/// free_search: Wallet storage "free_search" operation handler
```
dkulic: For export functionality it is needed some get_all or similar which will return search handle, so that wallet can retrieve all records from database.

doc/design/wallet/wallet-design.md
vimmerru: I suggest to leve error code. It isn’t system free and allows internal handler and state validation.

doc/design/wallet/wallet-design.md
vimmerru: Correct, i32 for handles. Count should be usize. I will update.

doc/design/wallet/wallet-design.md
vimmerru: Got it, will provide solution

PR #595, opened 2018-03-26
Samples and guides

keichiri:
First batch of scripts in Python and Java for how-tos


dhh1128:
@keichiri You need to sign off your commit. This tells Hyperledger that you are acknowledging that you can't contribute IP encumbered by a patent, that you own copyright to what you commit, etc. See https://github.com/probot/dco#how-it-works. You do this by modifying your commit command -- git commit -s -m "commit message" (the -s is what does the sign-off by adding "Signed-off-by: <you>" to the commit message). If you've already pushed, you can modify the head commit of your branch and then re-raise the PR. We can't accept the PR until it has this sign-off in it.

PR #609, opened 2018-03-27
Add restart pool transaction

SergeyPalamarchuk:


libindy/out_libindy_pod/indy_anoncreds.h
```diff
@@ -0,0 +1,167 @@
+#ifndef __anoncreds__included__
```
jovfer: this is temporary directory and should not be included into VCS.

libindy/src/services/ledger/mod.rs
```diff
@@ -161,6 +161,26 @@ impl LedgerService {
             .map_err(|err| CommonError::InvalidState(format!("Invalid pool_config request json: {:?}", err)))
     }
 
+    pub fn build_pool_restart(&self, identifier: &str, action: &str, schedule: Option<&str>) -> Result<String, CommonError> {
+        let schedule = match schedule {
+            Some(schedule) => Some(serde_json::from_str::<HashMap<String, String>>(schedule)
```
jovfer: @ashcherbakov @spivachuk @Toktar 
AFAIK from our last discussion, the schedule for pool restart should be just single time (not map node->time), shouldn't it?

Toktar: Yes, the schedule just single time. I'll send the documentation soon. The query looks like this:
{'identifier': 'M9BJDuS24bqbJNvBRsoGg3', 
'operation': 
    {'schedule': '2018-03-28 16:00:41.442782+00:00', 
     'action': 'start', 'type': '118'
     }, 
'reqId': 60900, 
'protocolVersion': 1}

jovfer: @SergeyPalamarchuk please re-check and fix sdk code if needed.

Toktar: @jovfer , @SergeyPalamarchuk , field 'schedule' was renamed to 'datetime'

PR #614, opened 2018-03-28
Added guidelines for contributing to indy wrappers

faisal00813:
Signed-off-by: Mohammad Abdul Sami <faisal00813@gmail.com>


wrappers/README.md
```diff
@@ -0,0 +1,36 @@
+# Guidelines for Libindy Wrappers
```
jovfer: As this readme acutally contains Guidelines for wrappers developer, may be it's better to don't use base `README.md` filename, but rename it to something like `guidelines.md` ?

jovfer:
Moved to #615

PR #622, opened 2018-04-04
New wallet - initial commit

dkulic:
This is initial commit so we can work on interfaces.


dkulic:
No signoff

PR #630, opened 2018-04-06
Bugfix/refresh

Artemkaaas:


jovfer:
It would be nice to add appropriate integration or unit test for error case.

PR #640, opened 2018-04-12
Payments Interface Design

vimmerru:
* Added description of Payment Method API


brentzundel:
What is here looks good.
I don't see a way to set fees (build_fees_txn) or query fees (build_get_fees_txn).
Those were mentioned in the "Token library and CLI WBS" document.

I'm not sure if this is the place to include it, or if another pull request against this document should be made in the future, but there will also need to be libindy functionality to generically support 'paid credentials.'  
In my view this could be done with an indy_issuer_create_paid_credential_offer that mimics indy_issuer_create_credential_offer, and an indy_prover_create_paid_credential_req that mimics indy_prover_create_credential_req.

jovfer:
@brentzundel 
1) I've added missed build_fees_txn with getter. Thank you for note!
2) For me, @vimmerru and @ashcherbakov 'paid credentials' is scope of separate discussion, PR, etc.

jovfer:
I've added my `approve` for the PR, but there were some changes from me, so another one review is required.

PR #653, opened 2018-04-14
[WIP] Update sodiumoxide from version 0.0.14 to 0.0.16.

peacekeeper:
This should fix this problem mentioned in Rocketchat: https://chat.hyperledger.org/channel/indy-sdk?msg=BkFvNq3zpyuHLqGBR

```
  = note: /usr/local/lib/../lib/libindy.so: undefined reference to 'crypto_stream_aes128ctr_xor'
          /usr/local/lib/../lib/libindy.so: undefined reference to 'crypto_stream_aes128ctr'
```


vimmerru:
@peacekeeper Thanks for pointing to new sodiumoxide. It can't be merged as is as it requires to modify our packages and environment to use new libsodium binaries.

dbluhm:
So it looks like the current problem here is that the CI/CD is using a version of libsodium that is too old to play nice with sodiumoxide 0.0.16. Would it be an acceptable workaround to build libsodium 1.0.16 on our ci dockerfile instead of installing libsodium 1.0.8 through apt?

vimmerru:
We have binary packages for multiple platforms. Ubuntu 10.04 is still the most common linux distro and we need the simple way to install our packages here.

AxelNennker:
Somebody did the change in the meantime without using this PR.
https://github.com/hyperledger/indy-sdk/blob/master/libindy/Cargo.toml#L60
This can be closed, I think.

jovfer:
In scope of #696 sodium was incremented to .16v . To pass CI ubuntu now use custom build of system libsodium instead of sodium18 from xenial apt repo

PR #672, opened 2018-04-23
Fix import in Python getting-started

### LINKED ISSUE #633: python samle import src.utils
from src.utils import get_pool_genesis_txn_path, run_coroutine

should be from utils

lodo1995:
resolves #633


samples/python/src/getting_started.py
```diff
@@ -8,7 +8,7 @@
 
 from indy.error import ErrorCode, IndyError
 
-from src.utils import get_pool_genesis_txn_path, run_coroutine
+from utils import get_pool_genesis_txn_path, run_coroutine
```
lovesh: Having absolute imports is better. The problem is a missing `__init__.py` file in `src`. `src` should be a package

lodo1995:
@lovesh I think I made some confusion with the commits, but now the imports are absolute, the `src` directory is a module and I added some info in `README.md` about how to run a script inside a module.

PR #685, opened 2018-04-26
[WIP] Add Null handler for Payments API

KitHat:


null-payment-plugin/src/payments/payment_callbacks.rs
```diff
@@ -0,0 +1,395 @@
...
+type CommonResponseCallback = extern fn(command_handle_: i32,
+                                        err: ErrorCode,
+                                        res1: *const c_char) -> ErrorCode;
+
+lazy_static! {
+    static ref IDS_COUNTER: AtomicUsize = ATOMIC_USIZE_INIT; //TODO use AtomicI32
+}
+
+pub extern fn nullpayment_init(cb: Option<extern fn(cmd_handle: i32, err: ErrorCode)>) {
+    let _cmd_handle = get_next_id();
```
vimmerru: Why we generate new cmd_handle here instead of pass user's value directly?

KitHat: Fixed this

PR #691, opened 2018-04-27
Default wallet storage implementation with encryption layer and wql

keichiri:
This is all the work so far.

Plugged wallet is being finished by Darko.
Enterprise wallet is being finished by Darko and Nem.

We expect that to be added in next couple of days.

After that, the commented tests will be uncommented

### DIFF
--- a/libindy/Cargo.toml
+++ b/libindy/Cargo.toml
@@ -44,7 +44,7 @@ libc = "0.2.21"
 log = "0.3.7"
 openssl = { version = "=0.9.24", optional = true }
 rand = "0.3"
-rusqlcipher = { version = "0.14.6", features=["bundled"] }
+rusqlite = "0.13.0"
 rust-base58 = {version = "0.0.4", optional = true}
 base64 = "0.6.0"
 serde = "1.0"
@@ -53,7 +53,7 @@ serde_derive = "1.0"
 sha2 = "0.6.0"
 sha3 = "0.6.0"
 rmp-serde = "0.13.6"
-sodiumoxide = {version = "0.0.14", optional = true}
+sodiumoxide = { git="https://github.com/ccfb3ee765a58cae/sodiumoxide.git", branch="xchacha20-poly1305-support", optional = true }
 time = "0.1.36"
 zmq = "0.8.2"
 lazy_static = "1.0"
--- a/libindy/src/api/mod.rs
+++ b/libindy/src/api/mod.rs
@@ -89,6 +89,27 @@ pub enum ErrorCode
     // Attempt to open encrypted wallet with invalid credentials
     WalletAccessFailed = 207,
 
+    // Input provided to wallet operations is considered not valid
+    WalletInputError = 208,
+
+    // Decoding of wallet data during input/output failed
+    WalletDecodingError = 209,
+
+    // Storage error occurred during wallet operation
+    WalletStorageError = 210,
+
+    // Error during encryption-related operations
+    WalletEncryptonError = 211,
+
+    // Requested wallet item not found
+    WalletItemNotFound = 212,
+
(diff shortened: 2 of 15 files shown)


libindy/Cargo.toml
```diff
@@ -53,7 +53,7 @@ serde_derive = "1.0"
 sha2 = "0.6.0"
 sha3 = "0.6.0"
 rmp-serde = "0.13.6"
-sodiumoxide = {version = "0.0.14", optional = true}
+sodiumoxide = { git="https://github.com/ccfb3ee765a58cae/sodiumoxide.git", branch="xchacha20-poly1305-support", optional = true }
```
vimmerru: Only packages form crates.io are allowed to be merged to master

libindy/src/errors/wallet.rs
```diff
@@ -18,12 +23,20 @@ pub enum WalletError {
     AlreadyExists(String),
     NotFound(String),
     IncorrectPool(String),
-    PluggedWallerError(ErrorCode),
+    PluggedWalletError(ErrorCode),
```
vimmerru: I don't understand what is the difference between PluggedWalletError and StorageError

libindy/src/api/mod.rs
```diff
@@ -89,6 +89,27 @@ pub enum ErrorCode
...
+    WalletStorageError = 210,
+
+    // Error during encryption-related operations
+    WalletEncryptonError = 211,
+
+    // Requested wallet item not found
+    WalletItemNotFound = 212,
+
+    // Returned if wallet's add_record operation is used with record name that already exists
+    WalletItemAlreadyExists = 213,
```
vimmerru: What is the difference with existing WalletAlreadyExistsError ?

libindy/Cargo.toml
keichiri: Well that's what we have to figure how to do then. If sodiumoxide does not have it in the master by then, we'll have to fork it and push our own package to crates.io?

libindy/src/errors/wallet.rs
```diff
@@ -82,22 +116,158 @@ impl ToErrorCode for WalletError {
...
+            WalletError::ItemAlreadyExists => ErrorCode::WalletItemAlreadyExists,
+            WalletError::QueryError(_) => ErrorCode::WalletQueryError,
         }
     }
 }
 
 impl From<io::Error> for WalletError {
     fn from(err: io::Error) -> WalletError {
-        WalletError::CommonError(CommonError::IOError(err))
+        WalletError::CommonError(CommonError::IOError((err)))
```
vimmerru: why we need redundant braces?

libindy/src/api/mod.rs
keichiri: WalletAlreadyExists - when wallet itself exists already (wallet with given name)
WalletItemAlreadyExists - when the wallet item already exists (wallet record)

libindy/src/errors/wallet.rs
keichiri: PluggedWalletError was left for now, until we finish plugged wallet and figure out whether we should remove it. 
By that time, it will probably be removed

libindy/src/services/wallet/encryption.rs
```diff
@@ -0,0 +1,131 @@
+use std::collections::HashMap;
+
+use sodiumoxide::crypto::aead::xchacha20poly1305_ietf;
```
vimmerru: All crypto must be wrapper in utils\crypto. Direct links to libsodium in code isn't allowed

libindy/Cargo.toml
jovfer: And what about system libraries for all platforms?

keichiri: @jovfer What do you mean exactly?

PR #694, opened 2018-04-27
Add load test for libindy Crypto::auth_crypt.

jovfer:


libindy/tests/crypto.rs
```diff
@@ -591,3 +592,100 @@ mod high_cases {
...
+                time_diffs
+            });
+            results.push(thread);
+        }
+
+        let mut all_diffs = Vec::new();
+        for result in results {
+            all_diffs.push(result.join().unwrap());
+        }
+        let total_duration = SystemTime::now().duration_since(start_time).unwrap();
```
nveskovic: Total duraton should be calculated right after threads finish (after the `for` loop)

jovfer: @nveskovic I don't understand your comment, please add more details... It's calculated as you mentioned, right after `for` with joining to children threads. I see only one thing to optimize: unwrapping results and storing into Vec, but it's not significant, I think.

nveskovic: Since @vimmerru already merged it **before** any additional changes were made, then there is not much use of a comment here. Nevertheless, what I was thinking is

Total duration should be calculated right after 
```
for (wallet, verkey, data) in agents {
...
}
```
not after additional
```
let mut all_diffs = Vec::new();
for result in results {
     all_diffs.push(result.join().unwrap());
}
```

Not a big difference but why not having accurate data if it is only a matter of sequence what's done first

hope it is clear now

vimmerru: @nveskovic 

```for (wallet, verkey, data) in agents``` just finishes immediately as each loop iteration just starts new thread.  Next loop performs waiting for all threads and it is the correct place to put total time measurement.

jovfer: @nveskovic in other words, we should wait finishing all threads, so `.join()` call is required. Please see https://doc.rust-lang.org/std/thread/ .

PR #696, opened 2018-04-28
[WIP] Wallet Storage

vimmerru:
Fix before merge to master
- [x] system library dependence for updated libsodium
- [ ] ~restore iOS test for plugged wallet~
- [ ] remove `abort`-like code - will be moved to separate task
- [x] restore wrappers tests


libindy/Cargo.toml
```diff
@@ -53,10 +53,12 @@ serde_derive = "1.0"
 sha2 = "0.6.0"
 sha3 = "0.6.0"
 rmp-serde = "0.13.6"
-sodiumoxide = {version = "0.0.14", optional = true}
+sodiumoxide = { git="https://github.com/ccfb3ee765a58cae/sodiumoxide.git", branch="xchacha20-poly1305-support", optional = true }
```
jovfer: Should not be merged to master.

libindy/Cargo.toml
```diff
@@ -53,10 +55,12 @@ serde_derive = "1.0"
 sha2 = "0.6.0"
 sha3 = "0.6.0"
 rmp-serde = "0.13.6"
-sodiumoxide = {version = "0.0.14", optional = true}
+sodiumoxide = {version = "0.0.16", optional = true}
```
jovfer: This update may be problimatic for some platforms

jovfer: #653

vimmerru:
@dhh1128 

> Also, I want to eliminate the relationship between wallets and pools. A wallet should be able to hold info from multiple pools, with the DID method prefix guaranteeing uniqueness (that is, we store "did:sov:abcxyz", not just "abcxzy" as the wallet key.

I suggest to implement this as a dedicated PR. Too much for one PR.

PR #699, opened 2018-04-30
[WIP] remove duplicate code

lovesh:
Signed-off-by: Lovesh Harchandani <lovesh.bond@gmail.com>


libindy/src/services/crypto/mod.rs
```diff
@@ -43,22 +43,60 @@ impl CryptoService {
         }
     }
 
-    pub fn create_key(&self, key_info: &KeyInfo) -> Result<Key, CryptoError> {
-        let crypto_type_name = key_info.crypto_type
-            .as_ref()
-            .map(String::as_str)
-            .unwrap_or(DEFAULT_CRYPTO_TYPE);
+    pub fn get_crypto_type_name<'a>(crypto_type: &'a Option<String>) -> (&'a str) {
```
vimmerru: I like the idea of reducing boilerplate, but we need to follow some rules here:

1. Each function should have a single responsibility. 
2. Each function should have obvious responsibility and a good name
3. We shouldn't expose private functions
4. Private functions should have a name started with '_'

vimmerru: > pub fn get_crypto_type_name<'a>(crypto_type: &'a Option<String>) -> (&'a str)

It seems like non-obvious function and i suggest to remove it

libindy/src/services/crypto/mod.rs
```diff
@@ -43,22 +43,60 @@ impl CryptoService {
...
+    pub fn is_crypto_type_supported(&self, crypto_type_name: &str) -> bool {
+        self.crypto_types.contains_key(crypto_type_name)
+    }
 
-        if !self.crypto_types.contains_key(crypto_type_name) {
+    fn get_crypto_type(&self, crypto_type_name: &str) -> &Box<CryptoType> {
+        self.crypto_types.get(crypto_type_name).unwrap()
+    }
+
+    pub fn get_crypto_name_and_keypair<'a>(&self, crypto_type: &'a Option<String>,
```
vimmerru: > pub fn get_crypto_name_and_keypair<'a>(&self, crypto_type: &'a Option<String>,
>                                               seed: &'a Option<String>) -> Result<(&'a str, Vec<u8>, Vec<u8>), CryptoError>

This function mix a lot of responsibilities. So i suggest to remove this. but we can update ```get_crypto_type``` to cover the most of boilerplate without introducing single responsibility problems

libindy/src/services/crypto/mod.rs
```diff
@@ -43,22 +43,60 @@ impl CryptoService {
...
+            (verkey, DEFAULT_CRYPTO_TYPE)
+        }
+    }
+
+    pub fn is_crypto_type_supported(&self, crypto_type_name: &str) -> bool {
+        self.crypto_types.contains_key(crypto_type_name)
+    }
 
-        if !self.crypto_types.contains_key(crypto_type_name) {
+    fn get_crypto_type(&self, crypto_type_name: &str) -> &Box<CryptoType> {
```
vimmerru: - All unwraps must be checked on the same function level

vimmerru: It will be much better if this function will take responsibility of checking that crypto type is supported and return Result. It will allow to reduce boilerplate without introducing of functions like ```get_crypto_name_and_keypair```

jovfer:
@lovesh I will mark this PR as [WIP] because some discussion threads are in progress. Please remove this suffix after necessary changes will be done.

vimmerru:
No any progress here for month and codebase is changes significantly. I am closing this,

PR #711, opened 2018-05-03
Add Null handler for Payments API

KitHat:


vimmerru:
We can merge it as is now to unblock Artem, but there are places for improvements. Let's discuss together today.

PR #715, opened 2018-05-04
deployment config for nodejs wrapper

farskipper:
We'll need a npm user to publish this. Maybe this one https://www.npmjs.com/~hyperledger-ci ?

Then store the auth token (npmrc) in the `npm_credentials` file. Similar to `pypi_credentials`

### DIFF
--- a/Jenkinsfile.cd
+++ b/Jenkinsfile.cd
@@ -519,7 +519,9 @@ def nodejsWrapperPublishing(testEnv, isRelease) {
         def suffix = getSuffix(isRelease, "nodejs")
 
         testEnv.inside {
-            # TODO npm user for publishing ~/.npmrc
+            sh 'npm config set registry https://registry.npmjs.org/'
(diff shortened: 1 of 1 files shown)


vimmerru:
@farskipper Could you fix conflicts? Also as i understand we need integrate the same changes as for CI PR.

vimmerru:
@farskipper We can create new npm user, but i am a bit afraid that this user can post artifacts with the same name. What do you think about this?

farskipper:
@vimmerru Ok, I updated the branch and fixed the merge conflict.

Not sure what you meant by "afraid this user can post artifacts with the same name."  If we create a new npm user, I'll give it permission to publish to `npmjs.com/package/indy-sdk`; and this `Jenkinsfile.cd` is setup to publish pre-releases as well as full releases.

ArthurManz:
Why is this PR paused?

farskipper:
I can transfer ownership of the npm package `indy-sdk` to `~hyperledger-ci` or whatever user you'll use to publish the builds. Just let me know which user should own it.

PR #717, opened 2018-05-04
Cli payment commands

Artemkaaas:


vimmerru:
I suggest to merge it as is, but we will need to make some impovments

PR #719, opened 2018-05-04
Moved encryption to crypto dir. Changed cypher to Chacha20-Poly_IETF

dkulic:

### DIFF
--- a/libindy/src/utils/crypto/chacha20poly1305_ietf/sodium.rs
+++ b/libindy/src/utils/crypto/chacha20poly1305_ietf/sodium.rs
@@ -1,0 +1,105 @@
+extern crate sodiumoxide;
+
+use errors::common::CommonError;
+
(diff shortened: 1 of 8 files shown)


libindy/src/utils/crypto/chacha20poly1305_ietf/sodium.rs
```diff
@@ -0,0 +1,105 @@
+extern crate sodiumoxide;
+
+use errors::common::CommonError;
+
+use self::sodiumoxide::crypto::aead::chacha20poly1305_ietf;
+use sodiumoxide::crypto::auth::hmacsha256;
+use utils::byte_array::_clone_into_array;
+
+pub struct ChaCha20_Poly1305_IETF {
```
vimmerru: Just one minor request. Struct names must be camel case without underscores.

dkulic: Renamed.

PR #726, opened 2018-05-07
Feature/wallet update functionalities

keichiri:
Added:
* update_record_value
* add_record_tags
* update_record_tags
* delete_record_tags

to all layers (from WalletService to default wallet)

Also, removed old default wallet (forgot to remove before)


libindy/src/errors/wallet.rs
```diff
@@ -196,7 +197,10 @@ pub enum WalletStorageError {
 
 impl From<rusqlite::Error> for WalletStorageError {
     fn from(err: rusqlite::Error) -> WalletStorageError {
-        WalletStorageError::IOError(format!("IO error during storage operation: {}", err.description()))
+        match &err {
```
vimmerru: I um not comfortable with implicit checking for item availability.

PR #733, opened 2018-05-08
update embedded links for documents moved

kdenhartog:
Signed-off-by: Kyle Den Hartog <kdenhar@gmail.com>


kdenhartog:
Removed in place of PR #734 to move docs back to the original locations so that the links can be found by those who have already had them shared and included in documents referencing this work.

PR #734, opened 2018-05-08
updated links and added temporary place holder to share link

kdenhartog:
These commits reflect an update to the links after the DKMS work was moved to the doc/design/005-dkms repo. It also adds a placeholder file for a link that was placed in a document that was shared in a congressional hearing. It can be deprecated in 6 months (11/8/18) as stated in the place holder. The link provided in that place holder doc will remain persistent because it links to commit: 677a0439487a1b7ce64c2e62671ed3e0079cc11f


vimmerru:
Why we don't use relative links here?

kdenhartog:
I didn't add them at the beginning because I didn't have a relative link at the beginning. I'll submit a PR that switches them to relative links.

PR #745, opened 2018-05-11
Add new error codes for insufficient funds and no key for payment add…

KitHat:
…ress


wrappers/java/src/main/java/org/hyperledger/indy/sdk/ErrorCode.java
```diff
@@ -241,7 +241,17 @@
 	/**
 	 * Information passed to libindy is incorrect
 	 */
```
vimmerru: Seems we need to remove this?

wrappers/java/src/main/java/org/hyperledger/indy/sdk/ErrorCode.java
```diff
@@ -241,7 +241,17 @@
...
+
+	/**
+	 * Insufficient funds on inputs
+	 */
+	InsufficientFundsError(703),
+
+	/**
+	 * No key for this payment address in wallet
+	 */
+	NoPaymentAddressKeyError(704),
```
vimmerru: Seems you need to update java wrapper too

PR #746, opened 2018-05-11
Fix CLI CI/CD for payments on linux-based platforms.

jovfer:


jovfer:
Most probably we will change structure of builds and dependencies, so this PR is not actual now.

PR #768, opened 2018-05-17
[WIP] Rewrite libnullpay to act like a real ledger

KitHat:


libnullpay/src/payment_method.rs
```diff
@@ -1,227 +1,305 @@
...
-        (cb)(cmd_handle, err, res)
+        let inputs = from_str::<Vec<String>>(inputs_json);
+        let outputs = from_str::<Vec<UTXOOutput>>(outputs_json);
+
+        let (inputs, outputs) = match (inputs, outputs) {
+            (Ok(inputs), Ok(outputs)) => (inputs, outputs),
+            _ => { return ErrorCode::CommonInvalidStructure; }
+        };
+
+        let txn_type = match parse_operation_from_request(res) {
```
vimmerru: Can we just use "?" operator? A lot of matches like this doesn't look really nice

libnullpay/src/payment_method.rs
```diff
@@ -1,227 +1,305 @@
...
+        let res = unsafe { CStr::from_ptr(req_json).to_str() }.unwrap();
+        let inputs_json = unsafe { CStr::from_ptr(inputs_json).to_str() }.unwrap();
+        let outputs_json = unsafe { CStr::from_ptr(outputs_json).to_str() }.unwrap();
 
-    fn handle(cmd_handle: i32, _wallet_handle: i32, _submitter_did: *const c_char, req_json: *const c_char, _inputs_json: *const c_char, _outputs_json: *const c_char, cb: IndyPaymentCallback) -> ErrorCode {
-        let res = req_json;
-        let err = ErrorCode::Success;
-        (cb)(cmd_handle, err, res)
+        let inputs = from_str::<Vec<String>>(inputs_json);
+        let outputs = from_str::<Vec<UTXOOutput>>(outputs_json);
```
vimmerru: It can be better to define domain entities and implement FromJson crate. Also it will cause simpler execution flow. Let's discuss a bit later.

PR #777, opened 2018-05-21
Android builds arm arm64 and x86

faisal00813:
Signed-off-by: Mohammad Abdul Sami <faisal00813@gmail.com>


lodo1995:
@faisal00813 the path `/sdcard/Documents/.indy_client` is not portable to all devices. For example, my phone, which is a Samsung running KitKat with no sd card, does not have it. On my phone, the documents folder is `/storage/emulated/0`.

libindy/src/utils/environment.rs
```diff
@@ -11,7 +11,7 @@ impl EnvironmentUtils {
         if cfg!(target_os = "ios"){
             indy_client_dir = "Documents/.indy_client"
         }else if cfg!(target_os = "android"){ //TODO: FIX ME: Check for more secure location for the wallet.
-            indy_client_dir = "/sdcard/Documents/.indy_client"
+            indy_client_dir = env::var("EXTERNAL_STORAGE") +  ".indy_client"
```
lodo1995: On my phone, the predefined environment variable `EXTERNAL_STORAGE` does not have a trailing `/`. Wouldn't it be better to perform a path concatenation instead of a simple string concatenation?

doc/android-build.md
```diff
@@ -0,0 +1,37 @@
+# Building binaries of LibIndy for Android
+
+**Not ready for production use! Not fully tested.**
+
+## Prerequisites
+
+- Docker
```
vimmerru: Can i build it on windows and MacOS?

libindy/Cargo.toml
```diff
@@ -38,17 +38,17 @@ sodium_static = []
 indy-crypto = { version = "=0.4.1", optional = true }
 int_traits = { version = "0.1.1", optional = true }
 digest = "0.6.2"
-env_logger = "0.4.2"
+env_logger = "0.5.10"
```
vimmerru: Is it updated intentionally?

libindy/Cargo.toml
```diff
@@ -38,17 +38,17 @@ sodium_static = []
...
 errno = "0.2.3"
 etcommon-rlp = "0.2.3"
 generic-array = "0.8.3"
 hex = "0.2.0"
 libc = "0.2.21"
-log = "0.3.7"
+log = "0.4.1"
```
vimmerru: Is it updated intentionally?

libindy/Cargo.toml
```diff
@@ -38,17 +38,17 @@ sodium_static = []
...
 openssl = { version = "=0.9.24", optional = true }
 owning_ref = "0.3.3"
 rand = "0.3"
-rusqlite = "0.13.0"
+rusqlite = { version = "0.13.0", features=["bundled"] }
```
vimmerru: As i understand now we start static linking with sqlite on all platforms.

libindy/Cargo.toml
faisal00813: No. We require 0.4.2 for android_logger

libindy/Cargo.toml
faisal00813: Yes

libindy/build.rs
```diff
@@ -32,4 +32,40 @@ fn main() {
             }
         }
     }
+		match &target {
+		x if x.contains("aarch64-linux-android") || x.contains("armv7-linux-androideabi") ||
```
vimmerru: Can we just replace this complex if with just "linux-android" as it seems present in all arch strings?

libindy/Cargo.toml
faisal00813: On Android, we don't have SQLite binaries which we provide while building. Also, I can't recall any flag which suggests that we are doing static linking with SQLite.

doc/android-build.md
faisal00813: Only Linux is supported as of now.

libindy/Cargo.toml
vimmerru: The problem is that we don't want to force static linking on all platforms. Only for android and iOS

libindy/build_scripts/android/Dockerfile
```diff
@@ -0,0 +1,49 @@
+FROM ubuntu:16.04
```
vimmerru: I believe we need to rename it to something more obvious like android-build-env.dockerfile

doc/android-build.md
```diff
@@ -0,0 +1,37 @@
...
+
+## How to build.
+- Goto `indy-sdk/libindy/build_scripts/android` folder
+- Run `build.dependencies.locally.sh`
```
vimmerru: > Run `build.dependencies.locally.sh`

Will it build dependencies for all supported architectures?

libindy/build_scripts/android/dependencies/openssl/build.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+ARMV7_DIR=${PWD}/openssl_arm
+ARM64_DIR=${PWD}/openssl_arm64
+ARMx86_DIR=${PWD}/openssl_x86
+mkdir -p $ARM64_DIR
+mkdir -p $ARMV7_DIR
+mkdir -p $ARMx86_DIR
+
+git clone https://github.com/nsivraj/openssl_for_ios_and_android.git
```
vimmerru: Build must be repeatable. Se we can't use just master branch for dependencies. Use tag/release/commit

libindy/src/utils/environment.rs
```diff
@@ -1,13 +1,25 @@
 use std::env;
 use std::path::PathBuf;
+use std::fs;
 
 pub struct EnvironmentUtils {}
 
 impl EnvironmentUtils {
     pub fn indy_home_path() -> PathBuf {
         // TODO: FIXME: Provide better handling for the unknown home path case!!!
+
```
vimmerru: This code doesn't look correct for me.

- I see 2 unwraps. Will we crash if there is no EXTERNAL_STORAGE variable?
- Why we create directories here? It isn't responsibility of this function
- Flow also looks very non-obvious? Can you just use on match or if/if else/if else?

AxelNennker:
I suggest not to use cd ../../../.. to get to the root of indy-sdk from the build scripts but to use
e.g. for arm64

#!/usr/bin/env bash

BASEDIR=$(dirname "$0")

DEPS_FOLDER=$BASEDIR/dependencies
LIBZMQ=${DEPS_FOLDER}/zmq/libzmq_arm64
OPENSSL=${DEPS_FOLDER}/openssl/openssl_arm64
LIBSODIUM=${DEPS_FOLDER}/sodium/libsodium_arm64
$BASEDIR/build.sh -f arm64 21 aarch64-linux-android android_support ${OPENSSL} ${LIBSODIUM} ${LIBZMQ}

doc/android-build.md
vimmerru: Can you add this info to per-requirements? It is better to provide info about platform that we tested.

vimmerru:
It would be nice to have a script in libindy root folder that will build artifacts for all platforms.

vimmerru:
You use https://github.com/nsivraj/openssl_for_ios_and_android to build openssl, This script isn't performed in isolated environment. Does it require any pre-requirements? Can we just use pre-built binary from the same site?

vimmerru:
Build looks a bit over-complicated now. You use dockerfiles that just perform builds of one thing. It can be much faster (and easy to understand)  to have one dockerfile that setups all necessary environment and set of scripts to execute inside of this dockerfile.

vimmerru:
Is there a way to execute  tests?

libindy/Cargo.toml
vimmerru: ok, make sense

libindy/Cargo.toml
vimmerru: ok, make sense

doc/android-build.md
faisal00813: That info is present in _Known Issues_ segment at the bottom of the document.

doc/android-build.md
faisal00813: Yes. The script will build for x86, arm and arm64 architectures.

doc/android-build.md
vimmerru: ok

AxelNennker:
Android NDK 17 is out. Update this?

faisal00813:
@AxelNennker I don't see any immediate need for updating to NDK17. The changelog suggests deprecation of GCC and we use CLANG in our scripts.

faisal00813:
@vimmerru I have simplified the build scripts. Now a single script in root folder of each module (libindy and libnullpay) builds for all the architectures.

AxelNennker:
I switched out r16 and moved to r17 and rebuild using my scripts and had no problems doing that.
https://github.com/AxelNennker/indy-sdk/blob/android_builds/doc/android-build.md 
I think if you try to change the 16 to 17  and they still work then switch otherwise don't.

faisal00813:
Also addressed the concern regarding https://github.com/nsivraj/openssl_for_ios_and_android . Now we use a fork of https://github.com/nsivraj/openssl_for_ios_and_android which is in evernym account with a tag.

libindy/src/utils/environment.rs
faisal00813: Libindy will now panic! if it cannot find the EXTERNAL_STORAGE .

libindy/build.rs
```diff
@@ -32,4 +32,38 @@ fn main() {
             }
         }
     }
+		match &target {
```
jovfer: please clean-up code here. Seems like this section should be in `else if` for previous one (windows)

libindy/src/utils/environment.rs
jovfer: Agree with @vimmerru. Please fix all 3 points

libnullpay/Cargo.toml
```diff
@@ -6,7 +6,7 @@ build = "build.rs"
 
 [lib]
 name = "nullpay"
-crate-type = ["cdylib", "rlib"]
+crate-type = ["staticlib","cdylib", "rlib"]
```
jovfer: seems like staticlib should be android-specific target

libnullpay/build.rs
```diff
@@ -4,19 +4,29 @@ use std::path::Path;
 fn main() {
     let target = env::var("TARGET").unwrap();
     println!("target={}", target);
+    match &target {
```
jovfer: `match` construction is ugly here. please clean-up and use `if else`

libindy/Cargo.toml
```diff
@@ -66,7 +66,13 @@ lazy_static = "1.0"
...
 version = "0.5.0"
 default-features = false
-features = ["v4"]
\ No newline at end of file
+features = ["v4"]
+
+[target.'cfg(target_os = "android")'.dependencies]
+android_logger = "0.5"
+[target.'cfg(any(target_os = "android", target_os = "ios"))'.dependencies]
+rusqlite = { version = "0.13.0", features=["bundled"] }
```
jovfer: please specify common and android rusqlite dependence as neighbors to avoid partial update in the future

jovfer:
@faisal00813  After auto-merge test compilation is failed, please fix it also
https://ci.evernym.com/job/Indy%20SDK%20CI/view/change-requests/job/PR-777/8/execution/node/97/log/

libnullpay/Cargo.toml
faisal00813: I don't know if rust supports the conditional crate-type declaration. Does it? 
There are two issues related to it though, https://github.com/rust-lang/cargo/pull/5367 and https://github.com/rust-lang/cargo/issues/4900 .

PR #779, opened 2018-05-21
Fix lifetime issue for Wallet Search functionality

jovfer:


libindy/src/services/wallet/iterator.rs
```diff
@@ -9,17 +9,17 @@ use super::encryption::{decrypt_tags};
...
 
 
-impl<'a> WalletIterator<'a> {
-    pub fn new(storage_iter: Box<StorageIterator + 'a>, keys: &'a Keys) -> Self {
+impl WalletIterator {
+    pub fn new(storage_iter: Box<StorageIterator>, keys: &Keys) -> Self {
         WalletIterator {
             storage_iterator: storage_iter,
-            keys: keys
+            keys: keys.clone(),
```
keichiri: I know you did it because of lack of time, but I am thinking that maybe keys should be shared pointer (Rc)

Not to avoid unnecessary copying just, but also it might be safer to have keys at one place in memory, not all over the place possibly

jovfer: Exactly. In new one commit the one more same place with `TODO` marker is added. I will also put TODO here.

PR #787, opened 2018-05-23
Add negative tests for libnullpay

KitHat:


libnullpay/src/services/response_storage.rs
```diff
@@ -8,16 +8,19 @@ lazy_static! {
...
-    let req_id = get_val_from_obj(object, "reqId")?;
-    let req_id = val_to_u64(req_id)?;
-
+    let req_id = parse_req_id_from_request(request)?;
     let mut responses = RESPONSES.lock().unwrap();
     responses.insert(req_id.to_string(), response);
     Ok(())
 }
 
+pub fn parse_req_id_from_request(request: String) -> Result<u64, ErrorCode> {
```
vimmerru: Why this function borrows string? &str as input param seems better and allows to avoid unnecessary copying

vimmerru: Check also similar cases. It is more natural in Rust to provide slices instead of concrete container object or reference.

PR #795, opened 2018-05-24
fix wallet search when gt, gte, lt, lte and like are used with encryp…

keichiri:
Bugfix. Now error is properly returned when wrong type is used.

I also added tests for $like (that exist for other predicates) which were missing before


jovfer:
@Artemkaaas @vimmerru please review. CI is passed

PR #796, opened 2018-05-25
Fixed wrong encoded value of claim attribute

divacvl:
Fixed wrong encoded value of claim attribute in “Issue a Credential” How-To for Java, and added a comment how encoded value is actually computed.
Calculation of encoded value is discussed on [http://forum.sovrin.org/t/claim-values-encode/620/3], and the explanation should be present in How-To, so that claim different than the one in tutorial can be issued.

Signed-off-by: Vladan Divac <vladan.divac@evernym.com>


jovfer:
@divacvl could you fix DCO please? I'm not sure about exact reason, but one of the potential problem is inconsistency between commit author data and your github profile. Please add `vladan.divac@evernym.com` email to your GH account and update visible username to `Vladan Divac` as mentioned in your commit.

divacvl:
Hi,
I have updated GitHub profile as per the instructions.

PR #799, opened 2018-05-25
Implemented usage of pass phrase insted of base64 keys

Artemkaaas:
Signed-off-by: artem.ivanov <artem.ivanov@dsr-company.com>

PR #801, opened 2018-05-25
keys are now accessed via shared pointer instead of cloned (#798) + hotfix for anoncreds tests

jovfer:
#798


jovfer:
Fixes for CI are merged in #799

PR #802, opened 2018-05-25
Change python howtos to reference 'did' instead of 'signus'

jcace:
The module appears to be called 'did', not 'signus', so there is an error when trying to run the code for these howtos/tutorials. For instance, in the first howto:

``` 
File "write_did.py", line 17, in <module>
    from indy import pool, ledger, wallet, signus
ImportError: cannot import name 'signus' 
```

Also I added a small instruction to the first tutorial readme for how to install packages.


vimmerru:
@jcace Thanks for contribution. Our process require DCO checking for each PR. Here is info https://github.com/probot/dco#how-it-works. Could you update your PR according to DCO policy?

jcace:
Hi @vimmerru , thanks for the tip. Still getting used to contributing... I added the sign-off so I think it's good now! Thanks

PR #807, opened 2018-05-29
Pluged StateProof design.

jovfer:


doc/design/008-state-proof-pluggable-parsing/README.md
```diff
@@ -0,0 +1,127 @@
+## Legend
+There are some types of requests to Nodes in the Pool allowing to use StateProof optimization in Client-Node communication.
+Instead of sending requests to all nodes in the Pool, a client can send a request to a single Node and expect StateProof signed by BLS multi-signature.
+
+BLS multi-signature (BLS MS) guaranties that there was a consensus of Nodes which signed some State identified by State RootHash.
+StateProof (SP) is small amount of data which allows to verify particular values against RootHash.
+Combination of BLS MS and SP allows clients to be sure about response of single node is a part of State signed by enough Nodes. 
+
+## Goals
+Libindy allows to create and send custom transaction via plugged interface.
```
vimmerru: > Libindy allows to create and send custom transaction via plugged interface.

Not sure that it is 100% correct. More correct version:
1. Indy Node allows to extend supported requests with pluggins
2. libindy allows to send any request to node (we don't need plugins here), but some requests can be significantly optimized with signed state proofs.

doc/design/008-state-proof-pluggable-parsing/README.md
```diff
@@ -0,0 +1,127 @@
...
+Implementation of math for SP verification is a bit complicate for plugin logic.
+Therefore libindy should perform all math calculation inside.
+A plugin should provide handler to parse custom reply to fixed data structure.
+
+## API
+The signature of the handler is described below together with custom `free` call to deallocate result data.
+
+```rust
+extern fn custom_transaction_parser(reply_from_node: *const c_char, parsed_sp: *mut *const c_char) -> ErrCode;
+extern fn custom_free(data: *mut c_char) -> ErrCode;
```
lovesh: Which data is supposed to free? In `parsed_sp`?

doc/design/008-state-proof-pluggable-parsing/README.md
```diff
@@ -0,0 +1,127 @@
...
+``` 
+
+Libindy API will contain call to register handler for specific transaction type:
+```rust
+extern fn indy_register_transaction_parser_for_sp(command_handle: i32,
+                                                  txn_type: *const c_char,
+                                                  parser: custom_transaction_parser,
+                                                  free: custom_free,
+                                                  cb: extern fn(command_handle_: i32, err: ErrCode)) -> ErrCode;
```
lovesh: Plugin supplies `cb` and is called with the result of the SP verification so like a true/false or maybe some more elaborate `false`, right? Wouldn't the plugin want access to `reply_from_node` in `cb`

doc/design/008-state-proof-pluggable-parsing/README.md
```diff
@@ -0,0 +1,127 @@
...
+}
+
+/**
+ Subtrie variant of `KeyValuesInSP`.
+ 
+ In this case Client (libindy) should construct subtrie and append it into trie based on `proof_nodes`.
+ After this preparation each kv pair can be checked.
+*/
+struct KeyValuesSubTrieData {
+    sub_trie_prefix: Option<String>, /// common prefix of each pair in `kvs`. Should be used to correct merging initial trie and subtrie
```
lovesh: Why does it care that there is a prefix? The node returns the full keys as of now. It does not mention the prefix? Are you thinking of the optimisation that we planned for later?

doc/design/008-state-proof-pluggable-parsing/README.md
vimmerru: ```custom_transaction_parser```

This handler returns a null terminated string with result of request parsing by writing string pointer to memory located by ``````parsed_sp```` pointer.

Memory to store null terminated string will be allocated by plugin, so plugin should provide also handler for de-allocation of this memory.

doc/design/008-state-proof-pluggable-parsing/README.md
vimmerru: ```cb``` is called after libindy finished registration of this plugin. This operation is asynchronous and we need this ```cb``` to return result of registration operation. 

```parser: custom_transaction_parser,```` is the handler that plugin registers. You can find type ```custom_transaction_parser```  definition in previous lines. This handler consumes ```reply_from_node```

doc/design/008-state-proof-pluggable-parsing/README.md
jovfer: yes, SubTrie option is just reserve for future optimization. In the fist iteration all components (Node, libindy, plugins) will not support this variant.

doc/design/008-state-proof-pluggable-parsing/README.md
jovfer: yes, `parsed_sp` is out parameter for parser and memory will be allocated by plugin for this pointer. `custom_free` will de-allocate this buffer

doc/design/008-state-proof-pluggable-parsing/README.md
jovfer: this call is just API for registration of plugin's handlers. As any other API call of libindy it contains `cb` as last parameter to retrieve result of operation

lovesh: Ok, `err` in `cb` is the result of registration? How will the plugin know if `custom_transaction_parser` was able to verify the proof. So 
1. the plugin defines `custom_transaction_parser` that can extract SP from `reply_from_node` in `parsed_sp`.
2. indy-sdk verifies the SP, verification succeeds, then what does indy-sdk call? How will the plugin know if the verification was successful or indy-sdk is contacting another node for SP?

vimmerru: Plugin don't need to know anything about result of verification. Libindy will use provided data to verify SP and make decision about handling based on this info.

lovesh: Ok, so if the verification succeeds, sdk dispatches result to another plugin method?

vimmerru: No, we don't need plugin involvement anymore. The next logic is common for all requests.

PR #810, opened 2018-05-29
Add cache for transactions for future catch-ups

KitHat:


libindy/src/services/pool/catchup.rs
```diff
@@ -115,40 +118,67 @@ impl CatchupHandler {
...
-                    return Err(PoolError::CommonError(CommonError::InvalidState(
-                        "Local merkle tree greater than mt from ledger".to_string())));
+                    a => a
                 }
             }
         }
         Ok(CatchupProgress::InProgress)
     }
 
+    fn _try_to_catch_up(&mut self, most_popular_vote: &(String, usize, Option<Vec<String>>)) -> Result<CatchupProgress, PoolError> {
```
jovfer: I suggest to rename `most_popular_vote` to something like `confirmed_vote_for_ledger_status`. Not sure about exact name, but here it's not just most popular, now it's confirmed by consensus.

libindy/src/services/pool/catchup.rs
```diff
@@ -115,40 +118,67 @@ impl CatchupHandler {
...
+                return Err(PoolError::CommonError(CommonError::InvalidState(
+                    "Ledger merkle tree doesn't acceptable for current tree.".to_string())));
+            }
+        } else if target_mt_size > cur_mt_size {
+            self.target_mt_size = target_mt_size;
+            self.target_mt_root = target_mt_root.from_base58().map_err(|_|
+                CommonError::InvalidStructure(
+                    "Can't parse target MerkleTree hash from nodes responses".to_string()))?;
+            match hashes {
+                &None => (),
```
jovfer: missed consistency proof is something like protocol error. My personal vote to throw error for whole catchup process here
@vimmerru do you have another vision?

libindy/src/services/pool/mod.rs
```diff
@@ -357,38 +359,141 @@ impl PoolWorker {
...
+        let genesis_vec = PoolWorker::_genesis_to_binary(&p_genesis)?;
+
+        PoolWorker::_dump_vec_to_file(&genesis_vec, &mut file)
+    }
 
-        let f = fs::File::open(file_mame).map_err(map_err_trace!())?;
+    fn _dump_vec_to_file(v: &Vec<Vec<u8>>, file : &mut fs::File) -> Result<(), PoolError> {
+        v.into_iter().map(|vec| {
+            file.write_all(vec).map_err(map_err_trace!())?;
+            file.write_all("\n".as_bytes()).map_err(map_err_trace!())
```
jovfer: `\n` should not be used as delimiter if we are working with binary data

libindy/src/api/mod.rs
```diff
@@ -141,6 +141,9 @@ pub enum ErrorCode
     // Timeout for action
     PoolLedgerTimeout = 307,
 
+    // Timeout for action
```
jovfer: 1. Inconsistent description
1. Should be add to all libindy consumers: wrappers, plugin, cli

libindy/src/services/pool/catchup.rs
```diff
@@ -115,40 +118,67 @@ impl CatchupHandler {
...
-                let cur_mt_size = self.merkle_tree.count();
-                let cur_mt_hash = self.merkle_tree.root_hash().to_base58();
-                if target_mt_size == cur_mt_size {
-                    if cur_mt_hash.eq(target_mt_root) {
-                        return Ok(CatchupProgress::NotNeeded);
-                    } else {
-                        return Err(PoolError::CommonError(CommonError::InvalidState(
-                            "Ledger merkle tree doesn't acceptable for current tree.".to_string())));
+                return match self._try_to_catch_up(most_popular_vote) {
+                    //WARNING: If we receive invalid cache then it means that cache was cleared. We need to rebuild merkle tree and try to catch up once again
```
jovfer: I suggest to try to pop cache cleaning to this method. like
```
self._try_to_catch_up().or_else(|err| {
  drop_cache() // here or inside next if
  if err == InvalidCache { try_again() } else { err } 
})
```
It may be more clear and don't require WARNING comment

libindy/src/services/pool/mod.rs
```diff
@@ -391,6 +380,136 @@ impl PoolWorker {
...
+
+    pub fn drop_saved_txns(pool_name: &str) -> Result<(), PoolError> {
+        warn!("Cache is invalid -- dropping it!");
+        let mut p = EnvironmentUtils::pool_path(pool_name);
+
+        p.push("stored");
+        p.set_extension("btxn");
+        if p.exists() {
+            fs::remove_file(p).map_err(CommonError::IOError).map_err(PoolError::from)?;
+            Err(PoolError::InvalidCacheCleared)
```
jovfer: Right now all branches of the function's code return Err. May be we should return `Ok(())` in this case as actually cache dropping is successful?

libindy/src/errors/pool.rs
```diff
@@ -81,6 +85,7 @@ impl ToErrorCode for PoolError {
             PoolError::Terminate => ErrorCode::PoolLedgerTerminated,
             PoolError::Timeout => ErrorCode::PoolLedgerTimeout,
             PoolError::AlreadyExists(_) => ErrorCode::PoolLedgerConfigAlreadyExistsError,
+            PoolError::InvalidCacheCleared => ErrorCode::PoolLedgerInvalidCache,
```
jovfer: If it's internal error, may be we should not pass it into API.
@vimmerru sounds like we have first case when some error is actually internal. May be we can map this case to CommonInvalidState or even to `unreachable` macros

libindy/test.txt
```diff
@@ -0,0 +1,111 @@
+Pre-trans
```
jovfer: please remove it

PR #811, opened 2018-05-29
wallet key rotation

keichiri:
Signed-off-by: Janko Krstic <keichiri@protonmail.com>

### DIFF
--- a/libindy/src/services/wallet/mod.rs
+++ b/libindy/src/services/wallet/mod.rs
@@ -30,6 +30,7 @@ use self::storage::default::SQLiteStorageType;
 use self::storage::plugged::PluggedStorageType;
 use self::wallet::{Wallet, Keys, Tags};
 use self::indy_crypto::utils::json::{JsonDecodable, JsonEncodable};
+use utils::crypto::pwhash_argon2i13::PwhashArgon2i13;
 
 
 #[derive(Serialize, Deserialize, Debug)]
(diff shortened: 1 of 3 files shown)


libindy/src/services/wallet/wallet.rs
```diff
@@ -239,6 +252,12 @@ impl Wallet {
         Ok(())
     }
 
+    pub(super) fn rotate_key(&self, new_master_key: &[u8]) -> Result<(), WalletError> {
+        let new_keys = self.keys.encrypt(new_master_key);
```
jovfer: `new_keys` is misleading here, please rename it. Actually keys are not changed, only encrypted representation are changed.

libindy/src/services/wallet/mod.rs
```diff
@@ -1511,4 +1542,39 @@ mod tests {
...
+    #[test]
+    fn wallet_service_key_rotation() {
+        _cleanup();
+
+        let wallet_service = WalletService::new();
+        wallet_service.create_wallet("pool1", "test_wallet", None, None, &_credentials()).unwrap();
+        let wallet_handle = wallet_service.open_wallet("test_wallet", None, &_credentials()).unwrap();
+
+        wallet_service.add_record(wallet_handle, "type", "key1", "value1", "{}").unwrap();
+        let record = wallet_service.get_record(wallet_handle, "type", "key1", &_fetch_options(false, false, false)).unwrap();
```
jovfer: I think it's better to fetch something and check that after key rotation fetched value is the same

libindy/src/services/wallet/mod.rs
```diff
@@ -63,22 +64,30 @@ impl<'a> JsonDecodable<'a> for WalletConfig {}
 #[derive(Debug)]
 pub struct WalletCredentials {
     master_key: [u8; 32],
+    rekey: Option<Box<[u8; 32]>>,
```
jovfer: Is boxing really required here?

keichiri: I dont want to increase the size of WalletCredentials by 32 bytes since 99% of the time it will be empty

libindy/src/services/wallet/mod.rs
keichiri: yes, that is what asserts below do.

libindy/src/services/wallet/wallet.rs
keichiri: yes, new_keys might be misleading. I agree

libindy/src/services/wallet/mod.rs
jovfer: I see only asserts like `record.get_value().is_none()`
It's better to have something like
```
...
wallet.add_record(..., value1_const, ...)
...
assert_eq!(fetched_value, Some(value1_const));
...
rotate_keys();
...
assert_eq!(fetched_value, Some(value1_const));
```

libindy/src/services/wallet/mod.rs
vimmerru: @jovfer 
@keichiri 

I see the following problems here:

- 32 is a magic number
- Memory allocated for Keys should be zeroed after usage. See any sodiumoxide Key type
- Box here looks like non-obvious micro-optimization.

libindy/src/services/wallet/mod.rs
keichiri: oh yes you are right. I brainfarted and reused wrong test. I will change.My bad

libindy/src/services/wallet/mod.rs
keichiri: This (except the Box) is the same approach used for master key in credentials. 
So I think if it is an issue now, it must have been before, since it was implemented.

About Box, why is it nonobvious? It is totally normal practice to use memory smartly?

Ok, i will change all of this as soon as possible

jovfer: agree with Janko about the 2nd point from Slava's review. It's something like out of scope of this PR. This issue was introduced in first implementation of encrypted wallet and wasn't fixed while latest large refactoring from Wallet to Wallet+Storage.
As for Box, this is really micro-optimization. WalletCredential used as single object. 24 bytes vs minor improvement of readability - I prefer doesn't think why Box is needed here, but don't used for the first filed of the structure

keichiri: I removed Box. IT was not used for first field since it is mandatory, always present.

But I agree its not necessary so I removed it.

Anyway, about 2nd point. I agree that it is very needed, and I will open an additional PR soon for it.

I will push the change now

PR #706, opened 2018-05-02
Added a Contributor Resources page and a how to for signing commits. 

michaeldboyd:
- Added a doc similar to the indy-node MAINTAINERS.md that welcomes new contributors and describes the steps needed to get involved in the community and submit PRs
- Added a doc describing how to sign commits and optionally use GPG to verify their signatures. We have had many people misunderstand how DCO works, and this hopefully will help people more easily contribute to indy-sdk

PR #812, opened 2018-05-29
IS-707: Make CI fail on warnings

stevengubler:
# [IS-707](https://jira.hyperledger.org/browse/IS-707)

To be in compliance with CII certification standards, we must address all warnings from the compiler. To ensure this, we are rejecting any build warnings in our continuous integration.

Documentation of this change can be added once #706 (or similar) is merged. Refer to [IS-739](https://jira.hyperledger.org/browse/IS-739) for this change.


stevengubler:
As a sidenote, I approved this change with Nathan George, so let's please involve him on any discussion on whether or not this should be our strategy to comply with CII standards.

cli/src/main.rs
```diff
@@ -1,3 +1,6 @@
+// If compiling with feature "fatal_warnings", the build will fail on all warnings
```
jovfer: I'm not sure, what this line is the best place for this comment. May be we should move it to `Cargo.toml` file?
@vimmerru ^

Jenkinsfile.ci
```diff
@@ -1,5 +1,8 @@
 #!groovy
 
+def features = "--features fatal_warnings"
+def buildCommand = "cargo build ${features}"
```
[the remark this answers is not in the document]
stevengubler: `buildCommand` was initially a replacement for all entries, but when I pulled master, I had to change it up a little bit in order to get it to work with 
```groovy
def libindyFeatures = "--features sodium_static"
...
sh "cargo build $libindyFeatures"
```

I guess I'm not sure how to reconcile the specific needs of this specific build case and still abstract out our build command. Should I just remove the variable entirely, or is there a better way to do this?

cli/src/main.rs
stevengubler: I like that idea. Makes a lot more sense. Should I make the change or wait for @vimmerru's input?

stevengubler:
Alright, I made the changes you asked for. Anything else?

jovfer:
@vimmerru @Artemkaaas @KitHat please remove warnings in current master and merge this PR

PR #816, opened 2018-05-30
Fixed wrong encoded value of claim attribute in “Issue a Credential” …

jovfer:
…How-To for Java, and added a comment how encoded value is actually computed.

Signed-off-by: Vladan Divac <vladan.divac@evernym.com>


jovfer:
This is temporary PR to experiments about DCO check

PR #832, opened 2018-06-04
Feature Plugged StateProof parsing

jovfer:


libindy/src/api/ledger.rs
```diff
@@ -1616,4 +1616,45 @@ pub extern fn indy_parse_get_revoc_reg_delta_response(command_handle: i32,
...
     res
-}
\ No newline at end of file
+}
+
+pub type CustomTransactionParser = extern fn(reply_from_node: *const c_char, parsed_sp: *mut *const c_char) -> ErrorCode;
+pub type CustomFree = extern fn(data: *const c_char) -> ErrorCode;
+
+#[no_mangle]
+pub extern fn indy_register_transaction_parser_for_sp(command_handle: i32,
```
vimmerru: We need to have documentation for each method/type

libindy/src/api/ledger.rs
```diff
@@ -1616,4 +1616,45 @@ pub extern fn indy_parse_get_revoc_reg_delta_response(command_handle: i32,
...
+                                                      pool_handle: i32,
```
vimmerru: I am not sure that we need pool_handle here. Let's discuss.

libindy/src/api/ledger.rs
```diff
@@ -1616,4 +1616,45 @@ pub extern fn indy_parse_get_revoc_reg_delta_response(command_handle: i32,
...
+                                                      txn_type: *const c_char,
+                                                      parser: Option<CustomTransactionParser>,
```
vimmerru: I prefer to have Option inside of type definition. Otherwise user can get a feeling that this parameter is optional

libindy/src/commands/ledger.rs
```diff
@@ -212,6 +225,16 @@ impl LedgerCommandExecutor {
                     .expect("Expect callback to process ack command")
                     (result.map_err(IndyError::from));
             }
+            LedgerCommand::RegisterParserSP(pool_handle, txn_type, parser, free, cb) => {
+                info!(target: "ledger_command_executor", "RegisterParserSP command received");
+                self.register_parser(pool_handle, &txn_type, parser, free, cb);
```
vimmerru: Do we need to call it register_parser_for_sp instead of register_parser ???

libindy/src/commands/ledger.rs
```diff
@@ -335,6 +358,19 @@ impl LedgerCommandExecutor {
         };
     }
 
+    fn register_parser(&self, pool_handle: i32, txn_type: &str,
+                       parser: CustomTransactionParser, free: CustomFree,
+                       cb: Box<Fn(Result<(), IndyError>) + Send>) {
+        debug!("register_parser >>> pool_handle: {:?}, txn_type: {:?}, parser: {:?}, free: {:?}",
+               pool_handle, txn_type, parser, free);
+
+        let x: Result<i32, PoolError> = self.pool_service.register(pool_handle, txn_type, parser, free);
```
vimmerru: I suggest to rename "register" to something more obvious

libindy/src/commands/ledger.rs
```diff
@@ -335,6 +358,19 @@ impl LedgerCommandExecutor {
...
+        match x {
+            Ok(cmd_id) => { self.register_callbacks.borrow_mut().insert(cmd_id, cb); }
```
vimmerru: "register_callbacks" seems non-obvious name.

libindy/src/services/pool/mod.rs
```diff
@@ -277,7 +278,21 @@ impl PoolWorker {
                             })
                     })?;
                 }
-                &ZMQLoopAction::Timeout => {
+                ZMQLoopAction::Register(cmd_id, txn_type, parser, free) => {
+                    match self.handler {
+                        PoolWorkerHandler::TransactionHandler(ref mut th) => {
+                            th.parser_sps.insert(txn_type,
```
vimmerru: We need to store "parser_sps" on a service layer. All opened pools must handle this transaction type.

libindy/src/services/pool/transaction_handler.rs
```diff
@@ -288,37 +271,67 @@ impl TransactionHandler {
...
             return None;
         };
 
-        if !REQUESTS_FOR_STATE_PROOFS.contains(&xtype) {
-            //TODO GET_DDO, GET_TXN
-            trace!("TransactionHandler::parse_reply_for_proof_checking: <<< type not supported");
-            return None;
+        if REQUESTS_FOR_STATE_PROOFS.contains(&type_) {
+            trace!("TransactionHandler::parse_generic_reply_for_proof_checking: built-in");
+            TransactionHandler::parse_reply_for_builtin_sp(json_msg, type_)
```
vimmerru: > TransactionHandler::parse_reply_for_builtin_sp(json_msg, type_)

I don't like this approach. I prefer to have n dedicated functions that will be called for each transaction type without any distinction between "embedded" and "external" sp parser.

libindy/src/api/ledger.rs
jovfer: We always use Option around custom callback type. We can change our approach, but I don't like to mix different in codebase. So after additional discussion we can raise separate PR for API changes (this, payments, wallet, etc)

libindy/src/services/pool/transaction_handler.rs
jovfer: I prefer to do it in separate PR...

libindy/src/services/pool/mod.rs
jovfer: Related to https://github.com/hyperledger/indy-sdk/pull/832#discussion_r192691740
I prefer to have more flexibility here.
The case when application use more then 1 pool is something unusual and required by complex logic. For complex cases it's better to have flexible approach

libindy/src/api/ledger.rs
jovfer: https://github.com/hyperledger/indy-sdk/pull/832#discussion_r192704404

libindy/src/services/pool/transaction_handler.rs
```diff
@@ -539,10 +559,54 @@ impl TransactionHandler {
...
+                return false;
+            }
+
+            let proof_nodes = unwrap_or_return!(base64::decode(&parsed_sp.proof_nodes), false);
+            let root_hash = unwrap_or_return!(parsed_sp.root_hash.from_base58(), false);
+            match parsed_sp.kvs_to_verify {
+                KeyValuesInSP::Simple(kvs) => {
+                    for (k, v) in kvs.kvs {
+                        let key = unwrap_or_return!(base64::decode(&k), false);
+                        if !state_proof::verify_proof(proof_nodes.as_slice(),
```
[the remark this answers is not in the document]
jovfer: +1
but again (https://github.com/hyperledger/indy-sdk/pull/832/files/a82bf9e9f23c135eea448ce0242a77748b5ad312#r192699456) I would like to merge minimal PR to master and make API available for usage in plugin.

libindy/src/services/pool/mod.rs
```diff
@@ -783,6 +789,22 @@ impl PoolService {
         Ok(cmd_id)
     }
 
+    pub fn register_sp_parser(txn_type: &str,
+                              parser: CustomTransactionParser, free: CustomFree) -> Result<(), PoolError> {
+        REGISTERED_SP_PARSERS.lock()
+            .map(|mut map| {
```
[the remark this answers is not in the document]
jovfer: I think it also point to discuss on libindy API level. Should libindy override already registered txn types or not?
@vimmerru @lovesh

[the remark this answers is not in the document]
jovfer: After discussion with @vimmerru I've added check to doesn't allow overriding parser for default txn types and allow silent overriding previously registered custom type from API level.

PR #837, opened 2018-06-05
tell _which_ did already exists

AxelNennker:

The information that the did already exists is already in the name of the error DidError::AlreadyExistsError so adding the string "Did already exists" does not provide additional info.
Replacing it with the did.to_string() to provide that info to the caller.


vimmerru:
@AxelNennker Could you solve conflict and remove WIP tag after it?

vimmerru:
Seems merging caused absence of any changes.

AxelNennker:
Strange. Will look into why there are not file changed later. Maybe the change sneeked into some other PR already?

jovfer:
@AxelNennker seems like during some changes in master Did structure is changed. Not build failed with message:
```
error[E0599]: no method named `to_string` found for type `domain::crypto::did::Did` in the current scope
   --> src/commands/did.rs:220:77
    |
220 |             return Err(IndyError::DidError(DidError::AlreadyExistsError(did.to_string())));
    |                                                                             ^^^^^^^^^
    | 
   ::: src/domain/crypto/did.rs:37:1
    |
37  | pub struct Did {
    | -------------- method `to_string` not found for this
    |
    = note: the method `to_string` exists but the following trait bounds were not satisfied:
            `domain::crypto::did::Did : std::string::ToString`
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following trait defines an item `to_string`, perhaps you need to implement it:
            candidate #1: `std::string::ToString`
```
Could you fix it, please?

AxelNennker:
@jovfer fixed it. Sorry, no idea how this happened.

PR #838, opened 2018-06-05
[DESGIN] enhancements of pool connection logic

vimmerru:
This design proposes enhancements of pool connection logic in libindy


doc/design/009-efficient-connections/README.md
```diff
@@ -0,0 +1,44 @@
...
+The main idea of proposal is force libindy to close sockets as soon as possible, but
+still provide very limited keep-alive ability to avoid unnecessary CurveCP and TCP
+handshakes:
+
+1. Change ```indy_pool_open``` endpoint behavior. Instead keep sockets connected it will only perform CatchUp
+   and keep only information about nodes of this pool. Sockets creation will be performed when application
+   sends request with some limitations and optimizations (see details below).
+1. Introduce internal "pool connection" entity. Each "pool connection" is intended to be used with a specific pool only.
+1. "Pool connection" owns sockets and can be used to execute one or multiple requests.
+1. After creation "pool connection" becomes "active" for some pre-defined timeout (5 sec). "Active" means that context can be used to start execution of     new requests to pool.
```
ashcherbakov: It needs to be active not only based on timeout, but timeout AND a max number of sent requests

doc/design/009-efficient-connections/README.md
```diff
@@ -0,0 +1,44 @@
...
+Note that problem is complex and requires corresponded solution in Node codebase and
+pool network infrastructure, but this proposal is focused on libindy (client) side.
+
+## Proposed changes
+
+The main idea of proposal is force libindy to close sockets as soon as possible, but
+still provide very limited keep-alive ability to avoid unnecessary CurveCP and TCP
+handshakes:
+
+1. Change ```indy_pool_open``` endpoint behavior. Instead keep sockets connected it will only perform CatchUp
```
ashcherbakov: Also the results of catch-up need to be persisted, so we don't do catch-up from scratch on every `indy_pool_open`

doc/design/009-efficient-connections/README.md
```diff
@@ -0,0 +1,44 @@
...
+1. "Pool connection" owns sockets and can be used to execute one or multiple requests.
+1. After creation "pool connection" becomes "active" for some pre-defined timeout (5 sec). "Active" means that context can be used to start execution of     new requests to pool.
+1. When application tries to send request to pool libindy checks for already exists "pool connection" for target pool in "active" state.
+1. If there is no active "pool connection" then libindy creates a new one and uses it for sending request.
+1. If there is active "pool connection" then libindy re-uses it for sending request.
+1. "pool connection" opens sockets only "by request". if requests execution requires connection to only one node than only one socket will be created.
+1. "pool connection" determines node to perform new connection with round robin. If connection is already established than sockets will be re-used.
+1. libindy keep opened sockets connected until "pool connection" is active (5s from "pool connection" creation) or there is request that waits for
+   response on this socket. As soon as "pool connection" is switched from active state and sockets isn't needed for request it will be
```
ashcherbakov: Please mention that it waits for responses for some timeout, not infinitely. So the sockets will always be closed eventually.

doc/design/009-efficient-connections/README.md
vimmerru: I added this info

vimmerru:
@dhh1128 

> I've read the proposal and it all makes sense to me. I only have one question. It seems that you are proposing to keep a connection open for 5 seconds after creation. If the connection is used actively, would that reset the timeout (so we're really keeping it open for 5 seconds after last use), or would you close it after 5 seconds even if it was actively used at 4.9 seconds after creation?

The main idea is prohibit sending of new requests through this connection after 5 seconds from creation. not from the time of sending the last request.

doc/design/009-efficient-connections/README.md
```diff
@@ -0,0 +1,49 @@
...
+Note that problem is complex and requires corresponded solution in Node codebase and
+pool network infrastructure, but this proposal is focused on libindy (client) side.
+
+## Proposed changes
+
+The main idea of proposal is force libindy to close sockets as soon as possible, but
+still provide very limited keep-alive ability to avoid unnecessary CurveCP and TCP
+handshakes:
+
+1. Change ```indy_pool_open``` endpoint behavior. Instead keep sockets connected it will only perform CatchUp
```
lovesh: This is an idea for future; We dont support catchup using state proofs but a small ledger like pool ledger (maybe config too if we need it) should be done using state proofs, since it can be done with just one node and in one round trip (node sends the whole pool state in a single request with the signed root and since the signed root has a timestamp the sdk can check if its new enough)

doc/design/009-efficient-connections/README.md
```diff
@@ -0,0 +1,49 @@
...
+1. Persist ```indy_pool_open``` CatchUp results to start from updated pool ledger next time.
+1. Introduce internal "pool connection" entity. Each "pool connection" is intended to be used with a specific pool only.
+1. "Pool connection" owns sockets and can be used to execute one or multiple requests.
+1. After creation "pool connection" becomes "active" for some pre-defined timeout (5 sec) or until some pre-defined amount
+   of requests (5) were started through this "pool connection".
+1. "Active" means that context can be used to start execution of     new requests to pool.
+1. When application tries to send request to pool libindy checks for already exists "pool connection" for target pool in "active" state.
+1. If there is no active "pool connection" then libindy creates a new one and uses it for sending request.
+1. If there is active "pool connection" then libindy re-uses it for sending request.
+1. "pool connection" opens sockets only "by request". if requests execution requires connection to only one node than only one socket will be created.
```
lovesh: "requires connection to only one node", you mean only read request expecting state proof. Also currently libindy tries to connect to all known nodes but we need to talk to only `>f` nodes? Would that be changed too?

lovesh:
> The main idea is prohibit sending of new requests through this connection after 5 seconds from creation. not from the time of sending the last request.

But shouldn't you be increasing the timeout by say 1 sec (less than connection lifetime) on every sent request since a client maybe actively interacting

doc/design/009-efficient-connections/README.md
ashcherbakov: As I understand yes. If this is a get request, then a connection to one node only will be established.

ashcherbakov:
> But shouldn't you be increasing the timeout by say 1 sec (less than connection lifetime) on every sent request since a client maybe actively interacting
No, otherwise the goal (make sure that client connections are closed periodically) is not achieved.

lovesh:
@ashcherbakov The goal is not to close client connections periodically but on inactivity.

ashcherbakov:
@lovesh No. The goal is to close it periodically, because ZMQ can not close it from the node side. So, if there are malicious (or even valid) active clients sending requests from time to time, then the no more new clients can be connected. 
You can find more details here: https://docs.google.com/document/d/1XghaJtCrBmql-HdhNtKyof-MTaZUVJ5akJ-Tjlc2bfg/edit#

PR #842, opened 2018-06-06
Hotfix/dotnet fixes

srottem:
Fixes to revert the target framework to the minimum standard required to ensure the widest possible adoption and documentation corrections to match the namespaces changed in recent commits.


srottem:
I'm given to understand that the version being released next is 1.5.0.  Should the .NET assemblies and packaging be set with the same version number?

vimmerru:
@srottem Yes, It is better to use the same versions for the moment

PR #846, opened 2018-06-07
Check if wallet is not opened before deleting

keichiri:
Signed-off-by: Janko Krstic <keichiri@protonmail.com>


libindy/src/api/mod.rs
```diff
@@ -116,6 +116,9 @@ pub enum ErrorCode
     // Returned if provided wallet query is invalid
     WalletQueryError = 214,
 
+    // Returned if wallet it not closed, which is required for the operation (currently for deletion)
+    WalletNotClosed = 215,
```
vimmerru: Could we just change the error to CommonInvalidState? It will correspond to behavior with deletion of pool and we will not need to update all wrappers and cli

keichiri: Sure no problem

keichiri:
I will so the error maps to CommonInvalidState.
Although my understanding is that it is error used when there is something wrong with indy sdk's internal state. Something is corrupted or there is bug which created invalid state.
This is more of a user error how I see it, but I will change it

libindy/src/errors/wallet.rs
```diff
@@ -134,7 +134,7 @@ impl ToErrorCode for WalletError {
             WalletError::ItemNotFound => ErrorCode::WalletItemNotFound,
             WalletError::ItemAlreadyExists => ErrorCode::WalletItemAlreadyExists,
             WalletError::QueryError(_) => ErrorCode::WalletQueryError,
-            WalletError::NotClosed => ErrorCode::WalletNotClosed,
+            WalletError::NotClosed => ErrorCode::CommonInvalidState,
```
vimmerru: I suggest to remove WalletError::NotClosed at all, not just use different error code mapping. Mapping of errors should be one to one.

keichiri:
Alright. I removed NotClosed error completely. Although I still am not convinced that this is error that should be considered InvalidState. 
I do think that very granular errors like NotClosed is not good approach, since you might end up with hundreds errors in the end. But maybe new generic errors that would indicate error in the sequence of operations/state machine that is not valid can be added.
Or clarify what exactly InvalidState means, since it seems it is overloaded now

PR #850, opened 2018-06-07
added data structure changes to documentation

cam-parra:
As agreed on the Payment API needed to reflect a version number and an address in certain data structures.
Signed-off-by: cam-parra <caeparra@gmail.com>

### DIFF
--- a/doc/design/004-payment-interface/README.md
+++ b/doc/design/004-payment-interface/README.md
@@ -441,11 +441,16 @@ pub extern fn indy_add_request_fees(command_handle: i32,
 ///
 /// #Returns
 /// utxo_json - parsed (payment method and node version agnostic) utxo info as json:
-///   [{
-///      input: <str>, // UTXO input
-///      amount: <int>, // amount of tokens in this input
-///      extra: <str>, // optional data from payment transaction
-///   }]
+/// {
+///     "ver": <int>,                    // this field is included to allow for future backward compatability
+///     "utxo_json":
+///         [{
+///           "paymentAddress": <str>,// payment address: "pay:<currency>:<address><checksum>"
+///           "txo": <str>,          
+///           "amount": <int>,        // amount of tokens in this input
+///           "extra": <str>          // optional data from payment transaction
+///         }]
+/// }
 pub extern fn indy_parse_response_with_fees(command_handle: i32,
                                             payment_method: *const c_char,
                                             resp_json: *const c_char,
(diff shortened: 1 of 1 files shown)


doc/design/004-payment-interface/README.md
```diff
@@ -441,11 +441,16 @@ pub extern fn indy_add_request_fees(command_handle: i32,
 ///
 /// #Returns
 /// utxo_json - parsed (payment method and node version agnostic) utxo info as json:
-///   [{
-///      input: <str>, // UTXO input
-///      amount: <int>, // amount of tokens in this input
-///      extra: <str>, // optional data from payment transaction
-///   }]
+/// {
+///     "ver": <int>,                    // this field is included to allow for future backward compatability
```
vimmerru: I really don't understand how adding version to output param of public API can help in backward compatibility? Could you provide an example?

brentzundel: If for some reason the format of the txo string must change, the version would allow us to be compatible with old format. Same reason as adding version to ledger transaction.

vimmerru: Consumer of this API is 3d party application. What application will do with this version?

vimmerru: Do you expect that application should have the code that will work with different versions?

vimmerru: Also what app should do with unknown version (it supports v1 and v2, but received v3 after libindy update)? Crash?

brentzundel: Our thinking is that the version number allows LibSovToken to make non-breaking changes. Updates to libindy should not affect version number, only updates to libsovtoken.
The version number allows the 3rd party app to use saved txo strings from previous versions of libsovtoken in a non-breaking way.
Is there is a better way to do this? Or do you think this does not need to be done?

vimmerru: If you want "txo strings" to be compatible than you need to add version to "txo string". How adding the version to utxo_json can help with this? 

Also you don't need the version  in "txo string" right now, as absence of version means version 0. 

For example:

```
txo:sov:asdasdasd@1 - version 0
txo:sov:asdasdasd@1_v1 - version 1
```

vimmerru: Format of utxo_json also isn't under libsovtoken control at all. All you can control is format of txo strings

brentzundel: That is a good point. If we want to version the txo strings, we should do so inside of them. I agree that version should not be a part of utxo_json for libsovtoken.
Do you think that there may be a need for version number as part of utxo_json for libindy? what if structure of uxto_json changes in breaking way?

vimmerru: To be backward compatible libindy should be backward compatible in out params. It means that utxo_json can only be changed by adding new fields or by providing new similar method, but with extended behavior. Adding version to utxo_json doesn't help. So i suggest to don't add the version here.

We can think about adding some versioning in interface between libsovtoken and libindy only,

vimmerru:
@brentzundel  asI unerstand we don’t need this PR.  I am closing. You can reopen it if needed.

PR #851, opened 2018-06-07
do not loose zmqerror information

AxelNennker:

before this PR all zmqerrors were converted to IOError so information was lost and zmqerror was not exposed to the api.
LibIndy api call can result in IO errors but not all of them are zmq errors.

This PR allows the caller of the libindy api to know the difference between a zmq error and io errors


vimmerru:
It significantly contradicts with our vision of error handling. The only reason for new error code is ability to be properly recovered by application and not just for debugging. ZMQ is temporary solution and you breaks encapsulation here.

AxelNennker:
Hm, io error can be anything like a file write error. Maybe the name zmq error is wrong if zmq is only a temporary solution. What would this error class be called while still distinguishing it from some other IO error?

vimmerru:
What problem do you want to solve? How you plan to recovery in your app?

AxelNennker:
I think that a random io error is most likely something the app can't recover. A zmq error might be more likely to recover from. An io error is more likely an internal error of libindy, I think.

The general problem is better error messages. A random io error might not be recovered from while some zmq error fall into the category 'Please try again later'.

AxelNennker:
What I think is wrong with the original code is that zmqerror is mapped to common ioerror while all zmq errors are in fact pool errors. So, I think that there should be some PoolError::SecureTransportError that zmqerrors are mapped to. And that new error is visible through the API.

AxelNennker:
I think the new PR reflects better what I wanted to achieve. Thanks for making me think. :-)

vimmerru:
zmq raises errors very rarely when something is very wrong with networking on edge device. Most probably there will be no way to recover properly. 

> The general problem is better error messages. 

IOError contains error message as string. The only problem is that for now we don't expose this string in our API and we need the backward compatible way to expose messages. One idea that we have is add ```indy_get_error(command_handle)``` endpoint that can be called if app needs better error message.

AxelNennker:
I think zmq errors should not be munched into general ioerrors making them indistinguishable from other io errors. 
zmq errors are always pool/ledger errors but there is currently no way to tell that the app.
PoolLedgerSecureTransportError sounds like a fit.

Maybe there is a need for a 'something is very wrong and there is nothing you can do' error like 'HTTP 500'.

Axel

(LedgerNoConsensusError should be rare too and is a sign that something is very wrong and there is not much the app can do then except telling the user to call the hotline so that somebody starts kicking the ledger.)

vimmerru:
> zmq errors are always pool/ledger

These errors aren't related to pool/ledger at all. We don't expose network timeouts or bad answer from the node as IOError. Only error like "no network interface present".

AxelNennker:
Could you please point me to a file that uses zmq that is not pool.
I can't find on

Axel

ignisvulpis@namenlos:~/development/hyperledger/indy-sdk/libindy/src$ find . -name \*\.rs -exec fgrep 'crate zmq' {} \; -print
extern crate zmq;
./services/pool/mod.rs
extern crate zmq;
./errors/common.rs
extern crate zmq;
./errors/pool.rs

vimmerru:
> Could you please point me to a file that uses zmq that is not pool.

zmq is used only inside PoolService.

AxelNennker:
if zmq is only used inside PoolService then we much zmq::Error into an CommonError::IOError and not into a PoolError?

vimmerru:
Because they have completely different semantics. Pool errors are errors related to consensus, incompatible merkle tree and similar problems. They are io agnostic. IOError is error like "busy network interface" or "non-existing file".

AxelNennker:
I am not convinced and still think that the cause of the error should not be lost. Although many zmq error are io errors as their root cause others are not. There are json decode error and mismatching crypto algs and mismatching key length etc which are not io error.

Anyway I am closing this. Thanks for the discussion.

vimmerru:
If you see the cases were incorrect error is raised i suggest to discuss as it is very important question. Do you see places where ```json decode error and mismatching crypto algs``` returned is IOError?

AxelNennker:
I can't find the json decode error or crypto algs errors anymore. Sorry.

I searched through my ~/.cargo/.../zmq-*/**/rs files for error. Most of them are io error but not all.

Inside zmq there is e.g. DecodeError
https://github.com/erickt/rust-zmq/blob/master/src/lib.rs#L1163
This should be rare but could pop-up in libindy. Maybe it can never happen because libindy does not use that part of zmq.

zmq calls a few errors 'native zmq errors':
https://github.com/erickt/rust-zmq/blob/fde336758084a5c05f1d9a7e0753049d37b9f88e/zmq-sys/src/errno.rs#L38
https://github.com/erickt/rust-zmq/blob/7bfd3bacca2fbf4ef32b5fa0160d52af85793402/src/lib.rs#L230
    // native zmq error codes
    EFSM            = errno::EFSM as isize,
    ENOCOMPATPROTO  = errno::ENOCOMPATPROTO as isize,
    ETERM           = errno::ETERM as isize,
    EMTHREAD        = errno::EMTHREAD as isize,


There is probably nothing the app using libindy can do - still it feels that PoolLedgerSecureTransportError is 'more correct' than io error.
https://github.com/AxelNennker/indy-sdk/blob/zmqerror/libindy/src/errors/pool.rs#L32

I guess that if zmq is replaced by e.g. mutual TLS or something then other errors will pop up in pool/ledger.

vimmerru:
I am sorry, you just proposes now non-backward compatible changes that all Indy apps developers, Indy wrapper developers, Indy cli developers must adopt, but motivation doesn't seem so strong for such big change.

AxelNennker:
That's why I closed the PR. I agree that this is probably not worse the trouble but maybe later when a change is needed for other reasons...

PR #854, opened 2018-06-08
[WIP] match ref patterns

AxelNennker:
Just something on style. Some claim that this style is easier to read and understand.

I guess this PR is not _needed_ but I think it is good advice to follow one code style.

For further information visit:
[Clippy sez](https://rust-lang-nursery.github.io/rust-clippy/current/index.html#match_ref_pats)


keichiri:
I am curious, what are the benefits of this approach?

AxelNennker:
I am not saying that one way to do this is better than the other.
It seems that Clippy and rustfix agree on this style.
Currently libindy code does not sometimes one way and sometime the other way.

I am using the tools to learn indy-sdk and Rust and have no strong opinion one way or the other but maybe the Clippy people have a point and having one style feels the right thing to do.

vimmerru:
> I am curious, what are the benefits of this approach?

As i understand to satisfy Clippy linter. 

@AxelNennker 

Do you think Clippy is mature enough to be involved deeply to our development process?

AxelNennker:
At first I would set it up in a way that there is a warning if new code adds new Clippy warnings. I would not fail a PR for that.

Related but different:
I did a rustfmt some days ago on all libindy Rust files. rustfmt has initial Travis configuration... 

Both put an extra hurdle on new developers to contribute so not sure about all this. Developers paid to develop on indy, especially if they work for the same company, might be easier to convince - or not...

Being dogmatic on style might improve readability and some warnings by those tools might actually help find programming errors. YMMW

jovfer:
@vimmerru @AxelNennker Do we have any conclusion for this PR?

AxelNennker:
[clippy is available as a rustup component](https://internals.rust-lang.org/t/clippy-is-available-as-a-rustup-component/7967) from July. Seems Clippy is maturing. Should be in 1.29 or 1.20

AxelNennker:
I am closing this as resolving the conflicts with the current master is more work than this is worth.

PR #862, opened 2018-06-09
typedef Result for IndyError

AxelNennker:

As said here it is common practise to wrap errors and to improve readability by using a type definition for our errors e.g.: type Result<T> = result::Result<T, IndyError>;
https://doc.rust-lang.org/rust-by-example/error/multiple_error_types/wrap_error.html

The libindy code has the mentioned 'boilerplate' code already.

I think we can use '?' more often too
https://doc.rust-lang.org/rust-by-example/error/multiple_error_types/reenter_question_mark.html


AxelNennker:
Are the places this PR changed OK and it can be merged?

Axel

Related to this I am wondering Why does
SubmitAck(
i32, // cmd_id
Result<String, PoolError>, // result json or error
),
return a PoolError and not an IndyError?
in file https://github.com/hyperledger/indy-sdk/blob/master/libindy/src/commands/ledger.rs#L45

vimmerru:
> Related to this I am wondering Why does
> return a PoolError and not an IndyError?

Because it is Ack from Pool service, not request from the user

vimmerru:
> Are the places this PR changed OK and it can be merged?

It is ok, we will merge it, but most probably in a few days as we have a lot of important PRs in merge queue

vimmerru:
@AxelNennker Could you solve conflict and remove WIP tag after it?

vimmerru:
@AxelNennker Could you solve conflict and remove WIP tag after it?

AxelNennker:
I resolved the merge conflict. It stemmed from the introduction of two new methods import and export.

PR #864, opened 2018-06-11
IS-753: Improve Write a DID and Query its Verkey how-to in Python

trthhrtz:
Minor corrections, add common errors, make how-to work

Signed-off-by: Kuzma Leshakov <k.leshakov@yandex.ru>


trthhrtz:
There is an issue with `create_wallet()`.

In `python3-indy==1.4.0` it has the following structure: 

`async def create_wallet(pool_name: str, name: str,
xtype: Optional[str], config: Optional[str], credentials: Optional[str]) -> None:`

In particular, there is an issue with `credentials`.

`:param credentials: (optional) Wallet credentials json. List of supported keys are defined by wallet type.
     if NULL, then default config will be used.`

If we don't pass `credentials` here (in other words, if we pass `None` as it was before), then the following line: `c_credentials = c_char_p(credentials.encode('utf-8')) if credentials is not None else None` will pass `None` to Rust.

However, Rust version [checks](https://github.com/hyperledger/indy-sdk/blob/1f59c75b5423ed0bf83a30e6cde6a58aed1ffc46/libindy/src/api/wallet.rs#L183) if passed `credentials` is of type string and raises an Error.

Therefore, among other things, I added in this how-to: 
`wallet_credentials = json.dumps({"key": "wallet_key"})` which is passed later on as `credentials`.

PR #866, opened 2018-06-11
inmem_wallet is test only. Mark it as such

AxelNennker:

While testing something else I noticed that the compiler complained about something related to inmem_wallet but InmemWallet is only used to test the wallet service.
InmemWallet should never be compiled in cargo build but only in cargo test.

This PR moves inmem_wallet.rs to the wallet service directory and marks it as used only in tests


vimmerru:
@AxelNennker 

CI fails on compile:

```
  Compiling indy v1.4.0 (file:///var/lib/jenkins/workspace/Indy_SDK_CI_PR-866-OZTCI2MFWBCDIXNPDLFMKSUQY6Q6VWRDWULFDUA3M3QOKYZNLHDA/libindy)

error: couldn't read "tests/utils/../../src/utils/inmem_wallet.rs": No such file or directory (os error 2)

  --> tests/utils/mod.rs:44:9
```

AxelNennker:
I regularly run a RUST_TEST_THREADS=1 cargo test --lib
I don't hit the error: "error: couldn't read "tests/utils/../../src/utils/inmem_wallet.rs": No such file or directory (os error 2)". 

I changed the path now. Although as I don't hit the error the is like somewhat flying blind.

AxelNennker:
[off topic for this PR] there are many more warning while compiling libindy like unused imports and such. That makes seeing the real issues harder. Is somebody fixing those already?

vimmerru:
Yes, we got warning now and code requires corresponded hygiene. There is a PR that will fail CI in case of warnings present and corresponded ticket in Jira.

jovfer:
@AxelNennker `cargo test --lib` is not enough to local verification. This command just run "unit" tests (placed in `src/` folder near non-test codebase). There are "integration" tests in `tests` folder and these tests are also checked during CI/CD.
Personally I use `cargo test --lib` at initial development state, then switch to `cargo test --test <changed module>` and finally use `cargo test` to run whole set of tests.

AxelNennker:
Closing this as it has been dormant for too long and resolving the conflicts is too much work.

PR #867, opened 2018-06-12
[DESIGN] Design of export/import

dkulic:

### DIFF
--- a/doc/design/009-wallet-export-import/README.md
+++ b/doc/design/009-wallet-export-import/README.md
@@ -1,0 +1,123 @@
+# Wallet Export/Import Design
+
+Currently all encryption of the wallet is performed in libindy. Storage implementations may be plugged in.
+This design proposes portable export/import functionality so the "lock-in" of the user to the software
+which created the wallet is avoided.
+
+## Goals and ideas
+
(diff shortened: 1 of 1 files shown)


mikelodder7:
Is there a reason we are using passwords to derive the key instead of creating the key ourselves and just using the public key for encryption and saving the private key somewhere else? Is the reason convenience to the end user?

dkulic:
Looks like I cannot answer directly (the lack of privileges or similar), so I will write here.
1) Moderate for both (memory and cpu) is used, like in open/create of wallet.
2) The reason is to be in-line with open/create.

mikelodder7:
Is there a reason it needs to be same as open/create?
I’m mostly okay with it but open/create leave the wallet directly under your control whereas export could be moving somewhere that isn’t. If my concern is not big enough then just leave it. Moderate setting is okay for open/create

vimmerru:
> Is there a reason it needs to be same as open/create?

It just solves the problem in a similar way and don't have obvious drawbacks. This key is visible for user and introducing of new concept can just cause bad UX

vimmerru:
I am ok with this design

PR #868, opened 2018-06-13
Incremented protocol version

Artemkaaas:
…f genesisis transactions in old format


libindy/src/domain/ledger/request.rs
```diff
@@ -6,6 +6,8 @@ extern crate time;
 use self::indy_crypto::utils::json::JsonEncodable;
 
 
+const PROTOCOL_VERSION: u64 = 2;
```
ashcherbakov: Do we use it in `LedgerStatus` as well?

Artemkaaas: No. I'll add

libindy/src/errors/pool.rs
```diff
@@ -27,6 +28,7 @@ impl fmt::Display for PoolError {
             PoolError::Terminate => write!(f, "Pool work terminated"),
             PoolError::Timeout => write!(f, "Timeout"),
             PoolError::AlreadyExists(ref description) => write!(f, "Pool ledger config already exists {}", description),
+            PoolError::Outdated => write!(f, "Outdated"),
```
vimmerru: I believe we need better description

PR #886, opened 2018-06-19
serde_json::error to Indy Error

AxelNennker:
This PR allows to get rid of tons of map_error calls which make libindy's code hard to read.

By implementing From<serde_json::Error> for IndyError all map_error calls following serde_json can be removed.

There a many places in the code that would benefit from this. This code is easier to read.


vimmerru:
Also we miss context messages. I agree that error processing code looks too boilerplate now, but i am not sure that just changing to default is a right way to proceed. We will discuss this with core team.

AxelNennker:
@Artemkaaas 
I would argue that InvalidState after a json decode is a misleading anyway. If the result of a read from the ledger is not valid json then this should have been signalled through an error not by an invalid result string. The InvalidStructure is still capsuled in a PoolError so the developer still knows where this comes from.

vimmerru:
Not sure that we return InvalidState for invalid ledger data at all. if so then it needs to be changed, but for the following use case InvalidState works well and deeply used:

- We store some json in the wallet
- We read json from the wallet
- We try to deserialize json 
- We raise InvalidState because we detected invalid data in the wallet that should contain only valid data

AxelNennker:
@vimmerru Wouldn't the context be clear from the call stack when the error is raised? Usually there is a line number and file name so the developer knows what which decoding failed.

Is it possible to reduce the number of json strings as parameters anyway? I would like to see it that the API methods immediately convert the json string to proper objects and then pass references to those objects to lower level methods. fail fast

vimmerru:
> @vimmerru Wouldn't the context be clear from the call stack when the error is raised? Usually there is a line number and file name so the developer knows what which decoding failed.

We plan to expose these messages in the future through API. Anyway i am thinking about this proposal.

> Is it possible to reduce the number of json strings as parameters anyway? I would like to see it that the API methods immediately convert the json string to proper objects and then pass references to those objects to lower level methods. fail fast

Application defined jsons are always converted on commands layer and service layer works with domain structures. Anoncreds service can be exception because it uses json interface of indy-crypto.

AxelNennker:
@vimmerru regarding the wallet InvalidState: That would be still capsuled in a WalletError and InvalidStructure is a good name for that error. But if you want to call that

AxelNennker:
Anything I can do to progress this?

PR #887, opened 2018-06-20
Wrappers Export/Import

Artemkaaas:


wrappers/java/src/test/java/org/hyperledger/indy/sdk/wallet/ImportWalletTest.java
```diff
@@ -0,0 +1,44 @@
...
+import org.hyperledger.indy.sdk.did.Did;
+import org.junit.Test;
+
+import java.util.concurrent.ExecutionException;
+
+import static org.hamcrest.CoreMatchers.isA;
+import static org.junit.Assert.assertEquals;
+
+
+public class ImportWalletTest extends IndyIntegrationTestWithSingleWallet {
```
nveskovic: this is actually export-import test, so the other (export) test class is redundant

wrappers/java/src/test/java/org/hyperledger/indy/sdk/wallet/ExportWalletTest.java
```diff
@@ -0,0 +1,35 @@
...
+import org.junit.Test;
+
+import java.io.File;
+import java.util.concurrent.ExecutionException;
+
+import static org.hamcrest.CoreMatchers.isA;
+import static org.junit.Assert.assertTrue;
+
+
+public class ExportWalletTest extends IndyIntegrationTestWithSingleWallet {
```
nveskovic: we should add a cleanup for created file(s)

wrappers/java/src/test/java/org/hyperledger/indy/sdk/wallet/ImportWalletTest.java
```diff
@@ -0,0 +1,44 @@
...
+
+	@Test
+	public void testImportWalletWorks() throws Exception {
+		String did = Did.createAndStoreMyDid(wallet, "{}").get().getDid();
```
nveskovic: maybe to add different types of records, not only dids, so we can make sure export/import works for diff record types

wrappers/java/src/test/java/org/hyperledger/indy/sdk/wallet/ImportWalletTest.java
vimmerru: @nveskovic We use one module per each API endpoint. It can cause some code duplication, but it is a warranty of formal proof for each endpoint.

wrappers/java/src/test/java/org/hyperledger/indy/sdk/wallet/ExportWalletTest.java
vimmerru: It cleanups the whole storage

wrappers/java/src/test/java/org/hyperledger/indy/sdk/wallet/ImportWalletTest.java
vimmerru: Wrappers test tests only wrappers, not libindy. You can check your tests approach in https://github.com/hyperledger/indy-sdk/blob/master/doc/test-design.md

PR #925, opened 2018-07-05
Access wallet from cluster nodes

vimmerru:
* Changes of Wallet API to avoid persist of wallet configurations on libindy side
* Hygienic refactoring
* Review of internal terminology


libindy/src/api/wallet.rs
```diff
@@ -317,10 +286,10 @@ pub extern fn indy_list_wallets(command_handle: i32,
 ///
 /// #Params:
 /// wallet_handle: wallet handle returned by indy_open_wallet
-/// export_config_json: JSON containing settings for input operation.
+/// export_config: JSON containing settings for input operation.
 ///   {
-///     "path": path of the file that contains exported wallet content
-///     "key": passphrase used to derive export key
+///     "path": <strinng>, Path of the file that contains exported wallet content
```
e-nem-y: You have two `n` characters. `<strinng>` instead of `<string>
`

vimmerru: Thanks, will fix in the next PR

PR #940, opened 2018-07-11
Integrate tags based search to Anoncreds workflow

Artemkaaas:


doc/design/002-anoncreds/README.md
```diff
@@ -446,33 +446,61 @@ pub extern fn indy_prover_store_credential(command_handle: i32,
...
+                                         wallet_handle: i32,
+                                         cred_id: *const c_char,
+                                         cb: Option<extern fn(
+                                             xcommand_handle: i32, err: ErrorCode,
+                                             credential_json: *const c_char)>) -> ErrorCode
+```
+
+```Rust
+/// Gets human readable credentials according to the filter.
+/// If filter is NULL, then all credentials are returned.
```
vimmerru: We don't have filter now

doc/design/011-wallet-query-language/README.md
```diff
@@ -0,0 +1,18 @@
+## Wallet Query Language
+
+This language will be used to define queries in Non-secrets, Anoncreds search APIs.
```
vimmerru: I believe it is important to add the note about encrypted and un-encrypted tags

libindy/src/api/anoncreds.rs
```diff
@@ -795,133 +795,137 @@ pub extern fn indy_prover_store_credential(command_handle: i32,
...
-pub extern fn indy_prover_get_credential(command_handle: i32,
-                                         wallet_handle: i32,
-                                         cred_id: *const c_char,
-                                         cb: Option<extern fn(
-                                             xcommand_handle: i32, err: ErrorCode,
-                                             credential_json: *const c_char)>) -> ErrorCode {
-    trace!("indy_prover_get_credential: >>> wallet_handle: {:?}, cred_id: {:?}", wallet_handle, cred_id);
+pub extern fn indy_prover_get_credentials(command_handle: i32,
+                                          wallet_handle: i32,
+                                          filter_json: *const c_char,
```
vimmerru: Do we need to remove filter_json here to keep backward compatibility?

Issue #587, opened 2018-03-21
Docker getting_started unable to access keyserver


dastardlychimp:
When, I was trying to run the docker-compose in indy-sdk/doc/getting-started I wasn't able to get a key from ubuntu's key server.

> Step 5/22 : RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 68DB5E88
>  ---> Running in 5bb694c6f621
> Executing: /tmp/tmp.sAbcDRiKqb/gpg.1.sh --keyserver
> keyserver.ubuntu.com
> --recv-keys
> 68DB5E88
> gpg: requesting key 68DB5E88 from hkp server keyserver.ubuntu.com
> gpg: keyserver timed out
> gpg: keyserver receive failed: keyserver error
> 

I was able to fix it by:
1. replacing `keyserver.ubuntu.com` with `hkp://keyserver.ubuntu.com:80` in **/indy-sdk/doc/getting-started/getting_started.dockerfile** and **indy-sdk/ci/indy-pool.dockerfile**
2. rebuild with `docker-compose build --no-cache`
3. Run with `docker-compose up`

I think the keyserver addresses should be updated, or at the least, the documentation updated to address this.

Thanks.

vimmerru:
@dastardlychimp By default, apt-key adv uses port 11371. Most probably it is prohibited in your firewall. I don't want to change default settings as they works in 99% cases.

May be you can add troubleshooting note to the documentation for your case?

KitHat:
The issue occurred on my machine for a couple of times (for two or three minutes) and then it disappeared. Seems like it was a freeze of a keyserver.

PR #945, opened 2018-07-12
People behind a corporate firewall may have issue building the docker…

fabienpe:
… images because apt-key add may be blocked by the firewall. This fix replaces the call with something equivalent.

Signed-off-by: Fabien <fabienpe@users.noreply.github.com>


vimmerru:
@fabienpe Could you describe a bit how these changes will help with Firewall?

kdenhartog:
An alternative solution I found to this was to publish docker pool image to the docker hub. We could build this into the CI/CD pipeline of node. I also ran into this issue at hyperledger hackfest in February. There's also issue #587 which had this concern as well. It's not a common concern, but it would be good to document alternative solutions to solve this problem if we don't want to change the current method.

fabienpe:
Another workaround I just found is simply to replace
apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 68DB5E88 
by
apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv  68DB5E88

PR #958, opened 2018-07-17
Enabling ReadTheDocs for Hyperledger Indy

mjmckean:
Created a docs/ directory to hold all the files necessary for a consumer-facing documentation hosted at ReadTheDocs (modeled in part after Hyperledger Fabric's ReadTheDocs). Per  https://jira.hyperledger.org/browse/IS-664.


esplinr:
We would appreciate more explanation about the approach you are using:
* What is the role of the make file? Does it need to be executed by the CI pipeline?
* Why is the documentation in RST instead of MarkDown?
* What is the location where ReadTheDocs will look for the documentation: Off of which branch? How will we keep that pointed at the correct branch?

In the future, it will be easier for us to accept the pull request if we have information like this described in the issue. There are a lot of files modified here, and we need an overview to understand what we are looking at.

mjmckean:
> What is the role of the make file? Does it need to be executed by the CI pipeline?
ReadTheDocs builds html versions of documentation using Sphinx, and the make file that was added solely functions in this build. I don't think the CI pipeline would need to execute it, as its success and failure should only influence the ReadTheDocs output. 

> Why is the documentation in RST instead of MarkDown?
ReadTheDocs supports both ReStructuredText and MarkDown; however, I started building with RST because it seems to be better supported on ReadTheDocs. RST also works well with toctrees, which help organize documentation during the build. I spent a decent amount of time looking at how Hyperledger Fabric has structured their ReadTheDocs, and they use RST.

> What is the location where ReadTheDocs will look for the documentation: Off of which branch? How will we keep that pointed at the correct branch?
ReadTheDocs is currently linked to my branch (https://github.com/mjmckean/indy-sdk) through a web hook and updates each time I push to my branch. I have set it up this way as I develop the documentation so that I can push and see the results without interrupting the main branch. At any time we change that--as long as the files are there, the build will take minutes to complete.

Another option to explore would be having an indy-docs repo that hosts this consumer-facing documentation, which could clean up the indy-sdk PR list a bit. @esplinr suggested putting together a HIPE for this idea, which I will work on this week.

burdettadam:
After some reflection on the feedback from this PR, @mjmckean and I have a new HIPE (https://github.com/mjmckean/indy-hipe/tree/master/text/indy-docs-repo) proposing new documentation convention.

michaeldboyd:
This PR can be closed. In a day or two I will open a new PR based on the specifications in the new documentation framework HIPE: https://github.com/hyperledger/indy-hipe/pull/63

PR #959, opened 2018-07-18
Android Unit tests

faisal00813:
This PR has support for running unit tests on the headless Android emulator
Also has optimized Android build scripts


Jenkinsfile.cd
```diff
@@ -383,6 +384,48 @@ def ubuntuTesting() {
...
+                checkout scm
+
+                libindyVersion = getSrcVersion("libindy")
+                def libnullpayVersion = getSrcVersion("libnullpay")
+                echo 'Building android binaries: Building docker image'
+                buildEnv = dockerHelpers.build('indy-sdk', 'libindy/ci/ubuntu.dockerfile libindy/ci')
+                sh 'chmod -R 777 libindy/'
+                sh 'chmod -R 777 libindy/ci/'
+
+                //TODO: Enable emulator test when the unit test start passing on android device
```
jovfer: What status of this TODO? Do we plan to merge it?

libindy/src/utils/logger.rs
```diff
@@ -5,7 +5,7 @@ extern crate log;
 extern crate android_logger;
 
 use self::env_logger::Builder;
-use self::log::LevelFilter;
+use self::log::{Record, LevelFilter, Level};
```
jovfer: Should we switch default log level for android from `Trace` to `Info` or `Off` below?
https://github.com/hyperledger/indy-sdk/pull/959/files#diff-5d11eb6aa67bd3369d836a1a7f0ca492R23

Jenkinsfile.cd
faisal00813: Several unit tests of libindy are failing on Android. I suspect these are failing mostly because of the unavailability of the internet permissions to binary. My effort was mainly focused on executing the tests on Android regardless of the outcome of the tests.
Because of the failing unit tests, I think these tests should not run on CD pipeline without fixing them. 
Also, I noticed it is taking more than an hour to execute tests on the emulator. I have tried with increased Ram on the emulator but it does not have much effect on the time to execute tests.

jovfer: We see significant degradation on iOS emulator with a large amount of logging. And there is a trace level in current android config. May be it's a similar issue?

jovfer: Can we run some set of test (only `--lib` for example?) to be sure about this part of CI is working well

jovfer: BTW if you suspect, that the reason of failed tests is invalid permission in CI environment, the fix of this is must have part of creating the CI pipeline

jovfer: If the reason is something else than it's critical blocker for publishing any artifacts for android.

jovfer: @vimmerru FYI

vimmerru: @faisal00813 We need to understand why tests fail. Do you have logs?

jovfer: After additional discussion I can suggest the next steps:
1) specify appropriate ticket in this comment, not just `//TODO ...` but `//TODO IS-XXX: ...`
2) consider this PR as minimal solution for Android CD - add appropriate comment in Jira (with the link to additional ticket)
3) Mark status of android builds as experimental until we have clear vision about the root of the problem

jovfer:
@faisal00813 please update main readme section about installation for android. https://github.com/hyperledger/indy-sdk/blob/master/README.md#installing-the-sdk
Please specify where android artifacts can be found and also please note about experimental status of these artifacts (as we have skipped tests while CI/CD)

ci/android-zip-and-upload.sh
```diff
@@ -11,17 +11,32 @@ key="$3"
...
+[ -z $artifact ] && exit 6
+[ -z $is_stable ] && exit 7
 
-ssh -v -oStrictHostKeyChecking=no -i $key repo@192.168.11.115 mkdir -p /var/repository/repos/android/${artifact}/${branchName}/${version}-${buildNumber}
+if [ "${is_stable}" == "1" ]; then
+    ssh -v -oStrictHostKeyChecking=no -i $key repo@192.168.11.115 mkdir -p /var/repository/repos/android/${artifact}/stable/${branchName}/${version}
 
 cat <<EOF | sftp -v -oStrictHostKeyChecking=no -i $key repo@192.168.11.115
-cd /var/repository/repos/android/${artifact}/$branchName/$version-$buildNumber
+cd /var/repository/repos/android/${artifact}/stable/$branchName/$version
```
jovfer: `branchName` is not required here

ci/android-zip-and-upload.sh
```diff
@@ -0,0 +1,42 @@
...
+[ -z $branchName ] && exit 4
+[ -z $buildNumber ] && exit 5
+[ -z $artifact ] && exit 6
+[ -z $is_stable ] && exit 7
+
+if [ "${is_stable}" == "1" ]; then
+    ssh -v -oStrictHostKeyChecking=no -i $key repo@192.168.11.115 mkdir -p /var/repository/repos/android/${artifact}/stable/${branchName}/${version}
+
+cat <<EOF | sftp -v -oStrictHostKeyChecking=no -i $key repo@192.168.11.115
+cd /var/repository/repos/android/${artifact}/stable/$branchName/$version
```
jovfer: `$branchName` isn't required here

ci/android-zip-and-upload.sh
```diff
@@ -0,0 +1,42 @@
...
+
+[ -z $arch ] && exit 1
+[ -z $version ] && exit 2
+[ -z $key ] && exit 3
+[ -z $branchName ] && exit 4
+[ -z $buildNumber ] && exit 5
+[ -z $artifact ] && exit 6
+[ -z $is_stable ] && exit 7
+
+if [ "${is_stable}" == "1" ]; then
```
jovfer: please simplify this `if`
e.g.
```
if [ "${is_stable}" == "1" ]; then
 target_dir=...path1
else
 target_dir=..path2
fi
...
```

Jenkinsfile.cd
```diff
@@ -636,6 +761,9 @@ def publishingRCtoStable() {
                 echo 'Moving Ubuntu RC artifacts to Stable: libnullpay'
                 publishLibnullpayDebRCtoStable(testEnv)
 
+                echo 'Moving Android RC artifacts to Stable: libindy'
+                androidPublishing("1")
```
jovfer: this call will results in full rebuild android artifacts from scratch. If I understand build script correctly, we can just use RC archive, rename it and publish to stable directory

faisal00813: Fixed!

PR #968, opened 2018-07-19
Export file format hardening

vimmerru:
Changed export file format:
- Use MsgPack instead of custom entities serialization to be more reliable and allow extend-ability in backward compatible way
- Changed header format  to be more reliable and allow extend-ability in backward compatible way
- Use STOP message to make sure that there was no truncation of export file

Removed EXPERIMENTAL notice for import/export API endpoints.


doc/design/009-wallet-export-import/README.md
```diff
@@ -90,34 +90,64 @@ For deriving keys from passphrase **Argon2** memory-hard function is used with r
...
+
+* `header_length`: length of the serialized header as 4b unsigned little endian integer
+* `header`: MessagePack serialized header entity
+* `header_hash`: 32B **SHA-256** hash of the header.
+
+--- encrypted stream ---
+
+* `record1_length`: length the serialized record as 4b unsigned little endian integer
+* `record1`: MessagePack serialized record entity
+* `record1_hash`: 32B **SHA-256** hash of the serialized record.
```
jovfer: this hash seems redundant as we use ChaCha20Poly1305IETF

PR #972, opened 2018-07-20
Make network timeouts are configurable.

jovfer:


libindy/src/services/pool/types.rs
```diff
@@ -465,6 +465,27 @@ impl PoolConfig {
     }
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct PoolOpenConfig {
```
vimmerru: It is better to have in domain as changes can count backward compatibility problems

libindy/src/services/pool/types.rs
```diff
@@ -465,6 +465,30 @@ impl PoolConfig {
...
```
jovfer: @vimmerru @dsurnin @ashcherbakov do you have more suggestions about items to configure?

libindy/src/services/pool/types.rs
```diff
@@ -465,6 +465,27 @@ impl PoolConfig {
...
+impl Default for PoolOpenConfig {
+    fn default() -> Self {
+        PoolOpenConfig {
+            timeout: super::networker::POOL_ACK_TIMEOUT,
+            extended_timeout: super::networker::POOL_REPLY_TIMEOUT,
+        }
+    }
+}
+
+impl JsonEncodable for PoolOpenConfig {}
```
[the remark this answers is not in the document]
jovfer: I suggest to avoid mixed state in master (partial removed `JsonEncodable`). If we decide to do something else with JsonEncodable against "just remove" it will be easy to replace all JsonEncodable at once.

PR #978, opened 2018-07-23
Improve flexibility of network timeouts in libindy

jovfer:


vimmerru:
@jovfer We need documentation for timeouts and pre_ordered_nodes params in libindy and all wrappers

vimmerru:
Agreed that will be provided by next PR to don't block teams

PR #1010, opened 2018-07-31
Static json strings

AxelNennker:
Creating static json string by first creating a JSONObject and then stringifying it does not compile on Android because JSONObject might throw an exception.

I changed that to static strings. Please review.


jovfer:
@AxelNennker thank you for PR and helping with Android! Are other parts of java wrapper compatible with Android?

AxelNennker:
There seem to be other things that changed. JSONObject uses the method similar that does not exist on Android... Not sure what to do. 
1) Find a json library that has 'similar' or
2) bring your own 'similar' and change the wrapper
3) remove 'similar' if possible

https://github.com/hyperledger/indy-sdk/blob/4459d2f9b929caa23aadf0c71651f352d84dbde1/wrappers/java/src/test/java/org/hyperledger/indy/sdk/non_secrets/GetRecordTest.java#L30

AxelNennker:
I think that 'similar' can and should be removed from the code. 
E.g.: At the code linked to above I suggest to change the code to something like
assertEquals(id, actual.get("id"));
assertEquals(value, actual.get("value"));

Should I create a JIRA issue for this task?

jovfer:
After brief review I prefer 2nd option (create custom similar to be available on Android). But it's not a final vision...
> Should I create a JIRA issue for this task?
appreciated. Please describe all 3 options and motivation to do it (support Android (at least?)). Jira comments is better place to discuss rather this PR, i think

AxelNennker:
JIRA issue is here: https://jira.hyperledger.org/browse/IS-853

wrappers/java/src/test/java/org/hyperledger/indy/sdk/IndyIntegrationTest.java
```diff
@@ -56,28 +56,20 @@
...
-	protected static final String WALLET_CREDENTIALS =
-			new JSONObject()
-					.put("key", "key")
-					.toString();
-	protected static final String PLUGGED_WALLET_CONFIG =
-			new JSONObject()
-					.put("id", WALLET)
-					.put("storage_type", "unknown_type")
-					.toString();
+	protected static final String WALLET_CONFIG = "{ \"id\":" + WALLET + ", \"storage_type\":" + TYPE + '}';
```
Artemkaaas: Please check it. Seems we need to manually add quotes.
For me, WALLET_CONFIG will be `{ "id":Wallet1, "storage_type":default}`.

AxelNennker:
added the quotes.

PR #1028, opened 2018-08-02
marking old instructions as deprecated

suedonym:
Signed-off-by: Susan Bradford <susan.bradford@evernym.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,6 @@ It's good for development, but this build can be in 100+ times slower for some m
 If you would like to analyse CPU performance of libindy for your use case, you have to use release artifacts (`cargo build --release`). 
 
 ## How to start local nodes pool with docker
-THESE INSTRUCTIONS ARE NO LONGER VALID
 To test the SDK codebase with a virtual Indy node network, you can start a pool of local nodes using docker:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -164,6 +164,7 @@ It's good for development, but this build can be in 100+ times slower for some m
 If you would like to analyse CPU performance of libindy for your use case, you have to use release artifacts (`cargo build --release`). 
 
 ## How to start local nodes pool with docker
+THESE INSTRUCTIONS ARE NO LONGER VALID
```
jovfer: Right now it's main and recommended way to obtain local development pool. Why do you like to mark it as invalid?

jovfer: It's better to put the link to this section ( `
https://github.com/hyperledger/indy-sdk/blob/master/README.md#how-to-start-local-nodes-pool-with-docker ` ) to other places (e.g. RHEL readme https://github.com/hyperledger/indy-sdk/blob/master/doc/rhel-build.md#starting-up )

esplinr:
I think we made this change to the wrong repository. It should have been the Docker instructions in the README.md for Indy Node. We need to update these instructions per your feedback, and mark the Indy Node instructions as deprecated until we have the chance to update them.

PR #1034, opened 2018-08-03
remove redundant variant of checkResult

AxelNennker:
This variant of checkResult is not needed because at each place where checkResult is called there is a future which needs to complete with an exception if errorCode != success


AxelNennker:
I guess I have to do that again. What is here is not what I remember but matches my local branch. 
Wonders of git, I guess. Stay tuned.

PR #1048, opened 2018-08-07
encode and decode for Attributes

AxelNennker:

Added methods to encode and decode Attributes for https://jira.hyperledger.org/browse/IS-786

wrote that the encoding is application specific and the ZKPs don't care as long as the result of the encoding is a string representing a number. @swcurran 's counter argument was that it helps interoperabilty if common values are encoded in a common way.

I modelled this along the lines of https://von-agent.readthedocs.io/en/latest/_modules/von_agent/codec.html?highlight=codec but not 100%. Not sure about their float encoding as I think that if
e1 = encode(float1) and e2 = encode(float2) and float1 > float2 then
Number(e1) > Number(e2) etc

Of course all wrappers should use the same encoding or use a Rust implementation which is not 100% correct to put into libindy...

Should this be introduced then all examples should be changed...


AxelNennker:
Working with @sklump  one this to have the von_anchor/codec.py and this in sync.

jovfer:
related HIPE discussion https://github.com/hyperledger/indy-hipe/pull/47

jovfer:
marked as paused as HIPE discussion is in a progress

esplinr:
Much of this work is expected to be superseded later in 2019 by the work @brentzundel and @ken-ebert have been doing to standardize proof presentations in the W3C Verifiable Credential specification.

We are nervous about accepting this contribution because the new approach is likely to be backwards incompatible with this one.

Do you have an important need to deploy this capability now? Can you maintain it for your solutions outside of the official SDK?

AxelNennker:
Closing this because it is not really needed to standardize encoding. Would be just nice to have consistent examples.

PR #1050, opened 2018-08-08
Fixed: Fee can be set below zero

Artemkaaas:


cli/src/commands/ledger.rs
```diff
@@ -1301,7 +1301,7 @@ fn print_response_receipts(receipts: Option<Vec<serde_json::Value>>) -> Result<(
 }
 
 fn parse_payment_fees(fees: &Vec<&str>) -> Result<String, ()> {
-    let mut fees_map: HashMap<String, i32> = HashMap::new();
+    let mut fees_map: HashMap<String, u64> = HashMap::new();
```
jovfer: Is it possible to use SerdeValue::Number here and call smth like `is_positive` for it?

PR #1055, opened 2018-08-08
start emulator for target arch only

AxelNennker:

### DIFF
--- a/libindy/ci/setup.android.env.sh
+++ b/libindy/ci/setup.android.env.sh
@@ -62,8 +62,8 @@ create_avd(){
                 -f \
                 -c 1000M
 
-        ANDROID_SDK_ROOT=${ANDROID_SDK} ANDROID_HOME=${ANDROID_SDK} ${ANDROID_HOME}/tools/emulator -avd ${TARGET_ARCH}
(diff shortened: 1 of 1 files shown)


faisal00813:
@AxelNennker Were you able to run the x86 emulator on EC2 instance? I was facing issues related to QEMU when I tried to run x86 emulator. 
I believe it is fine for now since we are not running the emulator based tests in Jenkins CI.

AxelNennker:
I think running x86 emulator on EC2 is not the topic of this PR. The current script starts two emulator. One for the target arch which is arm anyway and then the second for arm without window and audio.
This PR starts exactly one emulator without window and audio and for the target arch which is arm.

I just saw that you removed the double emulator creation some days ago. That is good but now you are ignoring the TARGET_ARCH parameter. You should use this when you create the avd like you do when you delete the avd. 
For EC2 you can set the TARGET_ARCH to arm and everything works.

AxelNennker:
Hm, maybe you did not mean your question to be related to this PR... Then: No, I do not use EC2
I heard that the x86 emulator is 10x faster than the arm emulator and that CI running time is affecting the number of PRs Indy can handle per day. So speeding up CI by using the x86 emulator might improve that situation.

PR #1059, opened 2018-08-09
Added flag to use simplified security method for key derivation

Artemkaaas:

### DIFF
--- a/libindy/src/domain/wallet/export_import.rs
+++ b/libindy/src/domain/wallet/export_import.rs
@@ -2,18 +2,41 @@ use std::collections::HashMap;
 
 #[derive(Debug, Serialize, Deserialize)]
 pub enum EncryptionMethod {
-    ChaCha20Poly1305IETF { // **ChaCha20-Poly1305-IETF** cypher in blocks per chunk_size bytes
-    salt: Vec<u8>,  // pwhash_argon2i13::Salt as bytes. Random salt used for deriving of key from passphrase
-    nonce: Vec<u8>, // chacha20poly1305_ietf::Nonce as bytes. Random start nonce. We increment nonce for each chunk to be sure in export file consistency
-    chunk_size: usize, // size of encrypted chunk
+    ChaCha20Poly1305IETF {
+        // **ChaCha20-Poly1305-IETF** cypher in blocks per chunk_size bytes
+        salt: Vec<u8>,
+        // pwhash_argon2i13::Salt as bytes. Random salt used for deriving of key from passphrase
+        nonce: Vec<u8>,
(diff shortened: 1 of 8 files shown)


libindy/src/domain/wallet/mod.rs
```diff
@@ -14,16 +14,25 @@ pub struct Credentials {
...
 
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Metadata {
     pub keys: Vec<u8>,
     pub master_key_salt: Vec<u8>,
-}
\ No newline at end of file
+}
+
+fn default_false() -> bool { false }
```
jovfer: false is default for boolean

libindy/src/services/wallet/export_import.rs
```diff
@@ -71,7 +80,7 @@ pub(super) fn export(wallet: &Wallet, writer: &mut Write, passphrase: &str, vers
     Ok(())
 }
 
-pub(super) fn import(wallet: &Wallet, reader: &mut Read, passphrase: &str) -> Result<(), WalletError> {
+pub(super) fn import(wallet: &Wallet, reader: &mut Read, passhrase: &str) -> Result<(), WalletError> {
```
jovfer: typo

libindy/src/domain/wallet/export_import.rs
```diff
@@ -33,8 +56,12 @@ pub struct Header {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Record {
     #[serde(rename = "type")]
-    pub type_: String, // Wallet record type
-    pub id: String, // Wallet record id
-    pub value: String, // Wallet record value
-    pub tags: HashMap<String, String>, // Wallet record tags
+    pub type_: String,
+    // Wallet record type
```
jovfer: fix autoformat

jovfer:
@mikelodder7 let me rephrase you suggestion to be on the same page (and also share some background):
enum `EncryptionMethod` is used to provide backward-compatibility
Initially it was only one option here: it was called as `ChaCha20Poly1305IETF` and assume key derivation in `moderate` mode for `argon2i13` algorithm.

We can add almost any new options. And Artem just add one more specific `ChaCha20Poly1305IETFWithSimplify` which assumes `interactive` mode for `argon2i13` algorithm

1) Do you suggest to not add one more variant with fixed parameters, but add configurable variant like
`ChaCha20Poly1305IETFWithConfigForKeyDerivation` ?
2) if 1 is true, right now it seems like metadata to store is salt + opslimit + memlimit + alg. How do you think, should we fix this list of metadata fields or use some flexible way to allow extend it (it may decrease code readability)?

PR #1064, opened 2018-08-13
use constants from libsodium for sizes

lovesh:
Signed-off-by: lovesh harchandani <lovesh.bond@gmail.com>

### DIFF
--- a/libindy/src/utils/crypto/ed25519_sign/sodium.rs
+++ b/libindy/src/utils/crypto/ed25519_sign/sodium.rs
@@ -11,13 +11,14 @@ use utils::crypto::ed25519_box;
 use utils::crypto::randombytes::randombytes;
 
 pub const SEEDBYTES: usize = sign::SEEDBYTES;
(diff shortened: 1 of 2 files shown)


libindy/src/utils/crypto/ed25519_sign/sodium.rs
```diff
@@ -5,38 +5,40 @@ use errors::crypto::CryptoError;
...
 
 extern {
     // TODO: fix hack:
     // this functions isn't included to sodiumoxide rust wrappers,
     // temporary local binding is used to call libsodium-sys function
     pub fn crypto_sign_ed25519_pk_to_curve25519(
-        curve25519_pk: *mut [u8; 32],
-        ed25519_pk: *const [u8; 32]) -> c_int;
+        curve25519_pk: *mut [u8; PUBLICKEYBYTES],
+        ed25519_pk: *const [u8; PUBLICKEYBYTES]) -> c_int;
```
jovfer: May be it's better to use separate constant here? like for private keys.

PR #1079, opened 2018-08-19
Type safety

AxelNennker:
This PR converts the json string from libindy's api to a 'real' type as soon as possible.

This makes it easier for developers who are looking e.g. at the wrappers and want to know what the actual content of a json string needs to be to find the type in the api. Otherwise developers need to search the rust files to find the type - usually at the point were the rust code de-serializes the json string.
Also this increases the type safety of the rust code by a small amount because one json string is not distinguishable from another json string.


libindy/src/api/did.rs
```diff
@@ -47,29 +51,39 @@ use std::ptr;
...
+                                            did_info_json: *const c_char,
                                             cb: Option<extern fn(xcommand_handle: i32,
                                                                  err: ErrorCode,
                                                                  did: *const c_char,
                                                                  verkey: *const c_char)>) -> ErrorCode {
-    trace!("indy_create_and_store_my_did: >>> wallet_handle: {:?}, did_json: {:?}", wallet_handle, did_json);
+    trace!("indy_create_and_store_my_did: >>> wallet_handle: {:?}, did_json: {:?}", wallet_handle, did_info_json);
 
-    check_useful_c_str!(did_json, ErrorCode::CommonInvalidParam3);
+    check_useful_c_str!(did_info_json, ErrorCode::CommonInvalidParam3);
```
vimmerru: I suggest to define new macro check_useful_json! that will deserialize json to the variable with the same name. It will allow to remove a lot of boilerplate. Usage will be the following:

```
check_useful_json!(did_info, ErrorCode::CommonInvalidParam3, MyDidInfo);
``

**Comment by vimmerru** [comment]
We have tradeoff between early validation and encapsulation. As we moved all (or at least the most) of domain types to dedicated domain module it can be acceptable. I suggest to look to my comment about check_usefull_json! macro

File: libindy/src/api/did.rs (hunk 1 above)
Comment (AxelNennker) [reply libindy/src/api/did.rs]: I wanted to change the code as little as possible. Yes, I'll give it a go.

**Comment by AxelNennker** [comment]
Yes, I would like to move all the types up. Maybe they are even be exposed in the API so that wrappers can directly access the rust structs.

Also we could change the API so that there are methods with the 'real' types too next to the current json strings. We could write those in Rust and have an 'annotation' `#[derive(StructToJsonString)]` that generates the current API from the struct parameter versions. 

Taking it step by step to more idiomatic code.

**Comment by AxelNennker** [comment]
Now if only the other types were visible at API-level e.g. PoolInfo

File: libindy/src/api/did.rs [hunk 2]
```diff
@@ -47,29 +51,29 @@ use std::ptr;
...
+                                            did_info_json: *const c_char,
                                             cb: Option<extern fn(xcommand_handle: i32,
                                                                  err: ErrorCode,
                                                                  did: *const c_char,
                                                                  verkey: *const c_char)>) -> ErrorCode {
-    trace!("indy_create_and_store_my_did: >>> wallet_handle: {:?}, did_json: {:?}", wallet_handle, did_json);
+    trace!("indy_create_and_store_my_did: >>> wallet_handle: {:?}, did_json: {:?}", wallet_handle, did_info_json);
 
-    check_useful_c_str!(did_json, ErrorCode::CommonInvalidParam3);
+    check_useful_json!(did_info_json, ErrorCode::CommonInvalidParam3, my_did_info, MyDidInfo);
```
vimmerru: Rust allows re-definition of local variable. So we can use only ```did_info``` name and reassign it to struct after macros call.

jovfer: I'm agree with @AxelNennker that we should drop `_json` suffix here.

If we will use re-definition we should rename origin variables before. Right now only in API  but may be we will use this macro for other places?
@vimmerru do you suggest to rename API parameters to use re-definition?
If we wouldn't like to change API parameters names current solution seems like most correct.

vimmerru: I suggest to remove "json" suffix as it looks redundant for me. We have part of cases with suffix and part without suffix and i more like to drop it in all cases as it doesn't provide real benefits. 

In current code we are following approach to avoid creation of new variables in entities conversion chains and it looks more consistent to me. If in the future we will get the case when 2 variables will be needed we can just add optional parameter to the macro.

libindy/src/api/anoncreds.rs
```diff
@@ -55,6 +57,21 @@ pub extern fn indy_issuer_create_schema(command_handle: i32,
 
     trace!("indy_issuer_create_schema: entity >>> issuer_did: {:?}, name: {:?}, version: {:?}, attrs: {:?}", issuer_did, name, version, attrs);
 
+    let attrs: HashSet<String> = match serde_json::from_str(&attrs)
```
vimmerru: @AxelNennker I suggest to stop this PR on conversion of jsons to domain types. This early validation is a big question and will make this PR hard to merge.

AxelNennker: ok

AxelNennker: Any estimate one when there will be a decision on the big question?

As I said in the description the PR I think that this greatly improves understanding for developers using the API.

vimmerru: In my vision we can call STATIC-only validation in API layer, but this validation must be defined in domain layer. I suggest the following:

- We can can define trait Validatable with one method validate() -> Result<(), Error> in utils/validation
- For each domain struct that requires static validation we can implement this trait
- We can define additional macro check_usefull_valid_json! that will also call validate method on domain struct

@jovfer @AxelNennker What do you think?

One addition that i suggest is to return position error instead of CommonInvalidStructure here, but we need consensus here as it is breaking change.

vimmerru: Also we need consider to remove static validation from command and service layers in this case

AxelNennker: I would do this step by step.

Before this PR the API checked for valid cstr which this PR keeps and should keep.
What this PR adds is adding `check_useful_json!` which uses `serde_json::from_str::<$t>(&$x)` to parse the json string $x to the domain type $t.

I think this validation is enough for this step.

Not sure about a new trait. I haven'd tried it but serde_json might be all we need already and then implement from_str for the domain type... which can do additional validation steps on the instance variable.

// What is position error? The api only returns error code.

vimmerru:
@AxelNennker FYI i created https://jira.hyperledger.org/browse/IS-912 for early static validation

AxelNennker:
I don't see the difference between this PR and what the current code does with regard to the error code returned. The difference is when the error is detected not which error code is returned.
Isn't it invalid structure in both versions?
After this PR is merged the invalid json is immediately detected and the error invalid structure is returned.
In the current code the invalid json is detected later and the same error code is returned.
Or do I misunderstand the code?

// still don't know what position error is that you also mention in IS-912

PR #1086, opened 2018-08-22
Support of raw wallet keys

Artemkaaas:


libindy/src/services/wallet/export_import.rs
```diff
@@ -33,7 +33,8 @@ pub(super) fn export(wallet: &Wallet, writer: &mut Write, passphrase: &str, vers
             salt: salt[..].to_vec(),
             nonce: nonce[..].to_vec(),
             chunk_size,
-        }
+        },
+        KeyDerivationMethod::RAW => return Err(WalletError::CommonError(CommonError::InvalidStructure("RAW key derivation method is not acceptable".to_string())))
```
mikelodder7: Help me understand why raw keys for export and import are not allowed. Why would we only support key derivation?

vimmerru: I believe there is no reason except that we don't have use cases for this. 

@Artemkaaas @mikelodder7 I though a bit more. I believe it is better to keep all possible options for export too to keep API consistent.

doc/design/001-cli/README.md
```diff
@@ -127,14 +127,14 @@ indy> wallet <command>
 #### Wallet create
 Create new wallet with specified name:
 ```
-indy> wallet create <wallet name> key [storage_type=<storage_type>] [storage_config={config json}]
+indy> wallet create <wallet name> key [key_type=<key_type>] [storage_type=<storage_type>] [storage_config={config json}]
```
vimmerru: I suggest to rename key_type to key_derivation_method

cli/src/commands/wallet.rs
```diff
@@ -26,6 +26,10 @@ pub mod create_command {
     command!(CommandMetadata::build("create", "Create new wallet with specified name")
                 .add_main_param("name", "Identifier of the wallet")
                 .add_required_deferred_param("key", "Auth key for the wallet")
+                .add_optional_param("key_type", "Type of auth key. One of:
```
vimmerru: I suggest to rename key_type  to key_derivation_method

libindy/src/api/wallet.rs
```diff
@@ -519,6 +524,50 @@ pub extern fn indy_delete_wallet(command_handle: i32,
     res
 }
 
+/// Generate wallet master key.
```
vimmerru: Update description:

Generate wallet master key. Returned key is compatible with "RAW" key derivation method. It allows to avoid expensive key derivation for use cases when wallet keys can be stored in a secure enclave.

libindy/src/api/wallet.rs
```diff
@@ -473,13 +477,14 @@ pub extern fn indy_close_wallet(command_handle: i32,
...
 ///   "storage_credentials": optional<object> Credentials for wallet storage. Storage type defines set of supported keys.
 ///                          Can be optional if storage supports default configuration.
 ///                          For 'default' storage type should be empty.
-///   "key_derivation_method": optional<string> algorithm to use for master key derivation:
-///                            ARAGON2I_MOD (used by default)
-///                            ARAGON2I_INT - less secured but faster
+///   "key_derivation_method": optional<string> Type of wallet auth key:
+///                             ARAGON2I_MOD - derive secured wallet master key (used by default)
+///                             ARAGON2I_INT - derive secured wallet master key (less secured but faster)
+///                             RAW - raw wallet key master provided (skip derivation)
```
vimmerru: Update description:

raw wallet key master provided (skip derivation). RAW keys can be generated with indy_generate_wallet_key call

libindy/src/commands/wallet.rs
```diff
@@ -220,4 +229,22 @@ impl WalletCommandExecutor {
         trace!("_import <<< res: {:?}", res);
         Ok(res)
     }
+
+    fn _generate_key(&self,
+                     config: Option<&str>) -> Result<String> {
+        trace!("_generate_key >>>config: {:?}", secret!(config));
+
+        let config = serde_json::from_str::<serde_json::Value>(config.unwrap_or("{}"))
```
vimmerru: Use domain object instead

libindy/src/services/wallet/mod.rs
```diff
@@ -534,6 +500,43 @@ impl WalletService {
...
+            KeyDerivationMethod::RAW =>
+                (None, encryption::raw_master_key(key)?),
+            KeyDerivationMethod::ARAGON2I_INT | KeyDerivationMethod::ARAGON2I_MOD => {
+                let master_key_salt = encryption::gen_master_key_salt()?;
+                let master_key = encryption::derive_master_key(key, &master_key_salt, key_derivation_method)?;
+                (Some(master_key_salt[..].to_vec()), master_key)
+            }
+        };
+
+        let metadata = Metadata {
```
vimmerru: Can we make metadata enum based in key deriviation method in a backward compatible way?

libindy/src/domain/wallet/mod.rs
```diff
@@ -42,5 +43,5 @@ pub struct ExportConfig {
 #[derive(Debug, Serialize, Deserialize)]
 pub struct Metadata {
```
vimmerru: Seems we need to make metadata enum as content is different for different methods

PR #1096, opened 2018-08-28
Reduced unnecessary memory copy during json deserialization.

Artemkaaas:
Little improvement of the PR: https://github.com/hyperledger/indy-sdk/pull/1079.
Reduced unnecessary memory copy during json deserialization.


libindy/src/api/non_secrets.rs
```diff
@@ -41,7 +45,7 @@ pub extern fn indy_add_wallet_record(command_handle: i32,
     check_useful_c_str!(type_, ErrorCode::CommonInvalidParam3);
     check_useful_c_str!(id, ErrorCode::CommonInvalidParam4);
     check_useful_c_str!(value, ErrorCode::CommonInvalidParam5);
-    check_useful_opt_c_str!(tags_json, ErrorCode::CommonInvalidParam6);
+    check_useful_opt_json!(tags_json, ErrorCode::CommonInvalidParam6, HashMap<String, String>);
```
vimmerru: It is better to use domain type for tags instead of HashMap. In the future we can introduce static validation and validation can be different for different for different domain types.

PR #1121, opened 2018-09-05
Ledger check json

AxelNennker:
This PR has two parts.

1) check api parameters that are request json
1.1) more checks might still be added to check_request macro
1.2) This PR does not use check_useful_json with type Request<Value> because further down the line Value is use.
1.3) not all request parameters are checked because further down the line the parameter string is just submitted to the ledger without any Request-specific handling.

2) Request.rs was not Rust idiomatic
2.1) The Request new method was too similar to the normal constructor
2.2) Request::build_request did just to_string. The general goal is to have more specific types in the Indy code base. Not everything should be String but more specific so that developers know what legal content of parameters are. 
2.3) removed ledger_services.build_nym_request call in one place because I think ledger_services' methods that do not call out to the ledger should be called directly

### DIFF
--- a/libindy/src/utils/cstring.rs
+++ b/libindy/src/utils/cstring.rs
@@ -58,6 +58,27 @@ macro_rules! check_useful_opt_json {
     }
 }
 
+macro_rules! check_request {
+    ($x:ident) => {
+        let $x: Value = match serde_json::from_str(&$x) {
+            Ok(Some(val)) => val,
+            _ => {
+                trace!("indy_sign_and_submit_request: could not parse request as valid json: {:?}", $x);
+                return ErrorCode::CommonInvalidParam4;
+            }
+        };
+        if !$x.is_object() {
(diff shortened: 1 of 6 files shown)


libindy/src/utils/cstring.rs
```diff
@@ -58,6 +58,27 @@ macro_rules! check_useful_opt_json {
     }
 }
 
+macro_rules! check_request {
+    ($x:ident) => {
+        let $x: Value = match serde_json::from_str(&$x) {
+            Ok(Some(val)) => val,
+            _ => {
+                trace!("indy_sign_and_submit_request: could not parse request as valid json: {:?}", $x);
+                return ErrorCode::CommonInvalidParam4;
```
vimmerru: We shouldn't hardcore error code here CommonInvalidParam4

AxelNennker:
Resolving this gets me into merge hell. Sorry - I guess this happens when a PR get too old.
I am closing this.

PR #1122, opened 2018-09-05
validate_did is static

AxelNennker:

By turning validate_did into a static method the crypto_services variable is not needed any more and are deleted by this PR where possible.

### DIFF
--- a/libindy/src/commands/anoncreds/issuer.rs
+++ b/libindy/src/commands/anoncreds/issuer.rs
@@ -112,21 +112,18 @@ pub struct IssuerCommandExecutor {
     pub blob_storage_service: Rc<BlobStorageService>,
     pub pool_service: Rc<PoolService>,
     pub wallet_service: Rc<WalletService>,
-    pub crypto_service: Rc<CryptoService>
 }
 
 impl IssuerCommandExecutor {
     pub fn new(anoncreds_service: Rc<AnoncredsService>,
                pool_service: Rc<PoolService>,
                blob_storage_service: Rc<BlobStorageService>,
-               wallet_service: Rc<WalletService>,
-               crypto_service: Rc<CryptoService>) -> IssuerCommandExecutor {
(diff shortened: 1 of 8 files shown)


jovfer:
@AxelNennker now DID validation is very simple, but in general it may depends on crypto/did type (pure Indy, Sovrin, ...) and contains some logic like in one method above `validate_verkey` https://github.com/hyperledger/indy-sdk/pull/1122/files#diff-b2aa1e9288e2d0f2def56a2310115be5L369

So my personal feeling here - it's better to keep generic approach for the full-qualified DID with potential complex checks.

@vimmerru any thoughts?

AxelNennker:
One aspect is how Indy integrates into Hyperledger in general and whether the services, CryptoServices, and others  are the way to implement the abstraction to have different and several different services in Indy.

Currently the internal did format is not important for the outside world and validation is simple. If we export dids then we need to adhere to the did spec. I don't currently see that Indy needs to deal with dids that are not created by Indy. Probably also a question for @nage 
I prefer simpler code over overly flexible code when the flexibility is not happening soon or never.

jovfer:
One more concerns here: 
by design service in libindy is isolated item and there is no direct interaction between service, only throw command layer. If we will make some functions in some service as public static - we will lose compile time check about service dependency.
May be we should consider option to move DID validation from service to utils

AxelNennker:
Yes, maybe it is a good idea to move DID validation away from the crypto_service

vimmerru:
I am concerned about having static functions in a service layer as they contradict with the most of patterns for dependency injection and mocking. We don't use it too much for now, but can switch later. 

Also we will start usage of different types of DIDs in the near future. As a result we will need access to service state to get correct handler and will need to return back to instance method approach.

I suggest the following:

1. We can keep the code as is. 
2. We can move validate method to /utils/did.rs and use it.

Personally i am in favor of approach (1) as the current PR doesn't solve real problems, but can introduce complexity in the future refactorings.

PR #1131, opened 2018-09-10
indy-node and indy-plenum restrict endpoint to ip-address:port

AxelNennker:


AxelNennker:
There are tests in indy-node and in indy-plenum restricting endpoint to ip address and port
https://github.com/hyperledger/indy-node/blob/master/indy_common/types.py#L193
https://github.com/hyperledger/indy-plenum/blob/master/plenum/common/util.py#L557

PR #1132, opened 2018-09-10
move Endpoint to API because type safety

AxelNennker:

- enhance type safety
- easier to read api for api users

In the future add endpoint validity checks to Endpoint::new so the code fails early if an invalide endpoint is passed into the Indy api

### DIFF
--- a/libindy/src/api/did.rs
+++ b/libindy/src/api/did.rs
@@ -13,6 +13,7 @@ use serde_json;
 use self::libc::c_char;
 
 use std::ptr;
+use domain::ledger::attrib::Endpoint;
 
 
 /// Creates keys (signing and encryption keys) for a new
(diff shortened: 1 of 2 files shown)


AxelNennker:
This is https://jira.hyperledger.org/browse/IS-967 in Jira

libindy/src/api/did.rs
```diff
@@ -416,12 +417,13 @@ pub extern fn indy_set_endpoint_for_did(command_handle: i32,
     trace!("indy_set_endpoint_for_did: entities >>> wallet_handle: {:?}, did: {:?}, address: {:?}, transport_key: {:?}",
            wallet_handle, did, address, transport_key);
 
+    let endpoint = Endpoint::new(address.to_string(), Some(transport_key.to_string()));
```
jovfer: @AxelNennker Can it be just `let endpoint = Endpoint::new(address, Some(transport_key));` ?

AxelNennker:
Could someone do a `cargo update` and commit the new Cargo.lock, please?
Maybe I should not cargo update so often, it's a habit. Then I have to pay attention to not muddle my PRs with the new Cargo.lock.

PR #1133, opened 2018-09-10
Multi-thread libindy for CPU expensive calls

jovfer:
HIPE about [multi-threading in libindy](https://github.com/hyperledger/indy-hipe/tree/master/text/0012-concurrency-improvement).


AxelNennker:
Not sure this optimization is worth the added complexity and the new dependency on threadpool.

EdgeAgents like mobile apps have only a few of not only one wallet for the one user and key derivation does not take that long.
Some is true for validator nodes. Creation of wallet is a rare event.

Cloud Agents might create many wallets when onboarding a new user but this is event is started through a web server which multi-threaded anyway.

Which use case does this optimization address and is it worth it?

jovfer:
@AxelNennker This PR is in development state now (WIP tag in PR title = work in progress)

You can find HIPE about multi-threading in libindy [here](https://github.com/hyperledger/indy-hipe/tree/master/text/0012-concurrency-improvement).

We plan to use some thread pool (most probably it will be different crate) for all long operations. 

In this PR (as [1st phase](https://github.com/hyperledger/indy-hipe/tree/master/text/0012-concurrency-improvement#phase-1-short-term) we plan to support as operation for thread pool next 2 most CPU expensive actions:
* key derivation for wallet (derivation is also used for wallet_open and open is not rare operation)
* anoncreds private keys generation

Please let me know if you have something to add to this list.

libindy/src/api/mod.rs
```diff
@@ -225,3 +225,16 @@ pub enum ErrorCode
     // Extra funds on inputs
     PaymentExtraFundsError = 705
 }
+
+#[no_mangle]
+pub extern fn indy_set_crypto_thread_pool_size(size: usize) -> ErrorCode {
```
vimmerru: I believe it should be called indy_init() and require a json with crypto_thread_pool_size fields as an input.

PR #1138, opened 2018-09-13
Add protocol version config

phillipgibb:
LibVCX can work with protocol-version 2 but the agency that an agent connects to connects to the Sovrin test network which is currently protocol version 1.
In order to connect to the STN with VCX then the protocol version needs to be 1, at the moment it is  hardcoded to 2.


jovfer:
@glowkey could you review this PR?

vcx/libvcx/src/settings.rs
```diff
@@ -181,6 +184,18 @@ pub fn process_config_file(path: &str) -> Result<u32, u32> {
     }
 }
 
+pub fn get_protocol_version() -> usize {
+    let protocol_version = match get_config_value(CONFIG_PROTOCOL_VERSION) {
+        Ok(x) => x.parse::<usize>().unwrap_or(DEFAULT_PROTOCOL_VERSION),
+        Err(x) => DEFAULT_PROTOCOL_VERSION,
+    };
+    if protocol_version > DEFAULT_PROTOCOL_VERSION {
```
jovfer: @phillipgibb is any reason to limit possible protocol version by default one?

phillipgibb: not really, only because the STN was version 1, should be 2 now, but could be different in future.

jovfer: @phillipgibb  But this code *limit* result of the function to default version. In my understanding it's OK to have default one not the latest one, but vcx should allow to set never.

phillipgibb: I agree, but I had to do the code change locally so I could set the version to 1 for STN. I guess it has no point now because STN is going version 2 and I don't know when/if ever there will be a version 3. So maybe reject the pull request

jovfer: I suggest to just drop this particular if/else. So vcx will use 1 by default (or may be 2 as STN already have v2)
+ there will be ability to set any new version via vcx config file

jovfer: May be we should have one more constant - something like `MAX_SUPPORTED_PROTOCOL_VERSION`

phillipgibb: yes, great idea. Yet it will need to be updated whenever the protocol version of the ledger updates - guessing not anytime soon

jovfer: Like
```
let protocol_version = match get_config_value(CONFIG_PROTOCOL_VERSION) {
        Ok(x) => x.parse::<usize>().unwrap_or(DEFAULT_PROTOCOL_VERSION),
        Err(x) => DEFAULT_PROTOCOL_VERSION,
    };
    if protocol_version > MAX_SUPPORTED_PROTOCOL_VERSION {
        err_trace and return default or max ? return critical error?
    } else {
        protocol_version
    }
```

jovfer: or may be
```
   let protocol_version = match get_config_value(CONFIG_PROTOCOL_VERSION) {
        Ok(x) if x <= MAX_SUPPORTED_PROTOCOL_VERSION => x.parse::<usize>().unwrap_or(DEFAULT_PROTOCOL_VERSION),
        Err(x) => ??? critical error ????,
    };
```
not sure about `get_config_value` behavior

phillipgibb: I can't contribute anymore to this pull request, my mastering of rust doesn't warrant too much messing around. 
But I think that  there should be an error, or at least a warning to inform the 'user' that the version is not supported (and may default to the max version?)

PR #1142, opened 2018-09-15
Openssl0.10.12

AxelNennker:
bumped rust openssl version to 0.10.12 which used openssl 1.1.1 which supports ed25519, tls1.3 and has better Android support


jovfer:
@AxelNennker do you suggest to increment system openssl library version (to 1.1.1?) also?

AxelNennker:
Not necessarily. Ubuntu is usually many (minor) versions back.
For Android the 1.1.1 version would need to be on https://repo.sovrin.org/android/libindy-deps/openssl/
We could also try vendored=true if cross-compiling and see if rust-openssl does the right thing.

jovfer:
@vimmerru I think we should create some backlog task for separate migration to newer system openssl.

I suggest to merge this PR after Android unit tests

AxelNennker:
At some point in time, before wallet import and export was implemented, openssl was used for some BigNumber stuff missing in Rust like is_prime, rand etc but currently openssl is also used to compute hashes in utils/crypto/hash/openssl.rs
I think that should those hashes should be computed with the crate sha2 so reducing dependency on openssl. https://jira.hyperledger.org/browse/IS-1002
The remaining bn routines should then replaces by pure rust and openssl removed entirely.

AxelNennker:
> @vimmerru I think we should create some backlog task for separate migration to newer system openssl.
> 
> I suggest to merge this PR after Android unit tests

I think this PR adds to indy-sdk.
As @vimmerru said in https://github.com/hyperledger/indy-crypto/pull/126 this does not change the openssl version used but allows developers who have the latest openssl version to use it.

PR #1152, opened 2018-09-20
nodejs logging support

farskipper:
Added
`setDefaultLogger(pattern)`
`setLogger(logFn)`


vimmerru:
@farskipper Could you solve the conflict? After this we can merge it.

farskipper:
@vimmerru should be good to go now

PR #1166, opened 2018-09-28
initial rust wrapper PR

cam-parra:
Signed-off-by: Cam Parra <caeparra@gmail.com>


wrappers/rust/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "rust-libindy-wrapper"
```
vimmerru: I believe it should be named indy

PR #1170, opened 2018-10-01
[WIP] Pack and Unpack functions aligned with AMES HIPE

kdenhartog:
This is code to align with [AMES HIPE](https://github.com/hyperledger/indy-hipe/pull/43).


kdenhartog:
Originally I was going to build on his work, but I diverged from that and made a [commit](https://github.com/hyperledger/indy-sdk/pull/1170/commits/1033002a635700efeed06258bf8db85c40a9a74d) to remove all the microledger code. Would you prefer that I remove all that work and resubmit PR?

jovfer:
yes, we should avoid to include commit history of different subject. Please clean-up the history and update PR

libindy/src/services/route/mod.rs
```diff
@@ -0,0 +1,523 @@
...
+use std::rc::Rc;
+
+pub struct RouteService { }
+
+impl RouteService {
+    pub fn new() -> RouteService {
+        RouteService {}
+    }
+
+    pub fn pack_msg(&self, plaintext: &str, recv_keys: &Vec<String>, my_vk: Option<&str>, is_authcrypt: bool,
```
dbluhm: This approach definitely works (awesome job, btw, really looking forward to having this) but I think that separating out auth and anon pack out into separate functions may improve overall understandability of this method. Is there a compelling reason to mix both into the one function?

kdenhartog: when I originally designed it I was combining them. With this new iteration I've been rewriting they are now separate making for much cleaner code to read.

kdenhartog:
more updates added to align with the suggestions brought up. The code needs tested better still.(refactoring the old ones as well as adding new ones). I also need to update the API so the wrappers gain this functionality.

jovfer:
@kdenhartog I suggest to try interactive rebase and drop microledgers commits

kdenhartog:
moved rebased branch to a separate branch called "multiplex-rebase" will close this and open a new PR.

jovfer:
moved to #1210

PR #1175, opened 2018-10-02
Refactor Inmem storage shared library and Update unit tests to run against dynamic library

ianco:
Added a new shared library containing inmem wallet storage plug-in (to demonstrate and test dynamic storage loading).

Added python wrapper functions to load wallet storage from shared library.

Updated unit tests to optionally use dynamic storage rather than default via environment parameters:

STG_CONFIG - json configuration string to pass to the wallet on creation and open
STG_CREDS - json credentials string to pass to the wallet on creation and open
STG_TYPE - storage type to create
STG_LIB - c-callable library to load (contains a plug-in storage)
        (if specified will dynamically load and register a wallet storage)
STG_FN_PREFIX - prefix for all plug-in functions (allows standard function naming)

For example to run tests against the inmem wallet:

STG_TYPE=inmem_custom STG_LIB=../samples/storage/storage-inmem/target/debug/libindystrginmem.dylib STG_FN_PREFIX=inmemwallet_fn_ cargo test issuer_create_and_store_credential_def_works

### DIFF
--- a/libindy/Cargo.lock
+++ b/libindy/Cargo.lock
@@ -1,5 +1,13 @@
 [[package]]
 name = "aho-corasick"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "memchr 0.1.11 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
+name = "aho-corasick"
 version = "0.6.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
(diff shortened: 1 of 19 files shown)


ianco:
Note that this PR is superceded by https://github.com/hyperledger/indy-sdk/pull/1238 (Postgres wallet storage plug-in), which includes all the changes in this PR.

vimmerru:
@ianco 

See my comments in HIPE https://github.com/hyperledger/indy-hipe/pull/49. I believe we need to align what we want to achieve before deep discussion of code.

PR #1177, opened 2018-10-03
Majorminor

AxelNennker:
unclear how the 'opened wallet' commit ended up in this PR.
Can you cherry-pick?


vimmerru:
When we write 0.10.12 it means that major.minor version should be fixed and patch version can be increased with cargo update.

We use lock file that locks exact versions, but consider cargo update to be regular procedure to get all available patches.

Allowing minor versions to be updated looks much more dangerous procedure and requires analysis of each crate. A lot of crates don't have strong release process. can avoid some SemVer rules and introduce security issues.

In my vision such updates are better to perform manually. We discussed this inside of core team and suggest to keep current process with fixing minor version.

@AxelNennker I am closing this PR. If you still need further discussion we can re-open it. Also if you want to updated some crates to achieve specific goal and analysed the changes we can accept these PRs.

PR #1178, opened 2018-10-04
[WIP] Added boilerplate for VCX compatible dummy agent

vimmerru:
It is still non-ideal and requires some cosmetic refactoring, but it exposes working endpoints, illustrates complete async flow and should allow to start messages handling implementation in parallel.


vcx/dummy-cloud-agent/sample-config.json
```diff
@@ -0,0 +1,17 @@
+{
+  "agent": {
+    "wallet_id": "agency_wallet_id",
```
e-nem-y: Could we change `agency` to `agent` ?

PR #1179, opened 2018-10-04
Supported hexadecimal seed

Artemkaaas:


libindy/src/services/crypto/mod.rs
```diff
@@ -348,22 +351,33 @@ impl CryptoService {
...
+        let seed = seed.unwrap();
+
+        let bytes = if seed.as_bytes().len() == ed25519_sign::SEEDBYTES {
+            // is acceptable seed length
+            seed.as_bytes().to_vec()
+        } else if seed.ends_with('=') {
+            // is base64 string
+            base64::decode(&seed)
+                .map_err(|err| CommonError::InvalidStructure(format!("Can't deserialize Seed from Base64 string: {:?}", err)))?
+        } else if seed.as_bytes().len() / 2 == ed25519_sign::SEEDBYTES {
```
vimmerru: It should be 

```
if seed.as_bytes().len() == ed25519_sign::SEEDBYTES * 2
```

PR #1180, opened 2018-10-04
Store tails from generator

AxelNennker:
This PR handles https://jira.hyperledger.org/browse/IS-1026

tails.rs has a sha256 hasher but finalize is never called and removing the hasher code has no effect on `RUST_TEST_THREADS=1 cargo test lib`.
 
A hasher is created here:
https://github.com/hyperledger/indy-sdk/blob/master/libindy/src/commands/anoncreds/tails.rs#L82
and for this hasher finalize is never called.

The code calls service.finalize but that calls finalize on a different hasher.
 
This bug is not caught by any test.


AxelNennker:
Trying to get rid of the superfluous commits. So closing this and trying again

PR #1181, opened 2018-10-04
fix store_tails_from_generator hash computation

AxelNennker:

tails.rs has a sha256 hasher but finalize is never called on that hasher and removing the hasher has no effect on `RUST_TEST_THREADS=1 cargo test lib`.
 
A hasher is created here:
https://github.com/hyperledger/indy-sdk/blob/master/libindy/src/commands/anoncreds/tails.rs#L82
and for this hasher finalize is never called.

The code calls service.finalize but that calls finalize on a different hasher.
 
This bug is not caught by any test.

This fixes https://jira.hyperledger.org/browse/IS-1026


libindy/src/commands/anoncreds/tails.rs
```diff
@@ -77,10 +77,10 @@ pub fn store_tails_from_generator(service: Rc<BlobStorageService>,
                                   rtg: &mut RevocationTailsGenerator) -> Result<(String, String), CommonError> {
     debug!("store_tails_from_generator >>> writer_handle: {:?}",writer_handle);
 
-    let blob_handle = service.create_blob(writer_handle)?;
-
     let mut hasher = sha2::Sha256::default();
 
+    let blob_handle = service.create_blob(writer_handle, hasher)?;
```
jovfer: In my understanding at this moment `Copy` will occurs for `hasher`. As result code in current PR is equivalent to original.

AxelNennker: it turns out that the hasher in tails.rs is superfluous and just a waste of time. 
New PR to follow soon

PR #1182, opened 2018-10-04
Indy-SDK 1.6.7

Artemkaaas:


wrappers/rust/README.md
```diff
@@ -0,0 +1,32 @@
+# indy
```
jovfer: * Add notes about experimental status of rust wrapper here
* look at original readme of rust wrapper to find more details and include it here

PR #1187, opened 2018-10-04
Replaced usages of env::home_dir on dirs::home_dir

Artemkaaas:


jovfer:
@AxelNennker  yes, nested extern crate here is required by re-usage utilities from `src` in `tests`.

PR #1190, opened 2018-10-05
Fix submit action for empty list nodes to send.

jovfer:


jovfer:
Will be merged as part as #1169

PR #1209, opened 2018-10-10
Updates Indy Crate to 1.6.7

hadleym:
@e-nem-y 
If this overlaps with any work/PR that you're doing, feel free to just close this PR, but the sooner this item is updated in the Cargo.toml the sooner we can move forward with more Libvcx development.


e-nem-y:
@hadleym I do not think that merging this will create any problems.

PR #1210, opened 2018-10-10
Pack and Unpack functions

kdenhartog:
This is the updated rebased code which removes microledger code and has history of the changes along the way. Some were messed up while rebasing, but the majority are there. This code is aligned with the [AMES HIPE](https://github.com/hyperledger/indy-hipe/pull/43).


jovfer:
@kdenhartog there are still some history related to microledgers
https://github.com/hyperledger/indy-sdk/pull/1210/commits/911b6589ab89fca24266175fa62f505f77ce527c
https://github.com/hyperledger/indy-sdk/pull/1210/commits/6c9569dcbc00d0701532c70049397313c4a764e5#diff-c97552f5bd97e0797ff7128a93d65025

Can it be deleted? Most probably it can be done by interactive rebase
1) select edit for commit 6c9569d and drop for 911b658
2) remove microledger file from the commit, amend
3) continue/finish rebase 
4) double check: result HEAD should be same with current one

Fill free to slack me if you have any doubts here or problems while rebase

kdenhartog:
@jovfer thanks for catching that. I'm glad you double checked it because I was having issues rebasing back that far. That whole commit can be removed so I'll fix it. Would squashing the commits together help?

@vimmerru I'll go in and add these changes. Are these requirements documented anywhere? I had a hard time finding details like this as I was working on it. I'm thinking this should be linked in a CONTRIBUTORS.md link describing these sorts of "tribal knowledge" type things. By that I mean, these are things that the maintainers know, but anyone who wants to contribute couldn't figure this out without making a PR or getting caught up to speed by a maintainer.

kdenhartog:
@vimmerru I'm running into some issues with adding these tests you've requested. How should I go about adding a DID that has more than one key? I know how to do this through the command layer, but I'm not finding anything that is available at the API layer.

libindy/src/services/crypto/mod.rs
```diff
@@ -425,6 +452,183 @@ impl CryptoService {
 
         Ok(res)
     }
+
```
kdenhartog: This through line 631 doesn't seem like it would fit in the crypto_service now, but I'm not exactly sure where the best place to put them would be. for the recipient focused functions, I plan to move those into the command layer which shouldn't be a problem. For the other two symmetrical crypto encrypt/decrypt functions though these don't feel like they fit, but I don't know a better place to put them. Any suggestions? @jovfer @vimmerru

kdenhartog:
@vimmerru @jovfer can you guy's take a look at my updates? I'm working on adding wrapper support next. The APIs are exposed now and tested using high cases module.

kdenhartog:
I'll add the work from PR 1314 into this PR as well.

libindy/src/commands/crypto.rs
```diff
@@ -116,29 +140,46 @@ impl CryptoCommandExecutor {
...
         self.crypto_service.validate_key(my_vk)?;
 
-        let key: Key = self.wallet_service.get_indy_object(wallet_handle, &my_vk, &RecordOptions::id_value())?;
+        let key: Key = self.wallet_service.get_indy_object(
+            wallet_handle,
+            &my_vk,
+            &RecordOptions::id_value(),
+        )?;
 
         let res = self.crypto_service.sign(&key, msg)?;
```
mikelodder7: If `key` is a private key, it needs to be cleared after sign.

libindy/src/commands/crypto.rs
```diff
@@ -162,118 +203,400 @@ impl CryptoCommandExecutor {
...
+        let my_key: Key = self.wallet_service.get_indy_object(
+            wallet_handle,
+            my_vk,
+            &RecordOptions::id_value(),
+        )?;
 
-        let res = self.crypto_service.encrypt_sealed(&their_vk, &msg)?;
+        let res = self
+            .crypto_service
+            .authenticated_encrypt(&my_key, their_vk, msg)?;
```
mikelodder7: If `key` is a private key, it needs to be cleared here.

libindy/src/commands/crypto.rs
```diff
@@ -162,118 +203,400 @@ impl CryptoCommandExecutor {
...
-        let doc: Vec<u8> = base64::decode(&parsed_msg.msg)
-            .map_err(|err| CommonError::InvalidStructure(format!("Can't decode internal msg filed from base64 {}", err)))?;
-
-        let nonce: Vec<u8> = base64::decode(&parsed_msg.nonce)
-            .map_err(|err| CommonError::InvalidStructure(format!("Can't decode nonce from base64 {}", err)))?;
-
-        let decrypted_msg = self.crypto_service.decrypt(&my_key, &parsed_msg.sender, &doc, &nonce)?;
-
-        let res = (parsed_msg.sender, decrypted_msg);
+        let res = self.crypto_service.authenticated_decrypt(&my_key, &msg)?;
```
mikelodder7: If `key` is a private key, it needs to be cleared.

libindy/src/commands/crypto.rs
```diff
@@ -162,118 +203,400 @@ impl CryptoCommandExecutor {
...
+        let my_key: Key = self.wallet_service.get_indy_object(
+            wallet_handle,
+            &my_vk,
+            &RecordOptions::id_value(),
+        )?;
 
-        let res = self.crypto_service.decrypt_sealed(&my_key, &encrypted_msg)?;
+        let res = self
+            .crypto_service
+            .crypto_box_seal_open(&my_key, &encrypted_msg)?;
```
mikelodder7: If `my_key` is a private key, it needs to be cleared.

jovfer:
@kdenhartog there are some problems in git history of this PR. I'm going to rebase it and make final state same to the current one

kdenhartog:
Sounds good. Thanks for fixing that.

I did find one mistake that crept in. Right now ComboBox which is used by `authenticated_encrypt ()`has a dependency on message_pack encoding rather than base64url encoding. This should be changed, but I believe it will cause a breaking change to the original Authcrypt function. 

Instead of breaking the old functionality to support new, I'm thinking the best path forward is to create two `ComboBox` functions. Then when we decide `authcrypt` functions are able to be deprecated, we can remove the old `ComboBox`. 

What's your thoughts on this approach @jovfer

libindy/src/domain/crypto/combo_box.rs
```diff
@@ -1,18 +1,6 @@
...
 }
-
-impl ComboBox {
-    pub fn to_msg_pack(&self) -> Result<Vec<u8>, rmp_serde::encode::Error> {
-        rmp_serde::encode::to_vec_named(self)
-    }
-
-    pub fn from_msg_pack(bytes: &[u8]) -> Result<ComboBox, rmp_serde::decode::Error> {
-        rmp_serde::decode::from_slice(bytes)
-    }
```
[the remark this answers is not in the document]
kdenhartog: I brought back the old structure where this is handled in the command layer for the old APIs. I left the authenticated_encrypt in the service layer for the new pack/unpack APIs. Primary reason for this was to keep the size of the command functions smaller and more readable. Also, can you double check that when I pushed the new changes that it didn't rebrake the commit tree? I checked and didn't see any of the microledger related commits.

libindy/src/utils/crypto/base64/rust_base64.rs
```diff
@@ -3,11 +3,11 @@ extern crate base64;
 use errors::common::CommonError;
 
 pub fn encode(doc: &[u8]) -> String {
-    base64::encode(doc)
+    base64::encode_config(doc, base64::URL_SAFE)
```
jovfer: As far as I know this `URL_SAFE` alphabet is different for base64 against default one. In this case we must not to change this utility function behavior as it widely used in codebase

libindy/src/utils/json.rs
```diff
@@ -0,0 +1,19 @@
+extern crate serde;
+extern crate serde_json;
+
+use self::serde::{Serialize, Deserialize};
+use self::serde_json::Error;
+use std::string::String;
+
+
+pub trait JsonEncodable: Serialize + Sized {
```
jovfer: This trait was eliminated some times ago as useless, do we have some new reasons to restore it?

libindy/src/commands/crypto.rs
```diff
@@ -209,50 +288,368 @@ impl CryptoCommandExecutor {
...
+        })?;
+
+        //break early and error out if no receivers keys are provided
+        if receiver_list.is_empty() {
+            return Err(IndyError::CommonError(CommonError::InvalidParam4(format!(
+                "No receiver keys found"
+            ))));
+        }
+        
+        match sender_vk.is_empty() {
```
jovfer: I suggest to replace this `match` by `if`

libindy/src/commands/crypto.rs
```diff
@@ -209,50 +288,368 @@ impl CryptoCommandExecutor {
...
+        let (ciphertext, iv, tag) =
+            self.crypto_service
+                .encrypt_plaintext(message, &base64_protected, &cek);
+
+
+        self._format_pack_message(&base64_protected, &ciphertext, &iv, &tag)
+    }
+
+    fn _base64_encode_protected(&self, encrypted_recipients_struct: Vec<Recipient>, alg_is_authcrypt: bool) -> Result<String> {
+        let alg_val= match alg_is_authcrypt {
```
jovfer: I suggest to replace `match` by `if` here

libindy/src/commands/crypto.rs
```diff
@@ -209,50 +288,368 @@ impl CryptoCommandExecutor {
...
+                err
+            )))
+        })?;
+
+        //extract recipient that matches a key in the wallet
+        let (recipient, is_auth_recipient) = self._find_correct_recipient(protected_struct, wallet_handle)?;
+
+        //get cek and sender data
+        let (sender_verkey_option, cek) =
+            match is_auth_recipient {
```
jovfer: I suggest to replace this `match` by `if`

libindy/src/commands/crypto.rs
```diff
@@ -209,50 +288,368 @@ impl CryptoCommandExecutor {
...
+        for recipient in protected_struct.recipients {
+            let my_key_res : Result<Key> = self.wallet_service.get_indy_object(
+                wallet_handle,
+                &recipient.header.kid,
+                &RecordOptions::id_value()
+            ).map_err(|err| IndyError::WalletError(err));
+
+
+            if my_key_res.is_ok() {
+                match recipient.header.sender {
```
jovfer: I suggest to replace this `match` by something like
`Ok((recipient, recipient.header.sender.is_some()))`

libindy/include/indy_crypto.h
```diff
@@ -317,6 +317,123 @@ extern "C" {
...
+    ///
+    /// #Params
+    /// command_handle: command handle to map callback to user context.
+    /// message: a pointer to the first byte of the message to be packed
+    /// message_len: the length of the message
+    /// receivers: a string in the format of a json list which will contain the list of receiver's keys
+    ///                the message is being encrypted for.
+    ///                Example:
+    ///                "[<receiver edge_agent_1 verkey>, <receiver edge_agent_2 verkey>]"
+    /// sender: the sender's verkey as a string When "" is used in this parameter, anoncrypt is used
```
jovfer: I suggest to use `null` instead of empty string to switch between modes.

jovfer: There is another option to split into 2 separate methods but I prefer current approach (determine type by parameter value) instead of old one (`anon/auth_crypt`)

@vimmerru any thoughts?

kdenhartog: The consumers of this API (particularly the agent community) do not want a separate API.This is why it's been combined. In the case of using a null pointer rather than an empty string, I don't understand how to implent this.

libindy/src/utils/json.rs
kdenhartog: As far as I know, the only reason this still exists is because this PR is so old that most the changes you've made since I started working on this haven't been brought in.

jovfer: hm, we have some merge into this PR, may this deletion is missed. Lets come back to this question after final merging latest master to the PR

libindy/include/indy_crypto.h
kdenhartog: Fixed this, need to test this works with wrappers, but given we're doing it in other modules I assume it should be fine.

libindy/src/commands/crypto.rs
kdenhartog: added this suggestion

libindy/src/commands/crypto.rs
kdenhartog: Kept this as a match because the api now uses opt_c_str!() macro

libindy/src/utils/crypto/base64/rust_base64.rs
kdenhartog: fixed this by adding `encode_urlsafe()` and `decode_urlsafe()` functions and bringing the old implementations back.

libindy/src/api/crypto.rs
```diff
@@ -557,4 +561,190 @@ pub  extern fn indy_crypto_anon_decrypt(command_handle: IndyHandle,
...
+
+    let res = result_to_err_code!(result);
+
+    trace!("indy_auth_pack_message: <<< res: {:?}", res);
+
+    res
+}
+
+
+/// Unpacks a message packed using indy_pack_message which follows the wire message format
```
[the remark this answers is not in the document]
kdenhartog: I've added experimental to this so that consumers know that it's experimental. From our discussion earlier though because this is only using keys and key referencing is being handled at a different layer, I don't think this will cause problems.

libindy/src/utils/json.rs
kdenhartog: I was able to remove this without any problems. All my tests are still passing and I haven't seen any new ones fail either.

PR #924, opened 2018-07-05
[WIP] Improvements/expanding wallet test coverage

keichiri:

PR #1213, opened 2018-10-11
IS-1028 Wallet unit test coverage

Artemkaaas:


jovfer:
Moved from #924

PR #1217, opened 2018-10-15
Implement logging integration and update test suite to match the Java suite

srottem:
Logging has been implemented using the LibLog package which adds internal features for logging that will work with many of the .NET logging frameworks. This has necessitated updating the build target from netstandard 1.1 to 2.0, so some backward compatibility has been lost.

While updating I noticed that some previous work on the wrapper by Matt Raffel started trying to introduce strongly typed parameter objects and that a dependency on Newtonsoft.Json had been added for the purpose. I don't believe this is a good idea to have the wrapper take on 3rd party dependencies, so I have dropped it and commented out the code that was using it - it only featured in convenience methods related to the Wallet API and it doesn't appear to be critical.  If this is determined not to be the correct course it will be simple to re-implement or clean up later.

Note: the tests of the wallet export and import features fail - this may be a Windows related issue as it also fails in the Java tests.


patch.diff
```diff
@@ -0,0 +1,15302 @@
+From 4c71e13695a1b18cd76de6f340ca023ea5ca4399 Mon Sep 17 00:00:00 2001
```
jovfer: @srottem is this file redundant?
If yes - please remove it from history (rm, amend commit + force push)

srottem: Removed - apologies.

jovfer: thanks

PR #1223, opened 2018-10-15
Remove statically compiled payment plugins

e-nem-y:
- Remove libnullpay and libsovtoken from code
- Remove references to statically compiled plugins
- Modify tests to expect a dynamicaly loaded payment plugin
- Modify build scripts for linux, android and ios
- Modify wrappers to reflect code changes
- Make Xenial build pass

Signed-off-by: Nemanja Patrnogic <patrnogicnemanja@gmail.com>


vcx/libvcx/src/api/vcx.rs
```diff
@@ -76,10 +76,14 @@ pub extern fn vcx_init (command_handle: u32,
         } else {
             match settings::process_config_file(&config_path) {
                 Err(e) => {
-                    println!("Invalid configuration specified: {}", e);
                     return error::INVALID_CONFIGURATION.code_num;
                 },
-                Ok(_) => (),
+                Ok(_) => {
+                    match settings::validate_payment_method() {
```
Artemkaaas: As far as I  understand VCX should be able to work without any payment plugins. So, `CONFIG_PAYMENT_METHOD` should be an optional field. And it's better to move it validation to `settings/validate_config` function.

e-nem-y: > And it's better to move it validation to `settings/validate`_config function.

The problem there is that there is another init function that calls `validate _config` which does only wallet export/import logic and does not require a plugin to be loaded for it to function. This one on the other hand is called on App load so it made sense to check it only here -- I tried with he function in the `validate_config` and it broke half of the tests

vcx/libvcx/src/api/vcx.rs
```diff
@@ -95,11 +99,6 @@ fn _finish_init(command_handle: u32, cb: extern fn(xcommand_handle: u32, err: u3
 
     ::utils::threadpool::init();
 
-    match ::utils::libindy::payments::init_payments() {
```
glowkey: It seems that the payment plugin is not being initialized anywhere.  In testing it gets initialized inside ::utils::devsetup::tests::init_plugin() but that function is for testing only.  How is the payment plugin initialized outside of testing?

vcx/libvcx/src/settings.rs
```diff
@@ -55,6 +56,10 @@ pub static DEFAULT_WALLET_KEY: &str = "foobar1234";
 pub static DEFAULT_THREADPOOL_SIZE: usize = 8;
 pub static MASK_VALUE: &str = "********";
 pub static DEFAULT_WALLET_KEY_DERIVATION: &str = "ARGON2I_INT";
+pub static DEFAULT_PAYMENT_PLUGIN: &str = "libnullpay.so";
+pub static DEFAULT_PAYMENT_INIT_FUNCTION: &str = "nullpay_init";
+pub static DEFAULT_PAYMENT_METHOD: &str = "null";
```
glowkey: If somebody wanted to initialize the "libsovtoken" payment plugin would all three have to be configured?  Are there configuration options for all three different values?

vcx/libvcx/src/settings.rs
```diff
@@ -245,6 +250,23 @@ pub fn get_wallet_credentials() -> String {
...
 
+pub fn validate_payment_method() -> Result<(), u32> {
+    let config = SETTINGS.read().unwrap();
+    if let Some(method) = config.get(CONFIG_PAYMENT_METHOD) {
+        if !method.to_string().is_empty() {
+            return Ok(());
+        }
+    }
+    return Err(error::MISSING_PAYMENT_METHOD.code_num);
+}
```
glowkey: It seems this function should try to load the payment plugin to test if it is valid, not just if it is present.

e-nem-y: If the plugin is loaded by the user and can be custom, how should we validate? Three char string?

vcx/libvcx/src/api/vcx.rs
e-nem-y: Application using libvcx needs to load the plugin beforehand and call init on the plugin for it to bind with libindy.

vcx/libvcx/src/settings.rs
glowkey: I think we should try to initialize the plugin using the init function.  If that fails then the plugin is invalid.

vcx/libvcx/src/settings.rs
e-nem-y: Yes, if libsovtoken is to be used these should be changed.

Config options are not available because only method is something that needs to be configured for libvcx, these DEFAULTS_* are just for testing purposes.

vcx/libvcx/src/settings.rs
e-nem-y: Ticked I was working on states that initialization of the plugin is the responsibility of the user application not libvcx:  https://jira.hyperledger.org/browse/IS-934

vimmerru: @glowkey @e-nem-y 

> It seems this function should try to load the payment plugin to test if it is valid, not just if it is present.

Loading can't be responsibility of vcx as it is impossible on the most of platforms. If it is possible we still don't know plugin placement and dll name. 

Initialization can be responsibility of vcx, but in this case we need to provide the name of init method. It looks very unsafe and unnatural. I am not sure that it is possible to call static linked function by name. For loaded plugins it is possible, but on Android and iOS we link statically.

vimmerru: One correction. It is possible to call only function by name when we load dll/so from our program, dynamic linking that loads so on program start doesn't allow us it too.

glowkey: If this is the case, that application needs to load plugin itself, then libnullpay (and libsovtoken) will need to create wrappers for the one "init" function.  Correct?

vimmerru: In my vision working workflow is the following:

- application links with a plugin or loads it
- application calls init function
- application provides the name of payment method to vcx config

vimmerru: No, but for dynamic languages we can provide generic API like:

```
- init_plugin(init_function_name)
- load_plugin(dll path)
```

glowkey: Should that wrapper work be done as part of this PR?

vimmerru: I believe it is unnecessary. For example, loading plugin from python or Java is a trivial task (few lines of code). As a result we already can get valuable demo.

vimmerru: My main concern is about ConnectMe.

e-nem-y:
What are the next steps regarding this PR @vimmerru? What needs to be changed?

glowkey:
Knowing this will break downstream components in a non-trivial way I think we either need to inform them and make a plan or do some more work on this PR to make it less disruptive.

jovfer:
@glowkey review status from you is marked as "changes requested" . Is it actual or outdated? Please update it if needed.
FYI @e-nem-y I would like to see approve from Dough first as my review will override final review status

PR #1242, opened 2018-10-24
Add note about file/socket limit for MacOS tests of libindy.

jovfer:


doc/mac-build.md
```diff
@@ -43,3 +43,13 @@
...
 In order to run local nodes on MacOS, it may be necessary to set up port mapping between the Docker container
 and local host. Follow the instructions in [Indy SDK README](https://github.com/hyperledger/indy-sdk#how-to-start-local-nodes-pool-with-docker)
+
+# IOError while running of whole set of tests on MacOS
+
+There is a possible case when some tests are failed if whole set of tests is run (`cargo test`).
+But failed tests will be successful in case of separate runs.
+If an error message like `IOError` `Too many open files` is present in logs when fails can be fixed by changing default limit.
+
+`ulimit -n <new limit value>`
```
vimmerru: @jovfer DO we need some guidance for exact linit value?

jovfer: It may depends on different params and it's a subject to change in the future as we can reorganize tests.
So we can estimate some very rough top limit, but I don't like this option

PR #1243, opened 2018-10-24
cli "did rotate-key" command should support resume mode

dkulic:
Added tempVerkey inside of result json of `indy_get_my_did_with_meta` call.
Added resume parameter to cli "did rotate-key" command.
Improved tests for cli "did rotate-key" command.


cli/src/libindy/did.rs
```diff
@@ -39,6 +39,7 @@ impl Did {
     }
 
     pub fn replace_keys_apply(wallet_handle: i32, did: &str) -> Result<(), ErrorCode> {
+        println!("replace_keys_apply({}, {})", wallet_handle, did);
```
jovfer: Please remove temporary line

cli/src/commands/did.rs
```diff
@@ -326,9 +394,9 @@ pub mod tests {
     pub const DID_MY1: &'static str = "VsKV7grR1BUE29mG2Fm2kX";
     pub const VERKEY_MY1: &'static str = "GjZWsBLgZCR18aL468JAT7w9CZRiBnpxUPPgyQxh4voa";
 
-    pub const SEED_MY2: &'static str = "00000000000000000000000000000My2";
-    pub const DID_MY2: &'static str = "2PRyVHmkXQnQzJQKxHxnXC";
-    pub const VERKEY_MY2: &'static str = "kqa2HyagzfMAq42H5f9u3UMwnSBPQx2QfrSyXbUPxMn";
+//    pub const SEED_MY2: &'static str = "00000000000000000000000000000My2";
```
jovfer: If MY2 group is no longer required please remove it.

dkulic: I left it because of order (now we have 1, 3, 4..), so later may be reused. :)

PR #1282, opened 2018-11-09
Postgres wallet storage plugin

ianco:
Originally in PR 1247, re-factored into a separate PR

This PR contains stand-alone plug-ins for in-mem and postgres wallet storage

Also includes a shared library "libindy-common" with code dependencies (copied) from libindy.  This is shared between the 2 plug-ins.

Depends on PR 1273 which includes a bug fix in the libindy storage plug-in (JIRA IS-1045)


vimmerru:
I don't like idea of introducing new top level library that will require dedicated testing, packages creation and deployment. This library has very small auditory - people who want to write storage plugin in Rust, but we can put this library inside of libindy and make it part  of libindy without copy/paste of code. What do you think?

vimmerru:
@ianco 

2. The second question is about your vision of contributed plugins. Do you provided them only as example or assume long term support?

vimmerru:
@ianco 

3. I still don't like idea of usage libindy tests as plugin's test harness. Can we remove this functionality from PR? It will significantly help in merging without reducing a lot of value.

In my vision test harness should be dedicated tool with well defined tests that will guide developers. It doesn't necessary mean copy/paste of code as we can use the same tool for testing of default storage as part of our CI/CD.

ianco:
Hi @vimmerru  thanks for the feedback!  I agree it's better to not introduce a new top-level library.  Most of the code is already in libindy - I copied it into a new library (instead of copying it into the postgres plug-in library) to try to avoid duplicate code, and to avoid a dependency on libindy.

The code in this new library is mainly from three places - libindy's util package, the error package, and the service/wallet package (and sub-packages).

If I can make some suggestions:

- Make the libindy util package (and sub-packages) public, then this code is available to third-party libraries like the postgres plug-in
- Likewise make the errors package public
- Create a new util/wallet package (and make it public), and move any wallet code which is generic (i.e. can be used across different plug-ins) to this new package (language.rs and storage.rs)

What do you think?  I can make these changes in the PR.

ianco:
2.  For in-mem, just as example.  This is just a copy of the existing util/inmem.rs plug-in re-packages as a stand-alone library.  For postgres, long-term support.  This is the plug-in BC is using in production, so we don't have any choice to support it!  :-D  However we would like to make it available to the community.

3.  I agree with your feedback re test harness.  This PR (1282) has removed this functionality (it is still in the previous PR 1247).  PR 1282 only contains the plug-in and code directly required to run the plug-in.

Note there are two other functions in the original 1247 PR that I will submit as separate PR's once this plug-in PR is merged:  updates to the python wrapper and CLI to support register plug-ins.

vimmerru:
> What do you think? I can make these changes in the PR.

It is important to understand that libindy isn't Rust crate. The only artifact that should be consumed is dynamic or static C-compatible library with only pub functions from api module, all another function will be mangled. 

So it is impossible to share any Rust-specific code or making it public. What we can do is create libindy-common inside of libindy folder and add libindy-common crate to dependencies of libindy.

Anyway it is too far from ideal solution as we make public a lot of private unstable interfaces. Copy paste can work better for this case.

@jovfer What do you think?

ianco:
OK, I see.  Probably makes sense to create libindy-common inside of libindy folder then.

vimmerru:
@ianco @jovfer 

I spend some time for thinking and now i am not sure that introduction of libindy-common this way will help us a lot:

1. Interfaces inside of libindy-common are non-public. They can be very dynamic and often part of refactoring. For example, we plan to switch to more modern error handling approach soon. 
2. Moving them to local shared crate will make them public and cause that any libindy refactoring will affect all dependent projects. I don't want to break samples during any refactoring of libindy.
3. Only storage plugins samples are main consumers of proposed shared crate now will consume only smart part of shared interfaces. 
4. It can make vision that you have to use Rust to write libindy plugins.
5. It will be hard to consume  libindy-common outside of libindy repo for creation of 3d party plugins. It will cause copy\paste anyway.

As a result i suggest to copy paste minimum functionality you need and remove libindy-common at all.

vimmerru:
@ianco 

We can consider creation of separate repo called something like libindy-wallet-storage-boilerplate and make them as community maintained project if you still really want to avoid any copy-paste.

vimmerru:
@ianco

I have feeling that having both "in-memory" and "postgress" plugins as a samples is a bit too much as we need to support more code in core project. Also "in-memory" plugin was designed as a test tool only and it isn't a good example from my point of view. It doesn't implement full wql and breaks some core assumptions about wallet persistence. I suggest the following:

1. Keep "in-memory" inside of libindy/tests as we have it now.
2. Use only "postgress" plugin as a sample.

Ideally i want to have "postgress" plugin in a dedicated repo maintained and published by community member outside of the core team and available as a binary artifact. Libindy can have the link to this repo in the root Readme.md.

Having "postgress" inside of samples can cause:

1. It will not be included to CI/CD
2. It can be outdated

vimmerru:
@ianco 

> As described above, a new command has been added to the CLI to register a new plug-in storage. Once registered, this can be used to create and open a new wallet, and supports all wallet functions.

CLI already supports loading any king of wallet storage plugins. You don't need to provide anything new. How it is designed to work:

1. Each libindy plugin should provide only one public function without params. Usually it is called *{plugin_name}_init()*. This function internally calls register_wallet_storage_type or register_payment_method. Corresponded callbacks usually mangled and doesn't available for direct call.
2. Application calls *{plugin_name}_init()* at the beginning and starts usage of registered storage types or payment methods.
3. cli provides command

```
indy> load-plugin library=<library-name> initializer=<init-func-name>
```

Corresponded design is here: https://github.com/hyperledger/indy-sdk/tree/master/doc/design/006-cli-plugins

For example of implementation you can look to libnullpay

samples/storage/storage-postgres/src/postgres_wallet.rs
```diff
@@ -0,0 +1,1781 @@
...
+lazy_static! {
+    // cache of active Postgres searches
+    // TODO figure out  athread-safe PostgresStorageIterator
+    // static ref POSTGRES_ACTIVE_SEARCHES: Mutex<HashMap<i32, Box<::postgres_storage::PostgresStorageIterator>>> = Default::default();
+    static ref POSTGRES_ACTIVE_SEARCHES: Mutex<HashMap<i32, PostgresWalletRecordSet>> = Default::default();
+}
+
+pub struct PostgresWallet {}
+
+impl PostgresWallet {
```
vimmerru: As i understand it is a public interface of plugin. You designed it different way that we are thinking. I suggest to check https://github.com/hyperledger/indy-sdk/tree/master/doc/design/006-cli-plugins and libnullpay as an available example.

vimmerru: I suggest don't use structs to group functions. Modules are designed for this in Rust. There is some old examples of this approach in our codebase, but now we are working on removing this.

esplinr:
Thanks Ian for working on this useful feature. We also appreciate that you created related JIRA issues in the Indy SDK project and submitted a HIPE so that we can discuss the various items included in this PR.

For future reference, the related discussions will also be useful:

https://github.com/hyperledger/indy-hipe/pull/49
https://jira.hyperledger.org/browse/IS-1049
https://jira.hyperledger.org/browse/IS-1052

cam-parra:
@vimmerru do we have any updates on this? I want to propose a nightly branch to be made so that we can unblock this PR

ianco:
@cam-parra  I need to review the requested changes I just haven't had time yet ...  hopefully sometime next week

cam-parra:
@ianco Thanks! I will be keeping an eye on this to help with the merge process

vimmerru:
@ianco Thanks for the update! Will review ASAP.

vimmerru:
@ianco @jovfer 

I believe we can accept PR in current state, but it requires some cleanup later:

1. Copy/paste is much more deep than required. It makes code hard to understand and cause misleading. For example, storage don't need any cryptography and don't know anything about encoding, we don't need so deep error handling and etc...
2.  Current code organization is also caused by copy/paste and can be simplified a lot
3. Correct lazy loading and avoiding of mutex-es

PR #1283, opened 2018-11-09
[WIP] New Inequality Predicates 

mikelodder7:
Once the indy-crypto crate at crates.io has been updated to 0.4.4 then Cargo.toml can be updated to point to it. Until then this won't pass the CI pipeline without Cargo pointing to the current version at https://github.com/hyperledger/indy-crypto


jovfer:
@mikelodder7 colud you try to specify Dev build of Indy crypto instead of git reference?
Like `indy-crypto = "0.4.4-dev-71"`

jovfer:
@mikelodder7 `high_cases::verifier_verify_proof::verifier_verify_proof_works_for_correct_proof` is failed on Jenkins. Could you double-check in your local env?

mikelodder7:
Okay I updated that test case and another one that were out of date. Those passed before but once I did a new pull against hyperledger/indy-sdk I saw the problem.

jovfer:
@mikelodder7 fix for tests seems like breaking-change in JSON API for application. In other words, new verifier/proover will be incompatible with old proover/verifier.

I've briefly review update for IndyCrypto and find https://github.com/hyperledger/indy-crypto/blame/master/libindy-crypto/src/cl/mod.rs#L847 seems like it's a root reason, isn't it?

vimmerru:
I added WIP until we finalize versioning approach and fix backward compatibility problems.

esplinr:
We are excited for these new inequality predicates, but the implementation would require existing proofs to be reissued. We either would need to deprecate the current signature type (CL=Camenish-Lisyanskaya), or do the work necessary to support it and a new signature type (CL2).

We eventually want to support multiple signature types. When a new one is added, the old one becomes deprecated and cannot be used for new credentials. We need to evaluate that scope of work before we can finalize these changes.

mikelodder7:
This isn’t a signature change. It’s a proof request proof response change. Those need to be versioned.

esplinr:
Yes they do, but the work has not yet been done.

jovfer:
This PR is paused, please find more details in https://jira.hyperledger.org/browse/IS-1131

PR #1286, opened 2018-11-13
[dotnet wrapper] Use value types for entity handles

tmarkovski:
This PR changes the handle type (wallet handle, search handle, etc) from IntPtr to int. This fixes an intermittent issue that occurs and results in "wallet handle invalid" exception. It also in line with the internal rust handle representation.
 Java wrapper uses the same approach.


mattraffel:
I've asked @srottem to look over these changes.   Can we hold off on a merge until we hear back?

wrappers/dotnet/indy-sdk-dotnet/AnonCredsApi/AnonCreds.cs
```diff
@@ -225,14 +225,14 @@ private static void ProverGetCredentialsCallbackMethod(int xcommand_handle, int
 #if __IOS__
         [MonoPInvokeCallback(typeof(ProverSearchCredentialsCompletedDelegate))]
 #endif
-        private static void ProverSearchCredentialsCallbackMethod(int xcommand_handle, int err, IntPtr search_handle, int total_count)
+        private static void ProverSearchCredentialsCallbackMethod(int xcommand_handle, int err, int search_handle, int total_count)
         {
             var taskCompletionSource = PendingCommands.Remove<CredentialSearch>(xcommand_handle);
```
mattraffel: I would like to see @srottem thoughts on changing these data types.  There is a slight difference between int and IntPtr that isn't obvious.

srottem:
What was the intermittent problem?  The difference between an IntPtr and an int is that int is a fixed 32 bit length while IntPtr can be larger or smaller depending on the architecture it's running on.  Since a Rust int should fit into an IntPtr I'm not sure why we would need to change the type.

tmarkovski:
The intermittent error I am seeing is WalletInvalidHandle = 200. I suspect this may be related to rust moving int handle values around on the stack. If this happens, the IntPtr the .net runtime references is no longer a valid handle, as the address value has changed. I looked at how the java wrapper handles this, and noticed it uses int value. I made this change and have been testing it internally for about a week before I made the PR. I didn't encounter the error I was seeing before. I am not entirely positive this was the cause of the problem, just that what I suspected and applied a fix, seemed to solve the problem.

srottem:
I'm not convinced that the change from IntPtr to int is why you're no longer experiencing the problem - IntPtr is usually used to hold pointers from C/C++ and maps to the intptr_t type and although Rust is probably not exposing the handles as this type I wouldn't expect it to have caused a problem.  Regardless, as long as the handles exposed by Rust can be stored reliably in a 32-bit integer I don't see a problem with making the change.

jovfer:
@tmarkovski @srottem 
> typedef int32_t       indy_handle_t;
it's a handle definition from libindy sources. So I see no reason to use `IntPtr` as it does *not* match to actual types FFI call. My vote is `int` as it's exactly int32 AFAIK
But I'm not expert in C# so I can miss something important about IntPtr while I try to google this nuances

tmarkovski:
Help me out here, my knowledge in Rust is very limited. This is the scenario I was describing.

C#: invokes a new thread and calls unmanaged libindy function
- Stack memory allocated for thread
Rust: functions creates an int32 handle with memory adress 0xFFFF, value 1
Rust: calls C callback function back to C# and passes value 1
Rust: calls another function to store value 1. At this point the value passed to the store function is copied on the stack at new memory address 0xFFFD (this is how I know rust handles function calls with value types)
C#: stores IntPtr with value 0xFFFF
- Actual stack memory lives as long as the thread is alive
- Thread ends, stack memory deallocated, C# keeps the IntPtr because there are strong references to it
At this point, Rust still knows the handle value of 1, but it's at address 0xFFDD, while C# still points to address 0xFFFF
When new thread allocates the memory at 0xFFFF at some point in time and writes new value, any attempt by C# to use the value at 0xFFFF as handle in Rust would throw "invalid handle".

Does this make sense as something that would theoretically be possible to occur?

jovfer:
@tmarkovski let me just simplify and point out the major nuance.
Rust return by callback some **value** as input parameter of the cb-function
Any wrapper must store this value, but not the temporary address.
I have not clear vision what particular will be stored in IntPtr, but if it will be address - it's misuse.
And there is another potential problem here - IntPtr size doesn't match to int32

PR #1288, opened 2018-11-13
Updated Rust wrapper to provide functions for logger setting

Artemkaaas:


wrappers/rust/src/logger.rs
```diff
@@ -0,0 +1,148 @@
+extern crate log;
+extern crate libc;
+
```
mattraffel: When this was an evernym project, our philosophy was all extern crate statements were only in the lib.rs file.   The reason was clarity:  it made it easy to update, correct references to external crates.   Do we want to continue this pattern now?

PR #1294, opened 2018-11-16
VE-787 VE-792 enable "public_did" in connection invitations

glowkey:
Also refactor how connections are built


jovfer:
@ryMarsh44 @hadleym Could you review please?

vcx/libvcx/src/lib.rs
```diff
@@ -76,7 +77,7 @@ mod tests {
     #[cfg(feature = "pool_tests")]
     #[test]
     fn test_real_proof() {
-        let number_of_attributes = 50;
+        let number_of_attributes = 10;
```
hadleym: Is reducing the number of attributes due to our agency limitations?

vcx/libvcx/src/messages/update_profile.rs
```diff
@@ -80,6 +80,17 @@ impl UpdateProfileData{
         }
     }
 
+    pub fn use_public_did(&mut self, did: &Option<String>) -> &mut Self {
```
hadleym: This would be more 'rust' like:
```
    pub fn use_public_did(&mut self, did: &Option<String>) -> &mut Self {
        if let Some(x) = did {
            self.payload.configs.push(AttrValue {
                name: "publicDid".to_string(),
                value: x.to_string(),
            });
        };
        self
    }
```

vcx/libvcx/src/lib.rs
glowkey: Yes.  It will make it more compatible.

PR #1301, opened 2018-11-19
 VE-787 VE-792 enable "public_did" in connection invitations

glowkey:


jovfer:
And probably there is some problem with python too https://ci.evernym.com/job/Indy%20SDK%20CI/view/change-requests/job/PR-1301/1/execution/node/1740/log/

PR #1307, opened 2018-11-21
RC 1.6.8

jovfer:


libindy/src/services/pool/state_proof/mod.rs
```diff
@@ -192,7 +192,9 @@ fn _parse_reply_for_builtin_sp(json_msg: &SJsonValue, type_: &str) -> Option<Vec
                                                           json_msg["ref"].as_u64()) {
                 trace!("TransactionHandler::parse_reply_for_builtin_sp: GET_CRED_DEF sign_type {:?}, sch_seq_no: {:?}", sign_type, sch_seq_no);
                 let marker = if ProtocolVersion::is_node_1_3() { '\x03' } else { '3' };
-                format!(":{}:{}:{}", marker, sign_type, sch_seq_no)
+                let tag = if ProtocolVersion::is_node_1_3() { None } else { json_msg["tag"].as_str() };
```
ashcherbakov: It would be great to have unit tests for every creation of the expected key

PR #1314, opened 2018-11-24
[IS-1094] refactor authenticated_encrypt/decrypt to service layer

kdenhartog:
refactored authenticated_encrypt and authenticated_decrypt from the command layer to the service layer, so that the functionality can be used in pack and unpack as well. Additionally changed some tests so they use asserts rather than unwraps to make it clear what is being tested.


libindy/src/services/crypto/mod.rs
```diff
@@ -296,6 +296,36 @@ impl CryptoService {
...
+
+        let msg_pack = msg.to_msg_pack()
+            .map_err(|e| CommonError::InvalidState(format!("Can't serialize ComboBox: {:?}", e)))?;
+
+        let res = self.encrypt_sealed(&their_vk, &msg_pack)?;
+
+        Ok(res)
+    }
+
+    pub fn authenticated_decrypt(&self, my_key: &Key, doc: &[u8]) -> Result<(String, Vec<u8>), CryptoError> {
```
vimmerru: @kdenhartog

Hi,

Thanks for contribution. I suggest to perform some changes to make our terminology and internal API better:

You added new `authenticated_encrypt/authenticated_decrypt` method, but they don't perform what is usually mean on this in cryptography. Look to https://en.wikipedia.org/wiki/Authenticated_encryption. It defines output of authcrypt as a pair (ciphertext, mac). 

In crypto service we already have a function that provides similar output it is `encrypt`. It can be better to rename existing `encrypt/encrypt_sealed` to `authcrypt/anoncrypt`, but on other hand we already have conception on combo-box and i suggest to start using original NaCl names `box/box_open` and `seal_box/seal_box_open`

What `authenticated_encrypt` is actually doing is creation of combo box and i suggest to rename it to `combo_box` and `authenticated_decrypt` tp `combo_box_open`. Existing method `create_combo_box` doesn't look valuable and we don't have corresponded `open` pair.

As a result I suggest the following:

1. Rename `authenticated_encrypt/authenticated_encrypt` to `combo_box/combo_box_open`
2. Rename `encrypt/decrypt/encrypt_sealed/decrypt_sealed` to `box/box_open/sealed_box/sealed_box_open`
3. Remove existing `create_combo_box` method and just inline it.

@jovfer what do you think?

vimmerru: @kdenhartog 

We discussed with Sergey that ```authenticated_encrypt``` is exposed as a public API. And it is better to keep it as is and remove number 1 from requested changes.

kdenhartog: All of these changes work except for the proposed "box" function name. This is a reserved keyword for rust it appears, so I've changed it to be "crypto_box" and "crypto_box_open". If we wanted to remain aligned with libsodium using this pattern, I'd suggest also changing "sealed_box" to "crypto_box_seal" and "sealed_box_open" to "crypto_box_seal_open". What are your thoughts on this additional change?

jovfer:
@vimmerru you have suggested to rename some function inside crypto service from encrypt/decrypt to box/box_open
We also have `CryptoType` crate, what do you think about it? Should methods in it be renamed too?
@kdenhartog FYI

cam-parra:
@jovfer Could we resolve these naming concerns soon? If it's not a major concern then we should merge

kdenhartog:
As an FYI, we're going to need to consume the AEAD functionality of sodium oxide. If we want this built into the crypto_type interface (I'm leaning towards not bringing it in), then I'd suggest we change the name of authenticated_encrypt/authenticated_decrypt as well.

jovfer:
@kdenhartog @vimmerru I suggest to merge this PR without renaming but with adding some TODOs. The first problem is broken CryptoType trait (generic purpose trait now uses specific data types). The second one - we have to extend it for new functionality or splint into 2 different traits. This changes I suggest to perform in #1210

@kdenhartog please create the ticket (or may be 2 separate) about problems with CryptoType and add TODO(s) in the codebase.

PR #1326, opened 2018-11-29
VE-787 refactor node, python wrappers to support full connection options

glowkey:
Signed-off-by: Douglas Wightman <douglas.wightman@evernym.com>


vcx/libvcx/tests/logging.rs
```diff
@@ -24,7 +26,7 @@ mod log_tests {
     use indy::future::Future;
 
     static mut COUNT: u32 = 0;
-    extern fn custom_log(_context: *const c_void,
```
Artemkaaas: I suggest reverting these changes. 
We already fixed indy-sys to use `libc` types.

PR #1332, opened 2018-12-03
IS-1101 and a small correction to VCX readme 

mattraffel:
Small correction to VCX readme


jovfer:
@glowkey @hadleym @ryMarsh44 from high-level point of view I agree with Artem. But we can miss some implementation specific for VCX, so I would like to have your opinion too
Do you have any objections or may be you would like to point out some drawbacks?

glowkey:
I believe {"key": "value"} is more user friendly and should be used by vcx_issuer_create_credential. I agree it is better for API and wrappers to use the same format.

mattraffel:
What does this break if we change this?   I haven't looked.  The change I made ensured the inputs didn't change.

glowkey:
While it would change the c-callable API it would not change the wrapper API so end-users won't see any changes.

mattraffel:
> While it would change the c-callable API it would not change the wrapper API so end-users won't see any changes.

This is a breaking change.  While the API signature doesn't change (aka it will compile), consumers will be broken because the format of the input (credential_data) will be broken.  That IS A BREAKING CHANGE.

jovfer:
@mattraffel As far as I know and understood from Douglas's message above, the `credential_data` for VCX is changed and yes, it's breaking change. But end-users consumes not C-callable API, but wrappers. And wrappers already have this new format of `credential_data`.
@glowkey please correct me, if I miss something.

vcx/libvcx/src/issuer_credential.rs
```diff
@@ -392,28 +416,41 @@ pub fn encode_attributes(attributes: &str) -> Result<String, IssuerCredError> {
...
+                let attrib_value = match i.as_str(){
+                    Some(v) => v,
+                    None => {
+                        warn!("Cannot encode attribute: {}", error::INVALID_ATTRIBUTES_STRUCTURE.message);
+                        return Err(IssuerCredError::CommonError(error::INVALID_ATTRIBUTES_STRUCTURE.code_num))
+                    },
+                };
+
+                warn!("Old attribute format detected. See vcx_issuer_create_credential api for additional information.");
+                attrib_value.to_string()
```
Artemkaaas: Here is redundant creating of String. You can leave &str in both cases.

vcx/libvcx/src/issuer_credential.rs
```diff
@@ -1041,4 +1079,224 @@ pub mod tests {
...
+            "city":["SLC"],
+            "address1":["101 Tela Lane"]
+            }"#;
+
+        let results_json = encode_attributes(TEST_CREDENTIAL_DATA).unwrap();
+
+        let results : Value = serde_json::from_str(&results_json).unwrap();
+        let address2 : &Value = &results["address2"];
+
+        assert_eq!("68086943237164982734333428280784300550565381723532936263016368251445461241953", address2["encoded"]);
```
Artemkaaas: Maybe better to use `encode` function to get expected values instead of hardcoded?

vcx/libvcx/src/issuer_credential.rs
```diff
@@ -1041,4 +1079,224 @@ pub mod tests {
...
+    #[test]
+    fn test_encode_bad_format_returns_error()
+    {
+        static BAD_TEST_CREDENTIAL_DATA: &str =
+            r#"{"format doesnt make sense"}"#;
+
+        match encode_attributes(BAD_TEST_CREDENTIAL_DATA) {
+            Ok(_) => assert!(false, "expected an error from encode_attributes"),
+            Err(_) => {}
+        }
```
Artemkaaas: maybe better:
`        assert_eq!(IssuerCredError::CommonError(INVALID_JSON.code_num), encode_attributes(BAD_TEST_CREDENTIAL_DATA).unwrap_err());
`
or just
`assert!(encode_attributes(BAD_TEST_CREDENTIAL_DATA).is_err())`
instad of match

vcx/wrappers/node/src/api/issuer-credential.ts
```diff
@@ -67,7 +67,7 @@ export class IssuerCredential extends VCXBaseWithState<IIssuerCredentialData> {
                          credentialName, price }: IIssuerCredentialCreateData): Promise<IssuerCredential> {
     try {
       const attrsVCX: IIssuerCredentialVCXAttributes = Object.keys(attr)
-      .reduce((accum, attrKey) => ({ ...accum, [attrKey]: [attr[attrKey]] }), {})
+      .reduce((accum, attrKey) => ({ ...accum, [attrKey]: attr[attrKey] }), {})
```
Artemkaaas: Why did you leave this mapping? It is absolutely useless.
`
> var attr = {key: "value"}
> const attrsVCX = Object.keys(attr).reduce((accum, attrKey) => ({ ...accum, [attrKey]: attr[attrKey] }), {})
> console.log(attrsVCX)
{ key: 'value' }
`

An interface `IIssuerCredentialVCXAttributes` should be changed also because it specifies the old format.

mattraffel: I needed someone to review to confirm that code wasn't needed any more.  correction checked in.

vcx/libvcx/src/issuer_credential.rs
```diff
@@ -392,28 +416,45 @@ pub fn encode_attributes(attributes: &str) -> Result<String, IssuerCredError> {
...
-            Some(x) => x,
-            None => {
+    let mut dictionary = HashMap::new();
+
+    for (attr, attr_data) in map.iter_mut(){
+        let first_attr : &str = match attr_data {
+            // old style input such as {"address2":["101 Wilson Lane"]}
+            serde_json::Value::Array(array_type) => {
+
+                let attrib_value : &str = match array_type.get(0).and_then(serde_json::Value::as_str) {
```
jovfer: @glowkey @mattraffel should we handle array[2] as specific case when encoding already done by application level? Or may be we should accept only array of single element.

mattraffel:
I will make a new PR

vcx/libvcx/src/issuer_credential.rs
jovfer: agree

PR #1320, opened 2018-11-27
Added static analysis support to CI

stevengubler:
We add our SA tool, clippy, as a component in our rust environment set up.

We also disabled errors from clippy for now until those breaking issues are addressed. In the meantime, this shouldn't break the CI  

Signed-off-by: Steven Gubler <steven.gubler@sovrin.org>

PR #1341, opened 2018-12-07
Update rust version to 1.29

stevengubler:
This is for to allow for clippy support in #1320


stevengubler:
https://github.com/hyperledger/indy-sdk/pull/1320#issuecomment-445086601

PR #1342, opened 2018-12-07
VCX: Allow to work with different wallet storage types

Artemkaaas:


glowkey:
I believe the vcx/libvcx/scripts/provision_agent_keys.py script should be updated to handle this new optional parameter.

jovfer:
@glowkey please check one more time: is it OK now?

PR #1356, opened 2018-12-13
Rust update 1.31

stevengubler:
Updating to 1.31 for #1320


stevengubler:
Crap, I branched off in the wrong spot...

PR #1378, opened 2018-12-21
IS-719 changes

mattraffel:


libindy/src/services/pool/request_handler.rs
```diff
@@ -285,7 +285,10 @@ impl<T: Networker> RequestSM<T> {
             }
             RequestState::Consensus(mut state) => {
                 match re {
-                    RequestEvent::Reply(_, raw_msg, node_alias, req_id) => {
+                    RequestEvent::Reply(_, raw_msg, node_alias, req_id) |
```
jovfer: In current codebase we have some lack of unit-tests coverage.
@mattraffel please fix it for updated functionality: add something like `request_handler_process_reqnack_event_from_consensus_state_works_for_consensus_reached` but for mixed events. E.g. check what for 4 nodes in pool 1 NACK + 1 REJECT hasn't finished request processing.

jovfer: Probably we should duplicate some test-cases for single and consensus types

mattraffel: Added a test case for mixed events.   I can add more if needed.  I do not understand the other request to "duplicate some test-cases".  Not sure what we should be duplicating and changing.  Can you clarify?

libindy/src/services/pool/request_handler.rs
```diff
@@ -945,6 +928,16 @@ pub mod tests {
             assert_match!(RequestState::Finish(_), request_handler.request_wrapper.unwrap().state);
         }
 
+        #[test]
+        fn request_handler_process_reply_event_from_consensus_state_works_for_consensus_reached_with_mixed_msgs() {
```
KitHat: I have some concerns about this test.
First we create a request handler with 1 node and `f = 0`. Then we send to it 3 responses -- it just doesn't make sense.
What I suggest to do is to write a test that would fail before and pass right now.
Example:
* Create a request handler with 8 nodes and `f = 2`
* Start with sending a `CustomConsensusEvent`
* Send one `Reply` and two `Reject`messages with different data in each `Reject`
* Check that state of request handler is still not finished
* Send one more `Reject` with data that equals to one of the previous `Reject` messages
* Check that state of request handler changed to `Finish`

I think that this one will be enough to verify the changes made for consensus case and reject. The same test should implemented for renack and both of them for single case.

mattraffel:
I am going to have to close this PR and start over because of the merge conflicts

PR #1397, opened 2019-01-11
[WIP] Indy get last error

Artemkaaas:


libindy/src/api/mod.rs
```diff
@@ -257,5 +258,31 @@ pub extern fn indy_set_runtime_config(config: *const c_char) -> ErrorCode {
 
     trace!("indy_set_runtime_config: <<< res: {:?}", res);
 
+    res
+}
+
+/// Get details for last occurred error
```
vimmerru: 1. I believe we need to provide more information on when this method should be called. 
2. Today we voted to avoid error_hanlde
3. I don't like the idea of returning error code here. It can cause recursion calls to resolve errors
4. We need to add information about returned pointer lifetime to description.
5. We need to add information on when we resolve back-trace

vimmerru: 6. As i remembered we decided to rename it to get_current_error.

libindy/src/utils/result.rs
```diff
@@ -1,46 +1,66 @@
...
-    ($result:ident, $default_value1:expr, $default_value2:expr) => {
+macro_rules! prepare_result_2 {
+    ($result:ident, $default_value1:expr, $default_value2:expr) => {{
+        trace!("prepare_result_2: >>> {:?}", $result);
         match $result {
             Ok((res1, res2)) => (ErrorCode::Success, res1, res2),
-            Err(err) => (err.into(), $default_value1, $default_value2)
-        };
-    }
+            Err(err) => {
```
vimmerru: TO Discuss: Should we clean previous error on success?

libindy/src/errors/mod.rs
```diff
@@ -2,15 +2,18 @@ use std::cell;
...
 use failure::{Backtrace, Context, Fail};
 use indy_crypto::errors::IndyCryptoError;
 use log;
 
 use api::ErrorCode;
+use utils::ctypes;
 
 pub mod prelude {
-    pub use super::{err_msg, IndyError, IndyErrorExt, IndyErrorKind, IndyResult, IndyResultExt};
+    pub use super::{err_msg, IndyError, IndyErrorExt, IndyErrorKind, IndyResult, IndyResultExt, LAST_ERROR, set_last_error};
```
vimmerru: Do we really need LAST_ERROR in prelude?

libindy/src/errors/mod.rs
```diff
@@ -360,3 +426,17 @@ impl<E> IndyErrorExt for E where E: Fail
     }
 }
 
+thread_local! {
+    pub static LAST_ERROR: RefCell<Option<CString>> = RefCell::new(None);
```
vimmerru: Does it need to be public?

Artemkaaas: we can get access to thread variable only inside `with` closure.
we can avoid making LAST_ERROR public by adding setter which accepts `*mut *const c_char` param (in my opinion this worse).

PR #1400, opened 2019-01-14
Wrappers get error details

Artemkaaas:


wrappers/python/indy/error.py
```diff
@@ -165,13 +166,17 @@ class ErrorCode(IntEnum):
...
     PaymentOperationNotSupportedError = 704,
-    
+
     # Extra funds on inputs
     PaymentExtraFundsError = 705
 
 
 class IndyError(Exception):
     # error_code: ErrorCode
+    # message: Optional[str]
```
vimmerru: Only backtrace should be optional from my point of view. Also it is good to add note how to get backtrace in comments.

vimmerru: And i suggest to rename backtrace to indy_backtrace

Artemkaaas: > Only backtrace should be optional from my point of view. Also it is good to add note how to get backtrace in comments.

I have put message and backtrace as optional fields to save backward compatibility (init IndyError with error code only).

libindy/src/errors/mod.rs
```diff
@@ -158,7 +158,14 @@ impl IndyError {
     pub fn extend<D>(self, msg: D) -> IndyError
         where D: fmt::Display + fmt::Debug + Send + Sync + 'static {
         let kind = self.kind();
-        self.to_indy(kind, msg)
+        let inner = Arc::try_unwrap(self.inner).unwrap();
+        IndyError { inner: Arc::new(inner.map(|_| msg).context(kind)) }
+    }
+
+    pub fn extend_with_type<D>(self, kind: IndyErrorKind, msg: D) -> IndyError
```
vimmerru: I suggest to rename it to “map”

PR #1402, opened 2019-01-14
fixes revocation for backwards compatibility with libvcx 0.1

glowkey:


vcx/libvcx/src/credential.rs
```diff
@@ -212,7 +215,8 @@ impl Credential {
         match self.state {
             VcxStateType::VcxStateOfferSent => {
                 //Check for messages
-                let _ = self._check_msg();
+                let result = self._check_msg();
+                println!("*******************result {:?}", result);
```
jovfer: Please replace println by logging (if possible)

PR #1371, opened 2018-12-18
Postgres cleanup 1113

ianco:
Added storage_credentials to CLI wallet delete

Code cleanup of postgres wallet:

- Removed unnecessary error codes
- Removed (duplicated) language.rs module

Other code is all required:

utils.crypto.base64 = encoding and decoding of tags, see:
     - https://github.com/ianco/indy-sdk/blob/postgres_cleanup_1113/samples/storage/storage-postgres/src/postgres_wallet.rs#L853
     - https://github.com/ianco/indy-sdk/blob/postgres_cleanup_1113/samples/storage/storage-postgres/src/wallet_storage/language.rs#L35

utils.crypto.chacha20poly1305_ietf = required for const ENCRYPTED_KEY_LEN, used by EncryptedValue, used to decode wallet storage parameters
Code organization is based on whether code is specific to postgres storage or copied from libindy:
     - https://github.com/ianco/indy-sdk/blob/postgres_cleanup_1113/samples/storage/storage-postgres/src/wallet_storage/storage.rs#L16

- postgres_storage module is specific to postgres
- api, errors, utils and wallet_storage are all "cut & paste" copied from libindy

PR #1404, opened 2019-01-15
[WIP] VCX add support for pluggable wallets (postgres) and a few questions

ianco:
Note please ignore changes under cli, samples, experimental, these are submitted under a separate PR https://github.com/hyperledger/indy-sdk/pull/1371 which is in progress being reviewed by @vimmerru .

This PR is W.I.P. - I have a few questions if someone can please help me:

1.  Is PR for pluggable wallets something you are interested in?  I have added support for create, open, close - delete and export/import are still outstanding, I can complete and submit if there is interest.

2.  Regarding long-running transactions - e.g. Faber sends Alice a PR, but won't get a response until Alice responds (potentially hours or even days) - I've assumed that the objects (Connection, Proof, etc.) can be serialized, stored (in the wallet?) and then retrieved and deserialized later, is this a correct assumption?  See updated Alice/Faber demo (e.g. https://github.com/ianco/indy-sdk/blob/vcx_prototype/vcx/wrappers/python3/demo/faber.py#L216).  Is this a correct assumption?  And if so is VCX going to support the save/restore of the serialized objects or does the application need to deal with that?

3.  I notice that VCX opens the wallet - this means the application can't have a previously opened wallet, because indy-sdk will throw a WalletAlreadyOpened error.  Is VCX going to deal with (proxy) all indy-sdk operations or should the VCX API take an (optional) wallet handle as a parameter?

4.  I notice VCX uses global variables to store current settings (https://github.com/ianco/indy-sdk/blob/vcx_prototype/vcx/libvcx/src/settings.rs#L238)?  Have you considered adding a context or context handle to VCX API to allow VCX to support multi-tenancy?  Currently it looks like it can only support a single wallet?

5.  In the Python demo, the Proof Response (Alice) always seg faults at the end, after sending the proof to Faber, is this something you also experience?


glowkey:
I've worked on VCX in the past so I'll try to answer these questions, but they aren't authoritative.

1) I think the long-term vision is for libvcx to stop managing pool and wallet handles internally but rather to accept the same pool and wallet handles created by libindy. In other words, you could use libindy to create and open whatever wallet you want (with whatever plugin you want) and then give that wallet handle to libvcx. This way libindy and libvxc would work together (within the same process) and use the same wallet without any WalletAlreadyOpened errors.

2) All libvcx objects can be serialized and saved by the application. They can then be deserialized as needed. These serialized objects are not stored in the wallet. It is up to the application to save these somewhere (I guess you could store them in the wallet using the non-secrets API). In other words, libvcx does not manage the life of the object. It is up to the application to create, serialize, destroy, deserialize as needed.

3) See answer for #1. 

4) Yes. There has been a request to add a context handle to each call so that multiple wallets/pools/etc can be handled within the same process. Some of this would be solved by #1, but not everything.

5) I've seen this randomly depending on python and system versions. 

I hope these answers were helpful.

ianco:
@glowkey  thanks for the info!

ianco:
Closing for now, will review and open a separate PR

PR #1405, opened 2019-01-15
update Indy link and DCO

vikpande:
update Indy link and add DCO sign off


vikpande:
pls ignore since the DCO didnt work-out

PR #1408, opened 2019-01-15
Nodejs wrapper - indy_get_current_error

farskipper:
For both async and sync errors, it calls `indy_get_current_error` and attaches the additional data to the IndyError object.


jovfer:
Will be merged as part of 1400

PR #1421, opened 2019-01-18
Vcx Release Throws Errors

hadleym:
Previously, the release function for each Libvcx object always returned a success, even when the handle was invalid or previously released.

Errors are now passed back from the release api call.

Node wrapper tests for garbage collection are now disabled as they will no longer behave predictably.


hadleym:
@glowkey Please review.

glowkey:
I've suggested a change offline specific to node wrapper

vcx/wrappers/node/src/utils/memory-management-helpers.ts
```diff
@@ -1,21 +1,14 @@
...
-    try {
-      const rc = this._releaseFn(this._handleRef)
-      if (rc) {
-        throw rc
-      }
-      return rc
-    } catch (err) {
-      throw new VCXInternalError(err)
-    }
+  public async release (): Promise<void> {
```
hadleym: Each object extends GCWatcher, so this is where the release function is defined.  It no longer throws an error.

vcx/wrappers/node/src/utils/memory-management-helpers.ts
```diff
@@ -31,14 +24,7 @@ export abstract class GCWatcher {
...
     weak.addCallback(weakRef, () => {
-      try {
-        const rc = release(handle)
-        if (rc) {
-          throw rc
-        }
-      } catch (err) {
-        throw new VCXInternalError(err)
-      }
+      release(handle)
```
hadleym: This try/catch is no longer necessary due to the changes in the code above.

hadleym:
Tests within the Node Wrapper that test if release() throws an error have been removed.

vcx/wrappers/node/src/utils/memory-management-helpers.ts
glowkey: Can you write a comment saying "why" we're not throwing an error? Unpredictable node GC for example.

PR #1428, opened 2019-01-21
Safely zero keys and prevent compiler optimization of zeroing

mikelodder7:
Signed-off-by: Michael Lodder <redmike7@gmail.com>

### DIFF
--- a/libindy/src/domain/crypto/key.rs
+++ b/libindy/src/domain/crypto/key.rs
@@ -1,4 +1,7 @@
 extern crate indy_crypto;
+extern crate zeroize;
+
+use self::zeroize::Zeroize;
 
 use named_type::NamedType;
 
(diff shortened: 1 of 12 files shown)


jovfer:
@mikelodder7 thanks for the PR, I hope we include it in nearest release.

@Artemkaaas please review

libindy/src/domain/crypto/key.rs
```diff
@@ -24,6 +27,20 @@ impl Key {
     }
 }
 
+//Can't figure out why this macro can't be found in `cargo test` because
+//if works for src/utils/crypto/randombytes/sodium
+//memzeroize!(Key, signkey);
+
```
Artemkaaas: It fails because of integration tests.
There is an import of some libindy modules by path here (tests/utils.mod.rs) including domain folder.
so, you need to import the file containing `memzeroize` macro there also.
because mod.rs contains submodules I suggest moving memzeroize macro to separate file.
steps to achieve memzeroize macro works here:
1) move memzeroize macro into a separate file (maybe src/utils/memzeroize.rs)
2) add 
 `#[macro_use]
pub mod zeroize;` into `src/utils/mod.rs`
3) add
 `#[macro_use]
#[allow(unused_macros)]
#[path = "../../src/utils/zeroize.rs"]
pub mod zeroize;` into `tests/utils/mod.rs`

libindy/tests/utils/mod.rs
```diff
@@ -1,7 +1,11 @@
-#![allow(dead_code)]
+#![allow(dead_code, unused_macros)]
```
jovfer: @mikelodder7 it's better to use `allow(some_warning)` for minimal context, like near module import and without `!` (if possible). But we can do it later and do not block the PR

PR #1430, opened 2019-01-22
Fix vcx java wrapper

seniorjoinu:
Fixed some native function signatures
Fixed some native function calls
Fixed some typos

We are now able to rewrite vcx agent demo (from python wrapper folder) in java.


jovfer:
@glowkey @hadleym could you review please?

PR #1434, opened 2019-01-24
[VCX] Filter payment addresses based on configured payment method

glowkey:


vcx/libvcx/src/utils/libindy/payments.rs
```diff
@@ -169,26 +169,39 @@ pub fn list_addresses() -> Result<Vec<String>, u32> {
     Ok(serde_json::from_str(&addresses).or(Err(error::INVALID_JSON.code_num))?)
 }
 
+fn is_valid_address(address: &str, method: &str) -> bool {
+    let prefix = format!("pay:{}", method);
+
+    if address.starts_with(&prefix) {
```
jovfer: I suggest to just return `address.starts_with(&prefix)` as result

vcx/libvcx/src/utils/libindy/payments.rs
```diff
@@ -169,26 +169,39 @@ pub fn list_addresses() -> Result<Vec<String>, u32> {
...
+        if is_valid_address(&address, &method) {
+            debug!("getting address info for {}", address);
+            let mut info = get_address_info(&address)?;
 
-        for utxo in info.utxo.iter() { balance += utxo.amount as u64; }
+            for utxo in info.utxo.iter() { balance += utxo.amount as u64; }
 
-        wallet_info.push(info);
+            wallet_info.push(info);
+        }
```
jovfer: @glowkey @hadleym I think it's useful to add something as `else` branch.
1) Should it be an error to throw from function as result?
2) If no, please add `warn` or `info` or `debug` log. I think it's should be debug+ level (not trace), but don't have enough context to suggest exact level...

glowkey: Good suggestion. I have added a warning log statement.

PR #1437, opened 2019-01-26
[dotnet wrapper] Add pack and unpack message wrappers. Unit tests. Version bump.

tmarkovski:

### DIFF
--- a/wrappers/dotnet/indy-sdk-dotnet-test/CryptoTests/PackUnpackMessageTest.cs
+++ b/wrappers/dotnet/indy-sdk-dotnet-test/CryptoTests/PackUnpackMessageTest.cs
@@ -1,0 +1,140 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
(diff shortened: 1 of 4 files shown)


kdenhartog:
Is the team using the pack/unpack stuff for streetcred alright with the message being of type byte[] instead of String? In the python wrapper, I made message type String, and handle the encoding within the wrapper to make it a little easier to use.

tmarkovski:
I thought about this as well, it makes sense since the byte array will always be UTF8 encoded string. Then I saw the PR for the java wrapper and made it consistent. I think the comments on the wrapper API's are clear enough to indicate what the output is.
For the agent framework, we wrote [few extensions](https://github.com/streetcred-id/agent-framework/blob/60f610d36cfa86eaa8e9b57792c299bcba6f7730/src/AgentFramework.Core/Extensions/FormattingExtensions.cs#L15) to work easier with the conversions and be able to do things like `packedArray.GetUTF8String()`.

PR #1438, opened 2019-01-27
Python wrapper updates for pack/unpack APIs

kdenhartog:


wrappers/python/tests/crypto/test_pack_message.py
```diff
@@ -0,0 +1,56 @@
...
+    assert json_message['ciphertext'] != ""
+    assert json_message['iv'] != ""
+
+@pytest.mark.asyncio
+async def test_pack_message_invalid_verkey(wallet_handle, verkey_my2):
+    sender_verkey = "INVALID"
+    receiver_verkeys = [verkey_my2]
+    with pytest.raises(IndyError) as e:
+        pack_message = await crypto.pack_message(wallet_handle, "pack_message", receiver_verkeys, "Invalid_VERKEY")
+    assert ErrorCode.WalletItemNotFound == e.value.error_code
```
Artemkaaas: please check `test_pack_message_invalid_verkey`, `test_pack_message_invalid_receiver_verkey` and `test_pack_message_invalid_wallet_handle` tests
they are looks strange

wrappers/python/tests/crypto/test_unpack_message.py
```diff
@@ -0,0 +1,54 @@
...
+from indy import IndyError
+from indy import crypto, did, wallet
+
+from indy.error import ErrorCode
+
+@pytest.mark.asyncio
+async def test_pack_message_and_unpack_message_authcrypt_works(wallet_handle, seed_steward1, seed_my1, pack_message):
+    #setup keys
+    sender_vk = await did.create_key(wallet_handle, json.dumps({'seed': seed_my1}))
+    steward_vk = await did.create_key(wallet_handle, json.dumps({'seed': seed_steward1}))
```
Artemkaaas: use `identity_my` and `identity_steward1` instead

wrappers/python/indy/crypto.py
```diff
@@ -391,80 +391,113 @@ def transform_cb(arr_ptr: POINTER(c_uint8), arr_len: c_uint32):
...
-                                       unpack_message.cb)
+    c_jwe_len = c_uint32(len(jwe))
+    unpack_message_bytes = await do_call('indy_unpack_message',
+                                         c_wallet_handle,
+                                         jwe,
+                                         c_jwe_len,
+                                         unpack_message.cb)
+
+    try:
+        unpack_message_str = unpack_message_bytes.decode('utf-8', 'strict')
```
Artemkaaas: why do you do `decode('utf-8', 'strict')` on result of `unpack_message` but don't on result of `pack_message`?

wrappers/python/indy/crypto.py
```diff
@@ -391,80 +391,113 @@ def transform_cb(arr_ptr: POINTER(c_uint8), arr_len: c_uint32):
...
+    except UnicodeDecodeError:
+        logger.debug("unpack_message: wrapper decode error <<< res: %r", unpack_message_bytes)
+        return unpack_message_bytes
```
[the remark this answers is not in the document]
kdenhartog: I agree with this conclusion. I'll return the error rather than return a different type.

wrappers/python/tests/crypto/test_pack_message.py
```diff
@@ -0,0 +1,56 @@
...
+from indy.error import ErrorCode
+
+@pytest.mark.asyncio
+async def test_pack_message_authcrypt_works(wallet_handle, seed_my1, verkey_my2, pack_message):
+    sender_verkey = await did.create_key(wallet_handle, json.dumps({'seed': seed_my1}))
+    receiver_verkeys = [verkey_my2]
+    packed_message_bytes = await crypto.pack_message(wallet_handle, pack_message, receiver_verkeys, sender_verkey)
+    packed_message_json = packed_message_bytes.decode("utf-8")
+    json_message = json.loads(packed_message_json)
+    assert json_message['protected'] != ""
```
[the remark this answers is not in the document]
Artemkaaas: `assert json_message['protected']` and `assert json_message['protected'] != ""`

wrappers/python/tests/crypto/test_pack_message.py
kdenhartog: I've updated this to have correct parameters except the part that should cause the error. This will make it easier to understand the test. Is that what looked strange or was it something else?

wrappers/python/indy/crypto.py
kdenhartog: pack_message does not assume the type of encoding that will be used by the consumer because it doesn't assume the transport layer used. For example, in an HTTP Request, a JSON object can be put directly into the body of the request. However, if the transport layer is a URL then it should be base64URL encoded. However, in unpack_message it can be assumed that the API consumer will be processing the JSON object returned. In this case, we are assuming that the language will process JSON based on a specific type (string) in which case we can handle the encoding.

For consistency sake, I believe it should be the same across all APIs, so we should update the rust, .NET (PR #1437) and Java wrapper. Additionally, my first pass on the nodejs wrapper will need some help too (PR incoming).

wrappers/python/indy/crypto.py
```diff
@@ -391,80 +391,112 @@ def transform_cb(arr_ptr: POINTER(c_uint8), arr_len: c_uint32):
...
 
-    message, sender_vk = await do_call('indy_unpack_messasge',
-                                       c_wallet_handle,
-                                       c_ames_json,
-                                       c_my_vk,
-                                       unpack_message.cb)
+    try:
+        unpack_message_str = unpack_message_bytes.decode('utf-8', 'strict')
+    except UnicodeDecodeError:
+        return UnicodeDecodeError
```
Artemkaaas: just leave  `unpack_message_bytes.decode('utf-8', 'strict')`
it doesn't make sense
`
    except UnicodeDecodeError:
        return UnicodeDecodeError
`

wrappers/python/indy/crypto.py
```diff
@@ -391,80 +391,112 @@ def transform_cb(arr_ptr: POINTER(c_uint8), arr_len: c_uint32):
...
-    Experimental. Deserializes a AMES json string and decrypts the message returning the message and the sender's verkey if it was an AuthAMES.
-    If it is an AnonAMES it will deserialize, decrypt, and return the message with an empty sender_vk string.
+    Packs a message by encrypting the message and serializes it in a JWE-like format (Experimental)
 
-    :param wallet_handle: wallet handler (created by open_wallet).
-    :param ames_json: a json string serialized using either AuthAMES or AnonAMES
-    :param my_vk: the verkey to authcrypt with
-    :return: message: the unencrypted message
-             sender_vk: the sender's verkey if AuthAMES, else an empty string
+    Note to use DID keys with this function you can call indy_key_for_did to get key id (verkey)
```
Artemkaaas: use `did.key_for_did` instead

wrappers/python/tests/crypto/test_pack_message.py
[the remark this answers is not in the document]
kdenhartog: ahh gotcha, thanks

wrappers/python/indy/crypto.py
jovfer: I strongly suggest to keep the same type for input parameter `msg` of pack and result of unpack.

and for current implementation I suggest to use bytes for pack/unpack as its current approach across all wrappers

PR #1442, opened 2019-01-28
 [VCX] enable generic connection messages, signing and verification

glowkey:


jovfer:
@Artemkaaas please review

glowkey:
I'm sorry, I don't understand. I had to change it from an enum to strings because the message type can be anything, not just one of the enum values.

vcx/libvcx/src/utils/logger.rs
```diff
@@ -140,7 +140,7 @@ impl LibvcxDefaultLogger {
 
         // ensures that the test that is calling this wont fail simply because
         // the user did not set the RUST_LOG env var.
-        let pattern = Some(env::var("RUST_LOG").unwrap_or("debug".to_string()));
+        let pattern = Some(env::var("RUST_LOG").unwrap_or("trace".to_string()));
```
hadleym: Trace default?

vcx/libvcx/src/credential_def.rs
```diff
@@ -18,7 +18,7 @@ pub struct CredentialDef {
     tag: String,
     name: String,
     source_id: String,
-    issuer_did: String,
+    issuer_did: Option<String>,
```
hadleym: I'm confused on how this is an optional parameter.  I guess its no different than when we had it originally be an empty string upon creation.  Perhaps in a future iteration we require it as a parameter to create?

glowkey: Yes, this is optional. It is not required in the code and must be optional in order to not break compatibility.

vcx/libvcx/src/utils/logger.rs
glowkey: Yes, most of the vcx statements are trace so that should be the default.

PR #1452, opened 2019-02-01
Android libnullpay test

felippeb:


felippeb:
The artifacts produced for libnullpay on both android and ios are not functional when consumed by our application team. Testing is needed on libnullpay to ensure that the artifact being build by cd is infact functional, as right now that is not the case.

nsivraj:
We do not need extensive testing of libnullpay.a or libnullpay.so BUT we do need to know that the artifact will at least load into the platform before it is uploaded onto the repository website -- https://repo.sovrin.org/ios/libnullpay/. Currently, there are artifacts loaded out on that repository website that cannot even be loaded into the platform.

jovfer:
@nsivraj agree. Ok, let's try to support minimal test procedure like build and try to call init function on target platform.
@felippeb Please take a look on CD pipeline, may be you can re-use something.
Also please clean-up the PR and keep only android-related scope.

libnullpay/ci/ubuntu.dockerfile
```diff
@@ -0,0 +1,68 @@
+FROM ubuntu:16.04
```
jovfer: I suggest to use same file for libindy and libnullpay

libnullpay/ci/setup.android.env.sh
```diff
@@ -0,0 +1,225 @@
+#!/usr/bin/env bash
```
jovfer: Is it possible to re-use similar file from libindy subdir?

Artemkaaas:
Done in the scope of PR: https://github.com/hyperledger/indy-sdk/pull/1729

PR #1460, opened 2019-02-06
Hotfix for nullpay logging on mobile platforms

mattraffel:
This PR is to get an artifact to the vcx team with logging disabled in libnullpay

### DIFF
--- a/libnullpay/src/lib.rs
+++ b/libnullpay/src/lib.rs
@@ -23,7 +23,6 @@ use std::ffi::CString;
 
 #[no_mangle]
 pub extern fn nullpay_init() -> ErrorCode {
-
     if let Err(err) = utils::logger::LibnullpayLogger::init() {
         return err;
     }
--- a/libnullpay/src/lib.rs
+++ b/libnullpay/src/lib.rs
@@ -23,9 +23,10 @@ use std::ffi::CString;
 
 #[no_mangle]
 pub extern fn nullpay_init() -> ErrorCode {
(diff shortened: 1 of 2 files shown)


libnullpay/src/lib.rs
```diff
@@ -23,9 +23,10 @@ use std::ffi::CString;
 
 #[no_mangle]
 pub extern fn nullpay_init() -> ErrorCode {
-    if let Err(err) = utils::logger::LibnullpayLogger::init() {
```
jovfer: I suggest to use platform specific directive instead of commenting code

libnullpay/src/utils/logger.rs
```diff
@@ -1,3 +1,6 @@
+#![allow(dead_code)]
```
jovfer: Please add reference to ticket in Jira devoted to removing this hotfix.
If possible and only few functions are unused, please move global directive to particular functions.

libnullpay/src/lib.rs
jovfer: Also please add comment and reference to ticket/bug in Jira about resolving mobile platform problem

mattraffel: I added directives.

libnullpay/src/utils/logger.rs
jovfer: After new changes below it's seems like these 2 allows are not required now, isn't it?

mattraffel: agreed.

PR #1444, opened 2019-01-30
Scripts to build libindy for Ubuntu using Docker

xaviernopre:
Signed-off-by: Xavier NOPRE <xnopre@gmail.com>

PR #1465, opened 2019-02-07
Docker to run vcx demo

xaviernopre:
Dockerization to run VCX demo with Docker
Ideally, to be merged after https://github.com/hyperledger/indy-sdk/pull/1444


xaviernopre:
@jovfer Same question as for #1444 ;-)

PR #1479, opened 2019-02-15
Make python VCX wrapper demo work out of the box on OSX.

Patrik-Stas:

### DIFF
--- a/vcx/wrappers/python3/demo/alice.py
+++ b/vcx/wrappers/python3/demo/alice.py
@@ -2,9 +2,11 @@ import asyncio
 import json
 from ctypes import cdll
 from time import sleep
+import platform
 
 import logging
 
+from demo_utils import file_ext
 from vcx.api.connection import Connection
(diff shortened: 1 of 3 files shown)


vcx/wrappers/python3/demo/alice.py
```diff
@@ -27,7 +28,7 @@
 
 async def main():
 
-    payment_plugin = cdll.LoadLibrary("libnullpay.so")
+    payment_plugin = cdll.LoadLibrary("libnullpay.dylib" if platform.system() == 'Darwin' else "libnullpay.so")
```
Artemkaaas: suggest adding const dictionary like:
`extention = {"darwin": ".dylib", "linux": ".so", "win32": ".dll"}`

Patrik-Stas: Thanks for review, I've just noticed there's already `file_ext()` to determine the postfix in demo_utils, so I've refactored it a bit and reused.

PR #1481, opened 2019-02-16
return value in completable future is integer vs String

nehalshah50:
I tested this on my local machine. The bug here produces JVM crash with following error.

13:24:49.845 [Thread-4938] DEBUG com.evernym.sdk.vcx.LibVcx.native.vcx.issuer_credential - src/issuer_credential.rs:236 | issued credential: alice_degree

#
# A fatal error has been detected by the Java Runtime Environment:
#
# SIGSEGV (0xb) at pc=0x00007fff6dc5d712, pid=55775, tid=0x000000000000a003
#
# JRE version: Java(TM) SE Runtime Environment (8.0_152-b16) (build 1.8.0_152-b16)
# Java VM: Java HotSpot(TM) 64-Bit Server VM (25.152-b16 mixed mode bsd-amd64 compressed oops)
# Problematic frame:
# C [libsystem_platform.dylib+0x1712] _platform_strlen+0x12
#
# Failed to write core dump. Core dumps have been disabled. To enable core dumping, try "ulimit -c unlimited" before starting Java again
#
# If you would like to submit a bug report, please visit:
# http://bugreport.java.com/bugreport/crash.jsp
# The crash happened outside the Java Virtual Machine in native code.
# See problematic frame for where to report the bug.
#


jovfer:
@nehalshah50 could you fix DCO check?
https://github.com/hyperledger/indy-sdk/pull/1481/checks?check_run_id=291635336

Artemkaaas:
`vcx_issuer_send_credential` doesn't return any value in the callback. The proper fix will be returning of `Void`.
We will fix it in a new PR.

PR #1487, opened 2019-02-18
w3c spec compatible DID/DID Doc API design proposal

vimmerru:


docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+1. create a new DID and 1 or multiple related keys
+1. publish corresonded DID Doc to Public Ledger if i have sufficient rights
+1. provide DID Doc to 3d party (Trust Ancher) for publishing to Public Ledger if i have no sufficient rights
+1. publish new DID Doc to microledger
+1. add second key to the DID DOC from a different edge device
+1. rotate a key from an edge device
+1. change authentication/authorization policies from one of the devices
+1. change service endpoints from an edge device
+1. support of multiple endpoints, each working with multiple edge devices
+1. know the recent key for this edge device to be used for signature and A2A communication using this edge device
```
lovesh: Not only "for this edge device" but for any edge device of this relationship.

docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+
+### Create new DID and DID DOC
+
+![Alt](./create-did-ddoc.svg "Create new DID and DID DOC")
+
+### Create key on 2nd Edge Device
+
+![Alt](./create-key-2nd-device.svg "Create key on 2nd Edge Device")
+
+### Rotate a key by an Edge Device who is authenticated to do the update
```
lovesh: Replace `who is authenticated` with `who is authorized`

docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+
+![Alt](./rotate-key-with-auth.svg "Rotate a key by an Edge Device who is authenticated to do the updatee")
+
+### Rotate a key by an Edge Device who is not authenticated to do the update
```
lovesh: I think you mean `not authorized` by `not authenticated`. How can a device not authorized to update still update? Do you mean updating itself? Each device should be able to update itslef without requiring any other device.

docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+
+## Design principles
+
+1. Each edge device stores the following information for every owned DID:
+    - Fully-qualified Did - Identifies data in storages
+    - Private and public keys associated with the DID from this edge device - in Wallet
+2. Each edge device can optionally store cached local copy of DID Doc for owned DID:
+    - The current DID Doc copy is used for signing/pack.
+    - The current DID Doc copy is synchronized between devices only through the Ledger (Public or microledger).
+    - DID Doc is never sent from one device to another.
```
lovesh: In case of microledger?

docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+    - DID is the only thing which is passed
+    - The receiver resolves the DID looking at the did method and going to the corresponding ledger (public or micro)
+    - This will allow communication between old agents (not supporting DID Doc) and new ones using the new approach
+4. The Ledger level is opaque. It doesn’t matter if this is public or micro ledger. Exact publishing and resolving strategy are
+   defined by a pluggable Resolvers and doesn’t affect workflows.
+5. To support DID Doc new transactions will be added to Indy Ledger:
+    - DID_DOC
+      - Consists of `did` and `doc`
+      - The ‘doc’ fields is a map of the form {field_id -> field} where field_id is a unique identifier of a field in the doc, and field is one of the fields (either array or a single-value)
+      - The following fields will be there (see [sovrin-did-method](https://sovrin-foundation.github.io/sovrin/spec/did-method-spec-template.html#sovrin-did-method):
```
lovesh: Missing ending parenthesis

docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+    - DID Doc is never passed from one Agent (Alice) to another (Bob)
+    - DID is the only thing which is passed
+    - The receiver resolves the DID looking at the did method and going to the corresponding ledger (public or micro)
+    - This will allow communication between old agents (not supporting DID Doc) and new ones using the new approach
+4. The Ledger level is opaque. It doesn’t matter if this is public or micro ledger. Exact publishing and resolving strategy are
+   defined by a pluggable Resolvers and doesn’t affect workflows.
+5. To support DID Doc new transactions will be added to Indy Ledger:
+    - DID_DOC
+      - Consists of `did` and `doc`
+      - The ‘doc’ fields is a map of the form {field_id -> field} where field_id is a unique identifier of a field in the doc, and field is one of the fields (either array or a single-value)
```
lovesh: So no nesting?

docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+```Rust
+// Create a new DID.
+//
+// Note: It doesn't create any records in the wallet.
+//
+// did_resolver_handle: resolver handle
+// method: did method to use
+// options: some options
+// returns - did: Generated DID
+pub fn indy_create_did(did_resolver_handle,
```
lovesh: When is this method needed alone? Can it have functionlity of `indy_create_did_key`

docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+                                   options) -> Future<ddoc_builder_handle>
+```
+
+#### Add new public key item to DID Doc builder
+
+```Rust
+// Add new public key item to DID Doc builder.
+//
+// Note: id can be full DID Doc reference or fragment.
+// In case of fragment reference will be related to current DID.
```
lovesh: How is the current DID tracked? Is there a way to query such state management?

docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+                                   key_json) -> Future<()>
+```
+
+#### Update public key item in DID Doc builder
+
+```Rust
+// Update public key item in DID Doc builder.
+//
+// Note: id can be full DID Doc reference or fragment.
+// In case of fragment reference will be related to current DID.
```
lovesh: Same question as above about current DID

docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+
+#### Delete public key item in DID Doc builder
+
+```Rust
+// Delete public key item in DID Doc builder.
+//
+// Note: id can be full DID Doc reference or fragment.
+// In case of fragment reference will be related to current DID.
+//
+// Note: If key is referenced inside of DID Doc validation error will
```
lovesh: What is meant by "referenced inside of DID Doc"?

docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+// returns - key_json: json that describes the key. Can be passed to did
+//                     doc builder
+pub fn indy_create_did_key(did_resolver_handle,
+                           did,
+                           options) -> Future<key_json>
+```
+
+### DID Doc creation, updating and publishing
+
+#### Create a new DID Doc builder instance
```
lovesh: Do methods `indy_did_doc_*_public_key`, `indy_did_doc_*_authorization` and `indy_did_doc_*_service_endpoint` make changes to the DID Doc in wallet? Or do they maintain in-memory delta of changes after each such call?

docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+```Rust
+// Finalize DID Doc builder and return corresonded DID Doc or DID Doc delta.
+//
+// Note: Format of delta can be method specific.
+// In case of path reference will be related to current DID.
+//
+// ddoc_builder_handle: builder handle
+// id: identifier of key
+// returns - did_doc_delta: DID Doc or DID Doc delta
+pub fn indy_did_doc_finalize(ddoc_builder_handle) -> Future<did_doc_delta>
```
lovesh: Does this persist the doc? Can this be called repeatedly?

docs/design/014-did-doc/README.md
```diff
@@ -0,0 +1,590 @@
...
+await did.did_doc_add_public_key(ddoc_builder_hanle,
+                                 'key-1',
+                                 key_info)
+
+# 19. Use this key for authorization
+await did.did_doc_add_authorization(ddoc_builder_hanle,
+                                    'key-1')
+
+# 20. Finalize builder and get corresponded DID Doc delta
+let did_doc_delta = await did.did_doc_finalize(ddoc_builder_hanle)
```
lovesh: Typo in `ddoc_builder_hanle` here and above in a few places

docs/design/014-did-doc/README.md
vimmerru: We don't store DID Doc in wallet at all. Wallet only stores keys related to DID.

docs/design/014-did-doc/README.md
vimmerru: @lovesh @dhh1128  @ashcherbakov  I am not sure that it is necessary and feasible to get complete Edge devices topology and key/device mapping information from DID Doc. Some applications can provide this information, but it is application level, not libindy level.  What edge device know:

- The list of keys related to this DID stored in wallet
- Device can resolve the latest version of DID Doc and based on this information understand what local keys can be used to perform DID Doc modifications and what keys can be used for A2A communication

But It can be impossible to know on what devices other keys stored without getting some additional information from application user.

docs/design/014-did-doc/README.md
vimmerru: @lovesh @dhh1128 @ashcherbakov 

1. I think you mean not authorized by not authenticated. 

Yes, it is mistake. Will fix.

2. How can a device not authorized to update still update? Do you mean updating itself? Each device should be able to update itslef without requiring any other device.

Let's consider the following use case:

1. Alice has iPhone. In wallet we she has key pair for some relationship DID that is authorized to perform DID updates (like adding new keys).
2. Alice wants to use also iPad to participate in this relationship.
3. The question is how we can achieve this? You can notice that we need to add key form new edge device (iPad) that isn't authorized to perform updates.

We suggest the following:

1. Alice generates new key pair for this DID on iPad
2. iPad application displays QR code (or doing something similar) with public key info
3. iPhone app scans QR code
4. iPhone app performs DID Doc update by adding public key from iPad
5. iPad now can be used for this relationship communications and DID Doc updates if iPhone app provided enough authorization rights for new key.

Similar use keys is creation of public DID when you need to contact Trust Anchor. It can be solved similar way.

Suggested API allows to perform described operations.

docs/design/014-did-doc/README.md
vimmerru: @lovesh @dhh1128 @ashcherbakov

> In case of microledger?

It isn't important how DID Doc is shared. There is some storage (public ledger, micro ledger, centralized storage, some ledger between cloud agents) that all connected devices have access and this storage is used to perform DID Doc synchronization. Otherwise you can easy get DID Doc out of think that can cause security and communication problems.

docs/design/014-did-doc/README.md
ashcherbakov: I agree with Slava that Alice's edge device doesn't need to know all keys for all other Alice's edge device. A good example is a static agent (an IoT device) that need to know only the key it owns.

docs/design/014-did-doc/README.md
ashcherbakov: Yes, we propose to separate at least the following layers/protocols:
1) Sub-protocol of how DID DOC created and updated by agents
2) Sub-protocol of how DID DOC updates are ordered and synchronized between Alice's devices .
Options: public ledger, BFT-like microledger, CFT-like microledger, Centralized approach, etc.
3) Sub-protocol of how and when Bob get the latest Alice's DID DOC.
We will update the docs with this thinking.

docs/design/014-did-doc/README.md
ashcherbakov: I think `field_id` can be a combined ID of multiple nested ids

docs/design/014-did-doc/README.md
ashcherbakov: I think it makes sense to have this method as a separate one because of the following:
1) DID can be generated independent of the key. It can be just a UUID, not necessary a part of a public key
2) It makes the API more clear and clean from my opinion, especially if we want to create multiple keys for a DID in one builder.
3) I think there can be use cases when we want just DID, and keys will be generated latter.

docs/design/014-did-doc/README.md
ashcherbakov: I didn't get the question. The DID is one of the arguments for DID DOC Builder.

docs/design/014-did-doc/README.md
ashcherbakov: This just creates a Delta. No changes to a local copy of DID DOC are made.
How exactly the local DID DOC copy (if any) is changed is defined by a different layer (see the comment with 3 layers).

docs/design/014-did-doc/README.md
vimmerru: > What is meant by "referenced inside of DID Doc"?

It means used in some DID Doc sections. For example, in authentication section.

docs/design/014-did-doc/README.md
kdenhartog: I like the idea of keeping the pattern similar to Trust Anchor establishment at the libindy API layer. Let's leave how the verkey is transported up to consumers of the API similar to what we do for trust anchor pattern.

docs/design/014-did-doc/README.md
kdenhartog: I think the impacts of the storage solution that is being discussed will play into this discussion. With regards to microledger sharing, I think within a domain (all of alice's agents) we can rely on  a storage sync layer to handle it. With updates across domains (e.g. bob publishes update to alice) we would likely use DIDComms to communicate the update, and alice would store it locally in her storage sync layer.

Worth noting storage syncing is still very early thinking, so this discussion will be an important use case when we're looking to design it.

dhh1128:
Please note that this design needs to be harmonized with the peer DID spec (https://github.com/openssi/peer-did-method-spec) and with the connection management protocol (https://github.com/hyperledger/indy-hipe/pull/104) before it is merged.

PR #1488, opened 2019-02-18
Vcx java wrapper demo

nehalshah50:
This is VCX java wrapper demo similar to Python Demo. I've tested this on MacOS and is in working condition. Instructions are in README.md


vcx/wrappers/java/demo/buildlibs.sh
```diff
@@ -0,0 +1,39 @@
...
+cd $current_dir
+cd ..
+./gradlew build
+
+cd $current_dir
+
+
+
+if [[ "$OSTYPE" == "linux-gnu" ]]; then
+        sudo cp ../../../libvcx/target/debug/libvcx.so /usr/lib/libvcx.so
```
jovfer: I suggest to use env variables instead of moving artifacts to system directories

nehalshah50: which env variable I should use?

jovfer: let me share context and reason of my suggestion to be more clear.
This PR introduce demo in java wrapper. And currently we are discussing a script in this demo called  `buildlibs.sh`
But in fact this script perform installation of manual built of libindy and libnullpay to *system* path (/usr/lib). It may be unexpected by user and damage his environment, like replace system stable version of libindy by something unstable or even experimental or broken.
So main point from my side is to avoid actual installation libraries to system path, but rather to use some env variable to make it available for demo, or copy it to local folder, or give a chance to user to choose the behavior.

jovfer:
Please fix DCO check https://github.com/hyperledger/indy-sdk/pull/1488/checks
and also totally exclude binaries from the git history

jovfer (approved):

kukgini:
Since VCX is forked and will be removed here, I think this PR should be dropped in this repository. https://github.com/hyperledger/indy-sdk/pull/2416

PR #1496, opened 2019-02-26
macos openssl 1.0.2q

Artemkaaas:


Water-bamboo:
How to build pass on iOS platform? but not masos

PR #1505, opened 2019-02-28
Restart catchup in case of outdated cache

Artemkaaas:
Fix for IS-1200: Client can't connect pool using valid genesis file


libindy/src/services/pool/catchup.rs
```diff
@@ -71,8 +72,13 @@ pub fn check_nodes_responses_on_status(nodes_votes: &HashMap<(String, usize, Opt
             let positive_votes_cnt = votes_cnt + (node_cnt - reps_cnt);
             let is_consensus_reachable = positive_votes_cnt < node_cnt - f;
             if is_consensus_reachable {
-                //TODO: maybe we should change the error, but it was made to escape changing of ErrorCode returned to client
-                return Err(err_msg(IndyErrorKind::PoolTimeout, "No consensus possible"));
+                if merkle_tree_factory::drop_cache(pool_name).is_ok() {
```
jovfer: if we decide to drop cache in case of catchup failure we probably have to drop it on timeout (few lines above)

libindy/src/services/pool/catchup.rs
```diff
@@ -52,41 +52,50 @@ pub fn check_nodes_responses_on_status(nodes_votes: &HashMap<(String, usize, Opt
...
+
+    let most_popular_not_timeout =
+        votes
+            .iter()
+            .max_by_key(|entry| entry.1);
+
+    let timeout_votes = timeout_votes.iter().last();
+
+    if let Some((most_popular_not_timeout_vote, vote_cnt)) = most_popular_not_timeout {
+        if *vote_cnt == f + 1 {
```
jovfer: else branch should cover case when consensus not possible

PR #1506, opened 2019-02-28
introduce CommandHandle WalletHandle etc

AxelNennker:
As per https://jira.hyperledger.org/browse/IS-1134 introduce more specific types for handles


jovfer:
@AxelNennker thank you for this painstaking work!
I'll try to review this PR while my flight this weekend.

AxelNennker:
Safe travels!

PR #1507, opened 2019-02-28
Corrected response handling in Consensus state

Artemkaaas:
Fix for IS-1137


libindy/src/services/pool/request_handler.rs
```diff
@@ -335,7 +335,12 @@ impl<T: Networker> RequestSM<T> {
                                 (RequestState::finish(), None)
                             }
                         } else {
-                            (RequestState::Consensus(state), None)
+                            state.denied_nodes.insert(node_alias.clone());
+                            if state.denied_nodes.len() + state.replies.len() == nodes.len() {
```
jovfer: should we also send status to awaiting requests?

PR #1517, opened 2019-03-07
IS-1212 Incremented Rust version

Artemkaaas:


jovfer:
Seems like it's not useful for IS-1212

PR #1518, opened 2019-03-07
Parallel tests

AxelNennker:
First step towards parallel tests
https://jira.hyperledger.org/browse/IS-1135

With the PR now many tests use their own wallet to test something and delete it afterwards.
Some global things remain for now. Marked some with FIXME.

This PR changes many file. Should you approve please merge soon because I fear a ton of merge conflicts otherwise.


libindy/tests/utils/anoncreds.rs
```diff
@@ -454,130 +454,149 @@ pub fn tails_writer_config() -> String {
...
+    wallet::close_wallet(wallet_handle).unwrap();
+
+    (issuer1_gvt_credential_def_json,
+     issuer1_gvt_credential_offer,
+     issuer1_gvt_credential_req,
+     issuer1_gvt_cred,
+     config)
+}
+
+pub fn init_common_wallet_once() -> (&'static str, &'static str, &'static str, &'static str) {
```
jovfer: @AxelNennker I don't see usage of this function in the PR. Is it temporary state or you are going to delete it?

AxelNennker: I think that is a relict of the former code where all tests use _one_ wallet and that is reused. So there is this 'once' stuff that ensures that the wallet is only initialized once.

I kept it because there are some or one test (parallel_auth_encrypt) that tests encryption in parallel and I thought that maybe that test  uses one common wallet. But it turned out that that test also uses separate wallets for the each thread it starts. But I thought maybe we should have a test where several threads work on one wallet in parallel and so I kept this method.

jovfer: @AxelNennker There is critical drawback introduced by current PR:
Wallet preparation for Anoncreds is very long operation and prepare wallet every time from scratch (even in parallel) will increase time of testing significantly.

I'm afraid that benefit from parallel run of tests will be totally eliminated by anoncreds wallet preparations.

AxelNennker: Problem is that CredentialDefinitions etc are generated without the possibility to NOT store them in a wallet. So they are generated again an again. 
I guess there is nothing we can do - speed things up?
Not sure whether creating one wallet exporting it and importing it into the separate wallets fast enough. Which would leave that one wallet in the indy_home. Maybe data generated during tests should be in the target directory instead of $HOME/.indy_client/ 
I'll look into excluding anoncreds from this PR or ...
I'm open for suggestions. More faster CI/CD servers? Running all the tests locally was never fast...

jovfer: I like the idea to try to export smth and re-use it in tests

AxelNennker: I'll try something. Although the first merge conflicts appeared. Maybe excluding the anoncread changes from this PR gets us forward and fix the anoncreds stuff later. Breaking this up in smaller chunks might be a good idea anyway.

jovfer: @AxelNennker it's up to you. To accept this PR I see 2 options right now:
1) implement importing wallet for anoncreds tests
2) exclude changes around shared wallet for anoncreds from this PR

jovfer:
@Artemkaaas @mattraffel I suggest to merge this PR as soon as https://github.com/hyperledger/indy-sdk/pull/1518#discussion_r266491335 will be resolved

libindy/tests/pairwise.rs
```diff
@@ -281,31 +281,31 @@ mod high_cases {
...
             assert_eq!(format!(r#"{{"my_did":"{}"}}"#, my_did), pairwise_info_without_metadata);
 
-            utils::tear_down_with_wallet(wallet_handle);
+            utils::tear_down_with_wallet(wallet_handle, "indy_set_pairwise_metadata_works_for_reset", &wallet_config);
         }
 
         #[test]
         fn indy_set_pairwise_metadata_works_for_not_created_pairwise() {
-            let wallet_handle = utils::setup_with_wallet();
+            let (wallet_handle, wallet_config) = utils::setup_with_wallet("indy_set_pairwise_metadata_works_for_not_created_pairwise");
```
Artemkaaas: `setup_with_wallet` can generate and return random string (as wallet name) instead of explicit passing.

AxelNennker: It is easier to see this way what did not get cleaned up.

libindy/src/services/wallet/mod.rs
```diff
@@ -816,139 +816,167 @@ mod tests {
...
+        test::cleanup_wallet("wallet_service_create_wallet_works_for_moderate_key_derivation");
+        {
+            let wallet_service = WalletService::new();
+            wallet_service.create_wallet(&_config_default("wallet_service_create_wallet_works_for_moderate_key_derivation"), &ARGON_MOD_CREDENTIAL, (&MODERATE_KDD, &MODERATE_MASTER_KEY)).unwrap();
+        }
+        test::cleanup_wallet("wallet_service_create_wallet_works_for_moderate_key_derivation");
     }
 
     #[test]
+    #[ignore]
```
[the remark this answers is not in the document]
AxelNennker: Because it always fails on my machine and it is unclear whether the resolution of the system clock is high enough to detect the difference of nearly equal values. I think this test does not make sense, please explain if you think it does.

libindy/src/services/wallet/mod.rs
```diff
@@ -816,139 +816,167 @@ mod tests {
...
     }
 
     #[test]
     fn wallet_service_create_wallet_works_for_interactive_key_derivation() {
-        _cleanup();
-
-        let wallet_service = WalletService::new();
-        wallet_service.create_wallet(&_config_default(), &ARGON_INT_CREDENTIAL, (&INTERACTIVE_KDD, &INTERACTIVE_MASTER_KEY)).unwrap();
+        test::cleanup_wallet("wallet_service_create_wallet_works_for_interactive_key_derivation");
+        {
```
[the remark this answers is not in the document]
AxelNennker: The variables are dropped when the block ends and there is a drop implementation if I remember correctly. 
In the next step regarding IS-1135 I am going to introduce a test! macro that wraps the block inside setup-cleanup.

libindy/src/services/pool/merkle_tree_factory.rs
```diff
@@ -348,11 +353,13 @@ mod tests {
 
         assert_eq!(node_state["Gw6pDLhcBcoQesN72qfotTgFa7cbuqZpkX3Xo6pLhPhv"], node1);
         assert_eq!(node_state["8ECVSk179mjsjKRLWiQtssMLgp6EPhWXtaYyStWPSGAb"], node2);
+
+        test::cleanup_storage("pool_worker_build_node_state_works_for_old_format");
     }
 
     #[test]
     fn pool_worker_build_node_state_works_for_new_format() {
```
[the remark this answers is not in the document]
AxelNennker: Overlooked this response yesterday evening, sorry. Have to look into why this fails for you. CI succeeds. I would like not to change more than necessary in this PR as it already changes a lot.

PR #1522, opened 2019-03-11
Wallet handle

AxelNennker:
Replace i32 by newtype WalletHandle where the i32 is a wallet handle.
https://doc.rust-lang.org/1.0.0/style/features/types/newtype.html


libindy/src/api/mod.rs
```diff
@@ -19,7 +19,12 @@ use utils::ctypes;
 
 pub type IndyHandle = i32;
 
-pub type WalletHandle = i32;
+//pub type WalletHandle = i32;
+#[derive(Debug, Hash, PartialEq, Eq, Copy, Clone)]
```
jovfer: It should be possible (and it will be very helpful and safe) to force Rust compiler to use i32 representation

AxelNennker: Yes, I added a repr transparent.
Rust documentation is here: https://doc.rust-lang.org/1.26.2/unstable-book/language-features/repr-transparent.html

vimmerru:
@jovfer @AxelNennker 

Just one suggestion to discuss. It was obviously bad decision to use i32 as handle type instead of usize or i65. i32 can't store pointer on x64 systems and as a result we need store Map<i32, cnxt> instead of just cnxt pointer. It causes that we can't avoid synchronization by accessing context by handle that complicates the code and reduces performance. As we define handle types it can be good idea to switch to pointer size types in future versions.

AxelNennker:
The advantage of i32 is that it is really opaque to the foreign code. No way for it to manipulate the handle or peek inside to manipulate the content. For libindy I see no disadvantages in using i32.

The only pointers libindy is exposing are those to strings which can't be helped. I often wish that not everything in libindy's API was a `*const c_char` (or i32). But that is the price for the best portability, I guess.

jovfer:
@AxelNennker I think @vimmerru mean the next scenario:
if cmd_handle has size equal (or greater) to pointer size, then application can put context pointer (cast to 32/64-bit integer) as command handle and while executing resulting callback just restore context from pointer. So there is no memory pointer shared from libindy to app.

vimmerru:
@AxelNennker @jovfer 

Obvious example https://github.com/hyperledger/indy-sdk/blob/master/wrappers/rust/src/utils/callbacks.rs 

i32 type requires maps, mutexes or dynamic function allocation (like in python)

AxelNennker:
Yes, the wrappers are special in that they hide the handles from the app.
I am at odds with the wrappers for the reason that they are used by libindy/tests and thus currently create a dev-dependency outside the crate of libindy. That is evil.
https://github.com/hyperledger/indy-sdk/blob/master/libindy/Cargo.toml#L96

I am not sure about the value of the Rust wrappers. Should I ever need to write a Rust application that needs code from Indy I think I would be tempted to just use e.g. Wallet directly. Then I would not have to deal with the C-API and all the Commands and Services.  
https://github.com/hyperledger/indy-sdk/blob/master/libindy/src/services/wallet/wallet.rs#L122
(I hope this thinking does not hurt the Indy architects feelings. In that case, sorry)
I know there is no guaranty for that API e.g. Wallet to be stable over time, so this might cost me later, but... live fast and break things.

PR #1537, opened 2019-03-15
Avoid unnecessary base64 encode

lovesh:


kdenhartog:
@lovesh can we get one of the BCGov team guys ( @swcurran can you let the person who worked on this know of this change ) to check and make sure the VON implementation works with these changes.

swcurran:
@cywolf should look at this.  He and @telegramsam did some work on this last week and Andrew has a test script that can be used to look into this.

jovfer:
@lovesh @kdenhartog @swcurran @TelegramSam I think update for the HIPE should be raised and discussed before merging this PR

libindy/src/commands/crypto.rs
```diff
@@ -397,15 +397,73 @@ impl CryptoCommandExecutor {
         // Use AEAD to encrypt `message` with "protected" data as "associated data"
         let (ciphertext, iv, tag) =
             self.crypto_service
-                .encrypt_plaintext(message, &base64_protected, &cek);
+                .encrypt_plaintext(message, &json_str, &cek);
 
         self._format_pack_message(&base64_protected, &ciphertext, &iv, &tag)
     }
 
+    pub fn unpack_msg(&self, jwe_json: Vec<u8>, wallet_handle: WalletHandle) -> IndyResult<Vec<u8>> {
```
jovfer: @lovesh please restore initial location of this function against others in the file. I would like to have clear diff for discussion in progress. And when we will be ready to merge - we can change the order

jovfer:
@lovesh should we close this PR in favor of https://github.com/hyperledger/indy-sdk/pull/1574

lovesh:
@jovfer The PR #1574 seem controversial whereas this PR is not.

jovfer:
Ok. I have no objections to merge this PR from code point of view. But it seems like breaking changes from data point of view and also requires small update of the HIPE + approve from community.

lovesh:
@jovfer Here is the HIPE PR https://github.com/hyperledger/indy-hipe/pull/115

kdenhartog:
I'll reach out to all the people who I know have independently implemented this to try and move this update forward. I'm alright with this even if it's a breaking change because we declared this experimental.

kdenhartog:
As I'm reading back through the JWE spec again, I think the way it's implemented is as described in the JWE spec. See steps 13, 14, and 15 described here https://tools.ietf.org/html/rfc7516#section-5.1

lovesh:
@kdenhartog Yes, you are correct.

kdenhartog:
Do you know what the efficiency gains by doing this then? I'd like to pragmatically weigh the optimization against the breaking changes and the JWE compliance (less significant in my mind, because people can't use off the shelf JWE libraries for this right now anyways). 

Also, DIF has setup biweekly calls to start discussing work that could impact this implementation. If we're going to perform Breaking changes, do you mind holding off a bit so we can do them all at once? (From my understanding, every implementation is willing to make the changes, so it's breaking - but it shouldn't have an impact going forward)

lovesh:
@kdenhartog I have not benchmarked it but they won't be substantial as this affects only the tag computation.

kdenhartog:
Given the current implementation is what is described in the JWE spec, and the computational efficiency isn't substantial, I would vote to not merge this PR. I think we should keep it as close as possible at this level and focus on finding efficiencies at the multi-wrap layer which has been described in the HIPE PR https://github.com/hyperledger/indy-hipe/pull/112

PR #1539, opened 2019-03-17
replace IndyHandle by separate types e.g. WalletHandle

AxelNennker:

Created types for each kind of handle e.g. WalletHandle, CommandHandle, ...

Also moved the type definition of IndyHandle in https://github.com/hyperledger/indy-sdk/blob/master/wrappers/rust/src/lib.rs#L40 to https://github.com/hyperledger/indy-sdk/blob/master/wrappers/rust/indy-sys/src/lib.rs because all types used in the api should be defined in indy-sys

To achieve real type safety more work needs to be done


wrappers/rust/src/blob_storage.rs
```diff
@@ -1,38 +1,39 @@
...
 use std::ffi::CString;
 
 use ffi::blob_storage;
 use ffi::ResponseI32CB;
 
 use utils::callbacks::{ClosureHandler, ResultHandler};
+use ffi::CommandHandle;
 
-pub fn open_reader(xtype: &str, config_json: &str) -> Box<Future<Item=IndyHandle, Error=IndyError>> {
+pub fn open_reader(xtype: &str, config_json: &str) -> Box<Future<Item=CommandHandle, Error=IndyError>> {
```
jovfer: Is it really CommandHandle in the result? AFAIR it should be blob reader handle, but I may be wrong

AxelNennker: actually that seems to be something I don't have yet. It is a ConfigHandle.
https://github.com/hyperledger/indy-sdk/blob/master/libindy/src/services/blob_storage/mod.rs#L128

AxelNennker: It seems I would have to add something like ResponseI32CB for each new type that replaces i32 e.g. ResponseBlobStorageConfigHandleCB. Which is a lot of work. Probably this should be done type by type.
Some more macros might help. Maybe reintroduce IndyHandle for now?

The amount of boilerplate in indy-sdk is quite overwhelming.

jovfer: I think temporary usage of IndyHandle until we will have all separates type is better, rather then using CommandHandle as result type for opened blob reader.

PR #1546, opened 2019-03-19
[WIP] A2A part of IndySDK 2.0

Artemkaaas:


docs/a2a-part-of-indy-sdk-2.0/README.md
```diff
@@ -0,0 +1,656 @@
...
+
+#### Choreography Diagram
+
+![Choreography Diagram](./credential-issuance.png)
+
+#### Workflow
+
+The Credential Issuance positive workflow is demonstrated on the following diagram:
+
+![Workflow](credential-issuance-positive-flow.svg)
```
Patrik-Stas: Unless I am missing something, the way IndySDK is currently designed requires prover to receive credential offer, before credential request can be submitted by prover. It seems from the diagram like this restriction will stay. Why?
If I have connections with an institution, it seems natural that I should be able to request something from them. 

In my organization we've implemented credential exchange prototype between mobile app/backend on top of IndySDK without VCX. We basically ended up with equivalent of something like "request credential offer" message type. Did we miss something? 
Aditionally we've also implemented some API on issuer side which enabled the receiving side to ask issuer what kind of credentials can he issue. So that was typically the API the credential requester called first. Once he knew what's there to be offered, he would send "request credential offer" message asking for a kind of credential. The issuer decided whether the requester is "eligible" to get such credential based some non-indy application logic/data.
Did we not "get" something about IndySDK design? Is/will be flow of an agent directly requesting other agent for specific type of credential supported?

dhh1128: @Patrik-Stas There is an attack that can be performed by a credential issuer if they do not present an offer. It allows them to embed data into a credential that the requester does not know about.

dhh1128: @brentzundel or @mikelodder7 or @lovesh may be able to clarify.

Patrik-Stas: @dhh1128 I see, fair enough if this kind of 3 step credential exchange protocol is security requirement. However, what about the functional requirement? Having the ability to ask someone to give you a credential of certain type?
Wouldn't it make sense to have something like "credential offer request" as a standard message type? (sounds awkward, a better name could be given).

lovesh: @Patrik-Stas The need for sending credential offer is for the issuer to send a nonce to ensure that the holder does know the value of the committed attributes. One such attribute is link secret. The holder never reveals the link secret, he only sends a commitment to the issuer. The holder uses the nonce sent as part of credential offer to send a proof (prior to issuance) to the issuer that he does know the link secret he is committing to. The nonce does not have to sent by the issuer always, the prover can select a nonce from another source like the current merkle root of the ledger or a randomness beacon.

> what about the functional requirement? Having the ability to ask someone to give you a credential of certain type?

Yes, we can introduce a message by holder indicating the "type" of credential it needs.

mikelodder7: The important part is the holder shouldn't be allowed to generate the nonce otherwise cheating is possible. Selecting a nonce like the merkle root that the issuer can verify is acceptable. Otherwise, like @lovesh says you need to have the nonce come from the issuer.

docs/a2a-part-of-indy-sdk-2.0/credential-presentation-positive-flow.puml
```diff
@@ -0,0 +1,94 @@
...
+
+P -[#green]> PH: <b> parse message
+P <-[#green]- PH: presentation ack message
+P -[#grey]-> P: application logic
+P -[#blue]> PSM: <b> check transition for presentation ack message
+PSM -[#blue]-> P: is_transition_acceptable
+P -[#blue]> PSM: <b> do step for presentation ack message
+PSM -[#blue]-> PSM: state: <b> PresentationAckReceived
+P -[#grey]-> P: application logic
+P -[#blue]> PSM: <b> do step
```
jovfer:
```suggestion
P -[#blue]-> PSM: <b> do step
```

PR #1562, opened 2019-03-26
Vcx mobile builds

stevengubler:
This only addresses android builds for now. 

@KitHat @Artemkaaas, could you guys take a look and review the logic in `androidBuildArtifact`?


NanoBytesInc:
Hey! I am not sure if this is the right place to do this, but it seemed like the same vein (and there are seriously no other ways to let you guys know that something is broken)

The Java wrapper is no longer working because the maven repository is down

```
https://repo.sovrin.org/repository/maven-public
```

@Artemkaaas can you look at this? Or point us at another repo?

stevengubler:
@NanoBytesInc You're correct, this is not the best place to report this kind of issue. :) In the future, the quickest way to get in contact with us is to reach out to the Sovrin devops team at **#devops** on our [chat.](https://chat.sovrin.org/) 

Were we're already aware of the issue and it has since been fixed.

PR #1570, opened 2019-04-01
IS-1199: Updated objective C wrapper

Artemkaaas:


cobear25:
I was just about to create this same PR myself. It looks like version 1.8.2 is finally there and working with the latest podspec, after running pod install though it doesn't seem to actually bring in the framework, it's just an empty pod. Any idea what might be going on there?

Artemkaaas:
@cobear25  Thanks. We will fix it. a nested folder `libindy-objc` is needed.

PR #1576, opened 2019-04-04
IS-1225: Fixing boolean type casting bug on windows

farskipper:
https://jira.hyperledger.org/browse/IS-1225

Indy boolean values were defined as unsigned integers. This works just fine on linux.
`typedef unsigned int  indy_bool_t`

However, this causes issues on windows because the extra memory is not managed.

| rust  | windows int  | hex value  |
| ----- | ---------- | ---------- |
| true  | 1527741697 | 0x5B0F7D01 |
| false | 543435264  | 0x20642A00 |

Rust only emits the 1 byte for booleans. But `unsigned int` tells the c++ compiler to look at 2 or 4 bytes. 

I propose we change it it to 
`typedef   bool     indy_bool_t`
This gives a hint to the windows c++ compiler and tells it to do the right thing. This also works fine on linux.


farskipper:
I changed it to `typedef unsigned char indy_bool_t` b/c I've heard `bool` may not be supported by all compilers. However, as far as I can tell, `unsigned char` will always be a 1 byte number, which is what rust emits.

This is working on linux and windows when I've tested.

jovfer:
@farskipper thank you for the this catch and the PR. After some additional investigation I decide to return back to your first proposal `bool` as it's supported on our target platforms and seems more accurate than char.

PR #1577, opened 2019-04-05
Make sure NodeJS VCX wrapper can work in 64bit systems

Patrik-Stas:
functions `sign(<buffer>)` and `verify(<buffer>, <buffer>)` don't work on 64 bit systems. I have 64bit NodeJS and libindy,vcx and nullpay compiled as 64bit libraries on OSX. NodeJS wrapper is using `ref` library to get memory address of the buffer to pass it down to rust binaries. However, because the original FFI type declarations are all `uint32`, it fails with error:
```
TypeError: “value” argument is out of bounds
```
as it's trying to pass address higher than 2^32, which current FFI declaration doesn't seem to allow.

The issue is fixed if I run my app with VCX changes bellow with env variable `LIBVCX_FFI_ARCHITECTURE=x64`

I don't really having to pass env variable to my app in order to make its dependency work properly. Do you think it would be better to specify this at VCX transpile time?


vcx/wrappers/node/src/rustlib.ts
```diff
@@ -23,14 +30,14 @@ export const FFI_CONNECTION_DATA = 'string'
...
-export const FFI_COMMAND_HANDLE = 'uint32'
-export const FFI_CREDENTIAL_HANDLE = 'uint32'
-export const FFI_PROOF_HANDLE = 'uint32'
-export const FFI_CREDENTIALDEF_HANDLE = 'uint32'
-export const FFI_SCHEMA_HANDLE = 'uint32'
-export const FFI_SCHEMA_NUMBER = 'uint32'
-export const FFI_PAYMENT_HANDLE = 'uint32'
-export const FFI_PRICE = 'uint32'
+export const FFI_COMMAND_HANDLE = FFI_UINT
+export const FFI_CREDENTIAL_HANDLE = FFI_UINT
```
jovfer: I'm not familiar with whole VCX API, but I think current changes may be incorrect for some part of API.
Like `vcx_issuer_credential_serialize` define u32 signature on public C API endpoint.

jovfer:
@glowkey @ryMarsh44 please review. I have some concerns and would like to double-check. Please see my comment above

vcx/wrappers/node/src/rustlib.ts
Patrik-Stas: @jovfer You might be right. I think to fix the particular issue which I've had, it's probably sufficient to do this change only for `FFI_UNSIGNED_INT_PTR`.
I can test it and if it works, I'll update pull request with this change only for the `FFI_UNSIGNED_INT_PTR`.

jovfer: @Patrik-Stas could you double-check all these constants and compare with original rust API?

I think proposal in your comment is correct: only `FFI_UNSIGNED_INT_PTR` should be fixed and probably there are some misuse or missed usage of it (at lease I can see one in `vcx/wrappers/node/src/rustlib.ts` `vcx_connection_sign_data` )

vcx/wrappers/node/src/rustlib.ts
```diff
@@ -236,10 +243,10 @@ export const FFIConfiguration: { [ Key in keyof IFFIEntryPoint ]: any } = {
     FFI_CALLBACK_PTR]],
   vcx_connection_send_message: [FFI_ERROR_CODE, [FFI_COMMAND_HANDLE, FFI_CONNECTION_HANDLE, FFI_STRING_DATA,
     FFI_STRING_DATA, FFI_STRING_DATA, FFI_CALLBACK_PTR]],
-  vcx_connection_sign_data: [FFI_ERROR_CODE, [FFI_COMMAND_HANDLE, FFI_CONNECTION_HANDLE, FFI_UNSIGNED_INT,
+  vcx_connection_sign_data: [FFI_ERROR_CODE, [FFI_COMMAND_HANDLE, FFI_CONNECTION_HANDLE, FFI_UNSIGNED_INT_PTR,
     FFI_UNSIGNED_INT, FFI_CALLBACK_PTR]],
-  vcx_connection_verify_signature: [FFI_ERROR_CODE, [FFI_COMMAND_HANDLE, FFI_CONNECTION_HANDLE, FFI_UNSIGNED_INT,
-    FFI_UNSIGNED_INT, FFI_UNSIGNED_INT, FFI_UNSIGNED_INT, FFI_CALLBACK_PTR]],
+  vcx_connection_verify_signature: [FFI_ERROR_CODE, [FFI_COMMAND_HANDLE, FFI_CONNECTION_HANDLE, FFI_UNSIGNED_INT_PTR,
```
Patrik-Stas: Correcting according to what's in rust:
```
pub extern fn vcx_connection_verify_signature(command_handle: u32,
                                  connection_handle: u32,
                                  data_raw: *const u8,
                                  data_len: u32,
                                  signature_raw: *const u8,
                                  signature_len: u32,
                                  cb: Option<extern fn(command_handle_: u32,
                                                       err: u32,
                                                       valid: bool)>) -> u32 {
```

vcx/wrappers/node/src/rustlib.ts
```diff
@@ -236,10 +243,10 @@ export const FFIConfiguration: { [ Key in keyof IFFIEntryPoint ]: any } = {
     FFI_CALLBACK_PTR]],
   vcx_connection_send_message: [FFI_ERROR_CODE, [FFI_COMMAND_HANDLE, FFI_CONNECTION_HANDLE, FFI_STRING_DATA,
     FFI_STRING_DATA, FFI_STRING_DATA, FFI_CALLBACK_PTR]],
-  vcx_connection_sign_data: [FFI_ERROR_CODE, [FFI_COMMAND_HANDLE, FFI_CONNECTION_HANDLE, FFI_UNSIGNED_INT,
+  vcx_connection_sign_data: [FFI_ERROR_CODE, [FFI_COMMAND_HANDLE, FFI_CONNECTION_HANDLE, FFI_UNSIGNED_INT_PTR,
```
Patrik-Stas: Correcting according to what's in Rust:
```
pub extern fn vcx_connection_sign_data(command_handle: u32,
                                  connection_handle: u32,
                                  data_raw: *const u8,
                                  data_len: u32,
                                  cb: Option<extern fn(command_handle_: u32,
                                                       err: u32,
                                                       signature_raw: *const u8,
                                                       signature_len: u32)>) -> u32  {
```

vcx/wrappers/node/src/rustlib.ts
```diff
@@ -9,12 +9,19 @@ export const VcxStatus = StructType({
...
+const UINTS_TYPES: IUintTypes = { x86: 'uint32', x64: 'uint64' }
+const ARCHITECTURE: string = process.env.LIBVCX_FFI_ARCHITECTURE || 'x86'
+const FFI_UINT: string = UINTS_TYPES[ARCHITECTURE]
+
 // FFI Type Strings
 export const FFI_ERROR_CODE = 'int'
 export const FFI_BOOL = 'bool'
 export const FFI_CONNECTION_HANDLE = 'uint32'
 export const FFI_UNSIGNED_INT = 'uint32'
-export const FFI_UNSIGNED_INT_PTR = ref.refType('uint32')
```
Patrik-Stas: I don't much about `ffi` library, but `FFI_UNSIGNED_INT_PTR` was not actually used anywhere. So I've fixed signatures of `vcx_connection_sign_data ` and `vcx_connection_verify_signature ` to use this declaration, as they should according to Rust function signatures. 
However, if I change signatures and simply keep `FFI_UNSIGNED_INT_PTR = ref.refType('uint32')`, it doesn't work. When using `ref.refType`, sign function will fail like this:
```
Demo Faber: [error]: Unknown Error: error setting argument 2 - writePointer: Buffer instance expected as third argument
Demo Faber: [error]: Error: Unknown Error: error setting argument 2 - writePointer: Buffer instance expected as third argument
    at new VCXInternalError (/Users/prague/dev/hyperledger/indy-sdk/vcx/wrappers/node/dist/src/errors.js:10:23)
    at Connection._callee4$ (/Users/prague/dev/hyperledger/indy-sdk/vcx/wrappers/node/dist/src/api/connection.js:232:23)
    at tryCatch (/Users/prague/dev/hyperledger/indy-sdk/vcx/wrappers/node/node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:65:40)
    at Generator.invoke [as _invoke] (/Users/prague/dev/hyperledger/indy-sdk/vcx/wrappers/node/node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:303:22)
    at Generator.prototype.(anonymous function) [as throw] (/Users/prague/dev/hyperledger/indy-sdk/vcx/wrappers/node/node_modules/babel-polyfill/node_modules/regenerator-runtime/runtime.js:117:21)
    at rejected (/Users/prague/dev/hyperledger/indy-sdk/vcx/wrappers/node/dist/src/api/connection.js:5:47)
    at <anonymous>
```
So I changed this line to 
```
export const FFI_UNSIGNED_INT_PTR = FFI_UINT
```
and with this, demo works fine.

Effectively nothing has changed. `ref.refType` was not previously used for the `vcx_connection_sign_data ` and `vcx_connection_verify_signature ` anyways. 
Whether it should be used, I don't know, is there a more `ffi` knowledgeable person around here?

Patrik-Stas: You can try out both cases with
`export const FFI_UNSIGNED_INT_PTR = ref.refType('uint32')`
and 
`export const FFI_UNSIGNED_INT_PTR = FFI_UINT`
By running new demo files, the commands to run them are added to `package.json`

vcx/wrappers/node/demo/faber-verify-signature.js
```diff
@@ -0,0 +1,74 @@
+import {CredentialDef} from "../dist/src/api/credential-def";
```
Patrik-Stas: These new demo files share common code with the classic credential exchange demo - lot's of code could be extracted and shared here, but I didn't want to do too many different things in this pull request. The point of demo here is really just to demonstrate the changes I made to `rustlib.ts`. Demo refactoring could be a separate pull request, I think.

Patrik-Stas:
Hi @jovfer I got back to this after a while with new changes and signing/signature verification demo script.

PR #1578, opened 2019-04-05
Ursa migration

mikelodder7:


libindy/src/commands/anoncreds/prover.rs
```diff
@@ -279,7 +279,7 @@ impl ProverCommandExecutor {
 
         let credential_request_metadata = CredentialRequestMetadata {
             master_secret_blinding_data: ms_blinding_data,
-            nonce: credential_request.nonce.clone()?,
+            nonce: credential_request.nonce.try_clone()?,
```
kdenhartog: leaving a note for myself to compare the differences between **clone()** and **try_clone()**

libindy/src/commands/anoncreds/tails.rs
```diff
@@ -80,7 +80,7 @@ pub fn store_tails_from_generator(service: Rc<BlobStorageService>,
     let version = vec![0u8, TAILS_BLOB_TAG_SZ];
     service.append(blob_handle, version.as_slice())?;
 
-    while let Some(tail) = rtg.next()? {
+    while let Some(tail) = rtg.try_next()? {
```
kdenhartog: compare differences between **next()** and **try_next()**

libindy/src/services/anoncreds/prover.rs
```diff
@@ -373,7 +373,22 @@ impl Prover {
...
                     .to_indy(IndyErrorKind::InvalidStructure, format!("Credential attribute value \"{:?}\" is invalid", attribute_value))?;
                 Ok(attribute_value >= predicate.p_value)
-            }
+            },
+            PredicateTypes::GT => {
+                 let attribute_value = attribute_value.parse::<i32>()
+                    .to_indy(IndyErrorKind::InvalidStructure, format!("Credential attribute value \"{:?}\" is invalid", attribute_value))?;
+                 Ok(attribute_value > predicate.p_value)
+             },
+             PredicateTypes::LE => {
```
kdenhartog: Are we leaving off "Greater than or Equal to" PredicateType for a reason?

libindy/src/services/pool/networker.rs
```diff
@@ -182,7 +183,8 @@ impl PoolConnection {
     fn new(mut nodes: Vec<RemoteNode>, active_timeout: i64, preordered_nodes: Vec<String>) -> Self {
         trace!("PoolConnection::new: from nodes {:?}", nodes);
 
-        thread_rng().shuffle(nodes.as_mut());
+        nodes.shuffle(&mut thread_rng());
+//        thread_rng().shuffle(nodes.as_mut());
```
kdenhartog: If this is not being used, let's delete this line.

libindy/src/services/wallet/language.rs
```diff
@@ -300,15 +300,16 @@ mod tests {
 
     use super::*;
     use self::rand::{thread_rng, Rng};
+    use rand::distributions::{Alphanumeric, Standard};
     use std::hash::Hash;
     use std::collections::HashSet;
 
     fn _random_vector(len: usize) -> Vec<u8> {
-        thread_rng().gen_iter().take(len).collect()
+        thread_rng().sample_iter(&Standard).take(len).collect()
```
kdenhartog: note to compare differences between **gen_iter()** and **sample_iter()**

libindy/tests/anoncreds.rs
```diff
@@ -3539,14 +3540,13 @@ mod medium_cases {
                "version":"0.1",
                "requested_attributes": json!({}),
                "requested_predicates": json!({
-                   "predicate1_referent": json!({ "name":"age", "p_type":"<=", "p_value":18 }),
+                   "predicate1_referent": json!({ "name":"age", "p_type":"!=", "p_value":18 }),
```
kdenhartog: Where is the change in p_type described in the API documentation? Can you add it, and additional tests that would support the new predicate types that are being added?

libindy/src/services/anoncreds/prover.rs
kdenhartog: This functionality isn't being exposed at the API layer, so adding this shouldn't be a reason to block this PR. It's prep work for when we decide to expose this to the API, at which point we should add tests.

libindy/tests/anoncreds.rs
kdenhartog: The change in p_types are not necessary at this point because the additional p_types aren't exposed at the API layer yet. This change is fine because it's a test to make sure the invalid p_type is properly throwing an error.

libindy/src/commands/anoncreds/prover.rs
kdenhartog: From what I can tell, this **try_clone()** function is performing a different operation than the **clone()** trait which was originally used. **try_clone()** appears to be generating a new BigNum. I got to this by looking at this line here: https://github.com/hyperledger/ursa/blob/0d03b1f071af652e3fa849241546dd0867ad3d51/libursa/src/bn/openssl.rs#L404

which then got me to look at the BigNum impl documentation in SSL, which says "Creates a new `BigNum` from an unsigned, big-endian encoded number of arbitrary length."

https://docs.rs/crate/openssl/0.10.20/source/src/bn.rs

Should we be generating a new nonce for this or do we want to be cloning an already created nonce?

libindy/src/commands/anoncreds/tails.rs
kdenhartog: The functionality appears to do the same things. However, this new function looks like it's generating a new Tails object each time and returning the new object rather than iterating on the same object. I was not able to find the impl of the **next()** trait in the std library to verify they do it differently. My only concern with this would be that it's not as memory efficient, which given the size of the Tails object may be a problem. Was there a reason for switching this?

libindy/src/services/wallet/language.rs
kdenhartog: From what I can tell, **sample_iter()** is not random. Is this intentional to constrain the test? In any case this shouldn't introduce any problems given it is used to generate randomness in tests.

libindy/src/commands/anoncreds/prover.rs
kdenhartog: No issues here. **clone()** is not a reference to a trait, but rather to an indy_crypto function found here. This is just a function rename to create more idiomatic rust. See the original function called here: https://github.com/hyperledger/indy-crypto/blob/1675e29a2a5949b44899553d3d128335cf7a61b3/libindy-crypto/src/bn/openssl.rs#L404

libindy/src/commands/anoncreds/tails.rs
kdenhartog: No problems here. This is not referring to the trait **next()** which is performed on an iterator, but rather to the **next()** function which can be found in Indy_crypto here: https://github.com/hyperledger/indy-crypto/blob/ca98135687a0674722f94836a2359ac5fd629b54/libindy-crypto/src/cl/mod.rs#L504

libindy/tests/crypto.rs
```diff
@@ -195,7 +195,7 @@ mod high_cases {
         use super::*;
 
         #[test]
-        fn indy_crypto_sign_works() {
+        fn ursa_sign_works() {
```
jovfer: Seems like tests names in this file was auto-replaced from composite prefixes `indy` `crypto` to `ursa`.
Please revert these changes.

libindy/src/services/pool/networker.rs
jovfer: +1

PR #1581, opened 2019-04-08
[WIP] Use one database for all wallets

jsaur:

This restructures the postgres storage plugin to use one database for all wallets, rather than a separate database for each wallet.
For your review @ianco

### DIFF
--- a/experimental/plugins/postgres_storage/Cargo.lock
+++ b/experimental/plugins/postgres_storage/Cargo.lock
@@ -903,7 +903,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "storage-postgres"
-version = "1.6.0"
+version = "1.7.0"
 dependencies = [
  "base64 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


ianco:
Thanks @jsaur this code is very good!

Regarding library initialization I see 3 options:

1. Have the library loader call the init function to create the database
2. Check if the database exists on wallet create, and create it if necessary (requires the create wallet function to have permissions to create database)
3. Have the postgres admin create the database in advance

Number 2 is sort of how it works now (database per wallet)

I also suggest keeping the existing scheme (one database per wallet) and make a configuration parameter so the postgres storage can be initialized to use either method (database per wallet or one database containing all wallets).

jsaur:
Talking it over with @ianco we want to allow both one wallet per database and all wallets in one database, which will probably involve a structure with one parent class and two children. Going to resume this again in late May.

jovfer:
@jsaur please remove "paused" label as soon as you resume work on this PR

cam-kiva:
@jovfer This is Cam (I'll be back on my normal github account when I am back in salt lake this should be temporary ). But we can remove the paused tag now... I will be working on this

jsaur:
@ianco has a better PR out there so closing this in favor of his

PR #1587, opened 2019-04-17
IS-841: Verifier verify proof request restrictions are satisfied #85

ryMarsh44:
-If a verifier requested a proof with specific attribute restrictions, the proof verification process would not ensure these requirements were met. This ticket ensures that if a proof request specifies a restriction (schema_id, schema_issuer_did, schema_name, schema_version, issuer_did, cred_def_id) the proof included specified requirements.

https://jira.hyperledger.org/browse/IS-841


libindy/src/commands/anoncreds/verifier.rs
```diff
@@ -116,6 +116,15 @@ impl VerifierCommandExecutor {
                                format!("Requested predicates {:?} do not correspond to received {:?}", requested_predicates, received_predicates)));
         }
 
+        if !self.anoncreds_service.verifier.verify_requested_restrictions(&self.anoncreds_service.prover,
+                                                                                     &proof_req,
+                                                                                     &proof,
+                                                                                     schemas,
+                                                                                     cred_defs)? {
+            warn!("Requested restrictions in Proof Request don't match Proof");
+            return Ok(false);
```
Artemkaaas: In my opinion, throwing `ProofRejected` is much better. 
 Libindy provides a function for getting error message and stack trace in addition to an error code. 
So, ProofRejected context should contain a message that describes what validation failed.
On every validation level function should return error with a message instead of bool flag. It will be much informative

Artemkaaas:
Also, add on every negative integration test devoted to restriction a positive one, please.

libindy/src/services/wallet/mod.rs
```diff
@@ -28,7 +28,7 @@ mod storage;
 mod encryption;
 mod query_encryption;
 mod iterator;
-mod language;
+pub mod language;
```
Artemkaaas: We rather need to think about moving the query language out of the wallet service to the proper place (utils may be).
I believe making it public is acceptable for the current, moment. We will change it later. 
Please, add a "TODO"  note.

libindy/src/commands/anoncreds/verifier.rs
ryMarsh44: I am trying to follow the same pattern that ```self.anoncreds_service.verifier.verify(...)``` is following. ```indy_verifier_verify_proof``` only throws an error if there is something structurally/formatted incorrectly. Otherwise, it returns a bool.

libindy/src/services/anoncreds/verifier.rs
```diff
@@ -127,4 +139,442 @@ impl Verifier {
...
         Ok(predicates_for_credential)
     }
+
+    pub fn verify_requested_restrictions(&self,
+                                         prover_service: &Prover,
+                                         proof_req: &ProofRequest,
+                                         full_proof: &Proof,
+                                         schemas: &HashMap<String, SchemaV1>,
+                                         cred_defs: &HashMap<String, CredentialDefinitionV1>) -> IndyResult<bool> {
+        let proof_attr_indexes = full_proof.requested_proof.unrevealed_attrs
```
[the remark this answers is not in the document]
ryMarsh44: This isn't the same comparison. This functionality collects all identifier indexes for a specific requested attribute or predicate so that the filters can be verified.

libindy/src/services/wallet/mod.rs
ryMarsh44: I added all the successful corresponding tests and added a `TODO`.

libindy/src/commands/anoncreds/verifier.rs
```diff
@@ -116,6 +116,15 @@ impl VerifierCommandExecutor {
                                format!("Requested predicates {:?} do not correspond to received {:?}", requested_predicates, received_predicates)));
         }
 
+        if !self.anoncreds_service.verifier.verify_requested_restrictions(&self.anoncreds_service.prover,
+                                                                                     &proof_req,
```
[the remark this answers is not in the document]
ryMarsh44: fixed.

PR #1588, opened 2019-04-18
prover credential deletion

sklump:

Credential deletion for prover wallets.

I've included wrapper code for Java, rust & python, and unit tests for Java and python (there are no anoncreds unit tests for the rust wrapper, which may be by design?). The feature would require indy group developers with know-how in dotnet, ios, javascript to extend their respective wrappers and unit tests accordingly.

At present, Ontario and BC use org-book wallets with millions of credentials, and there is still some churn in finessing them. Being able to delete credentials from a wallet would greatly help the effort, as an existing org-book wallet could survive the re-issue with all of its other keyed indy artifacts intact.

*ITEM:*
I've patched up the python non_secrets tests that were failing in master. It appears the principal cause here was that somewhere along the line `tag1` became `~tag1` in common.py.

*ADDENDUM, 2019-04-18 13:00 EDT*
Oh dear. It appears I have totally mangled this in with other work in progress. Please stand by. I will make this right.

*STATUS, 2019-04-18 15:45 EDT*
All tests pass. It's stable. My apologies for the drama.


sklump:
... I see now that there are a whole lot of unit tests under `indy-sdk/libindy/tests/` that I hadn't noticed before. There should be some cargo tests here to exercise the new API. I will get on that and update my PR.

libindy/tests/anoncreds.rs
```diff
@@ -3695,4 +3695,32 @@ mod medium_cases {
             assert_code!(ErrorCode::CommonInvalidStructure, res);
         }
     }
+
+    //NOTE: This test is destructive to the common wallet, removing a credential used in other tests: it goes last
+    mod prover_delete_credential {
```
Artemkaaas: I don't like an idea of tests based on running order.
I suggest leaving a test for `NotFoundError` only and moving deletion check to one of demo anoncreds tests or adding a new one.

libindy/src/api/anoncreds.rs
```diff
@@ -873,6 +873,49 @@ pub extern fn indy_prover_get_credential(command_handle: CommandHandle,
...
+/// Common*
+/// Wallet*
+#[no_mangle]
+pub extern fn indy_prover_delete_credential(command_handle: CommandHandle,
+                                         wallet_handle: WalletHandle,
+                                         cred_id: *const c_char,
+                                         cb: Option<extern fn(
+                                             command_handle_: CommandHandle,
+                                             err: ErrorCode)>) -> ErrorCode {
+    trace!("indy_prover_delete_credential: >>> wallet_handle: {:?}, cred_id: {:?}", wallet_handle, cred_id);
```
Artemkaaas: Please, add the function to `include/anoncreds.h` also.

libindy/tests/anoncreds.rs
sklump: Moved to a new anoncreds demo test.

jovfer:
@sklump could you take a look on compilation error? https://build.sovrin.org/job/indy-sdk/job/indy-sdk-ci/view/change-requests/job/PR-1588/13/execution/node/57/log/

seems like there was some refactoring and this pr should be updated a bit

sklump:
Sure, I'll take a look today.

sklump:
There is some janky stack overflow exception here in a part of the software that this PR does not touch.

Where do we go from here?

PR #1590, opened 2019-04-19
Code Coverage reporting

stevengubler:
This PR adds a Jenkinsfile and a dockerfile that enables us to gather test coverage. It works using kcov, a generic coverage tool for compiled languages, bash, and python. My thought is that we would have a separate job in jenkins that runs this on master once a week (because it is a costly operation), so that maintainers have a better view of what is being covered.

**Note:** libvcx is not included at this time because it does not use the same docker image for testing as the rest of indy-sdk. I can work to fix that soon.


stevengubler:
Closing because I forgot dco signoffs on some older commits. I'm just gonna squash them all

PR #1591, opened 2019-04-19
 Code coverage reporting

stevengubler:
This PR adds a Jenkinsfile and a dockerfile that enables us to gather test coverage. It works using [kcov](https://github.com/SimonKagstrom/kcov), a generic coverage tool for compiled languages, bash, and python. My thought is that we would have a separate job in jenkins that runs this on master once a week (because it is a costly operation), so that maintainers have a better view of what is being covered.

**Note:** libvcx is not included at this time because it does not use the same docker image for testing as the rest of indy-sdk. I can work to fix that soon.

### DIFF
--- a/Jenkinsfile.cov
+++ b/Jenkinsfile.cov
@@ -1,0 +1,104 @@
+#!groovy
+
+
+POOL_NET_NAME = 'pool-network'
+POOL_IP='10.0.0.2'
+
+
+testing()
+
+def testing () {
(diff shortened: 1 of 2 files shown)


stevengubler:
Also, if you look for the job `indy-sdk-cov` in jenkins, you should see the the job configuration using my fork of indy-sdk on branch `ci-test`. We would change that to this repo and `master` respectively.

PR #1595, opened 2019-04-22
ST-545 Removing of `set_txn_fees` builder 

Artemkaaas:


jovfer:
Set fees now will be used for aliases in auth rules

PR #1597, opened 2019-04-25
IS-1239:  Opened wallets are identified by id+path

Artemkaaas:


libindy/src/services/wallet/mod.rs
```diff
@@ -1028,6 +1035,38 @@ mod tests {
...
+            id: String::from("wallet_1"),
+            storage_type: None,
+            storage_config: None,
+        };
+
+        wallet_service.create_wallet(&config_1, &RAW_CREDENTIAL, (&RAW_KDD, &RAW_MASTER_KEY)).unwrap();
+        let handle_1 = wallet_service.open_wallet(&config_1,&RAW_CREDENTIAL).unwrap();
+
+        let config_2 = Config{
+            id: String::from("wallet_2"),
```
dhh1128: I don't see how this tests two wallets with the same ID. The two wallets here have different IDs.

Artemkaaas: fixed

PR #1599, opened 2019-04-25
IS-1229: Try to use `try_with` instead `with` for accessing to threadlocal variable to solve android error reading crash.

Artemkaaas:


AxelNennker:
This issue seems to be an indication that global variables are a bad idea especially in programs with threads. Time for a 2.0 version of the API that removes these error methods and the set protocol methods etc.
Global variables are not worth the maintenance cost.

PR #1605, opened 2019-04-30
Cache credential definitions and schemas

dkulic:


libindy/src/commands/non_secrets.rs
```diff
@@ -303,6 +366,256 @@ impl NonSecretsCommandExecutor {
...
+                            wallet_handle: WalletHandle,
+                            options_json: &str) -> IndyResult<()> {
+        trace!("purge_cred_def_cache >>> wallet_handle: {:?}, options_json: {:?}", wallet_handle, options_json);
+
+        let options = serde_json::from_str::<PurgeOptions>(options_json)
+            .to_indy(IndyErrorKind::InvalidStructure, "Cannot deserialize options")?;
+
+        let max_age = options.max_age.unwrap_or(-1);
+        let query_json = if max_age >= 0 {
+            let ts = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs() as i32;
```
jovfer: unwrap here may fail in theory on windows, isn't it? In this case I suggest to handle the error (may be as invalid state or even ignore somehow, not sure)

libindy/include/indy_non_secrets.h
```diff
@@ -257,6 +257,113 @@ extern "C" {
                                                                       indy_error_t  err)
                                                 );
 
+    /// Gets schema json data for specified schema id.
+    /// If data is present inside of cache, cached data is returned.
+    /// Otherwise data is fetched from the ledger and stored inside of cache for future use.
+    ///
```
jovfer: please add EXPERIMENTAL notice  here and everywhere in new API (headers, rust API, wrappers)

libindy/src/commands/non_secrets.rs
```diff
@@ -303,6 +366,256 @@ impl NonSecretsCommandExecutor {
...
+            }).to_string();
+            match self.wallet_service.get_record(wallet_handle, SCHEMA_CACHE, id, &options_json) {
+                Ok(record) => Ok(Some(record)),
+                Err(err) => if err.kind() == IndyErrorKind::WalletItemNotFound {Ok(None)} else {Err(err)}
+            }
+        } else {Ok(None)};
+        let cache = try_cb!(cache, cb);
+
+        if cache.is_some() {
+            let cache = cache.unwrap();
```
jovfer: please simplify by `if let Some(cache) = cache {`

libindy/src/commands/non_secrets.rs
```diff
@@ -303,6 +366,256 @@ impl NonSecretsCommandExecutor {
...
+            let min_fresh = options.min_fresh.unwrap_or(-1);
+            if min_fresh >= 0 {
+                let ts = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs() as i32;
```
jovfer: unwrap here may fail in theory on windows, isn't it? In this case I suggest to handle the error (may be as invalid state or even ignore somehow, not sure)

libindy/src/commands/non_secrets.rs
```diff
@@ -303,6 +366,256 @@ impl NonSecretsCommandExecutor {
...
+    }
+
+    fn _get_schema_continue(&self, wallet_handle: WalletHandle, ledger_response: IndyResult<(String, String)>, options: GetCacheOptions, cb_id: i32) {
+        let cb = self.pending_callbacks.borrow_mut().remove(&cb_id).expect("FIXME INVALID STATE");
+
+        let (schema_id, schema_json) = try_cb!(ledger_response, cb);
+
+        if !options.no_store.unwrap_or(false) {
+            let mut tags = Tags::new();
+            let ts = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs() as i32;
```
jovfer: I suggest to avoid unwrap here, see comment above for more details

libindy/src/commands/non_secrets.rs
```diff
@@ -303,6 +366,256 @@ impl NonSecretsCommandExecutor {
...
+            }).to_string();
+            match self.wallet_service.get_record(wallet_handle, CRED_DEF_CACHE, id, &options_json) {
+                Ok(record) => Ok(Some(record)),
+                Err(err) => if err.kind() == IndyErrorKind::WalletItemNotFound {Ok(None)} else {Err(err)}
+            }
+        } else {Ok(None)};
+        let cache = try_cb!(cache, cb);
+
+        if cache.is_some() {
+            let cache = cache.unwrap();
```
jovfer: I suggest to avoid unwrap here, see comment above for more details

libindy/src/commands/non_secrets.rs
```diff
@@ -303,6 +366,256 @@ impl NonSecretsCommandExecutor {
...
+            let min_fresh = options.min_fresh.unwrap_or(-1);
+            if min_fresh >= 0 {
+                let ts = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs() as i32;
```
jovfer: I suggest to avoid unwrap here, see comment above for more details

libindy/src/commands/non_secrets.rs
```diff
@@ -303,6 +366,256 @@ impl NonSecretsCommandExecutor {
...
+    }
+
+    fn _get_cred_def_continue(&self, wallet_handle: WalletHandle, ledger_response: IndyResult<(String, String)>, options: GetCacheOptions, cb_id: i32) {
+        let cb = self.pending_callbacks.borrow_mut().remove(&cb_id).expect("FIXME INVALID STATE");
+
+        let (cred_def_id, cred_def_json) = try_cb!(ledger_response, cb);
+
+        if !options.no_store.unwrap_or(false) {
+            let mut tags = Tags::new();
+            let ts = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs() as i32;
```
jovfer: I suggest to avoid unwrap here, see comment above for more details

libindy/src/commands/non_secrets.rs
```diff
@@ -303,6 +366,256 @@ impl NonSecretsCommandExecutor {
...
+                          wallet_handle: WalletHandle,
+                          options_json: &str) -> IndyResult<()> {
+        trace!("purge_schema_cache >>> wallet_handle: {:?}, options_json: {:?}", wallet_handle, options_json);
+
+        let options = serde_json::from_str::<PurgeOptions>(options_json)
+            .to_indy(IndyErrorKind::InvalidStructure, "Cannot deserialize options")?;
+
+        let max_age = options.max_age.unwrap_or(-1);
+        let query_json = if max_age >= 0 {
+            let ts = SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_secs() as i32;
```
jovfer: I suggest to avoid unwrap here, see comment above for more details

libindy/src/commands/non_secrets.rs
dkulic: In documentation is says: Returns an [`Err`] if `earlier` is later than `self`, and the error contains how far from `self` the time is.
I may return error, if I ignore it would work different than user expects.

PR #1608, opened 2019-04-30
Node.js wrapper - fixing ledger test and upgrading dependencies

farskipper:


wrappers/nodejs/test/ledger.js
```diff
@@ -154,7 +154,7 @@ test('ledger', async function (t) {
   // Auth Rule
   req = await indy.buildGetAuthRuleRequest(trusteeDid, 'NYM', 'ADD', 'role', null, '101')
   res = await indy.submitRequest(pool.handle, req)
-  var defaultConstraint = res['result']['data']['1--ADD--role--*--101']
+  var defaultConstraint = res['result']['data']['ADD--1--role--*--101']
```
Artemkaaas: Revert, please.  `1--ADD--role--*--101` is correct. The last indy-node versions return key in this format.

farskipper: Ok, the old key format has been restored.

PR #1610, opened 2019-05-01
[VCX] - enable update_state_with_message for connections, proofs, creds

glowkey:


vcx/wrappers/java/src/main/java/com/evernym/sdk/vcx/utils/UtilsApi.java
```diff
@@ -158,4 +158,8 @@ public void callback(int commandHandle, int err, String fees) {
         return future;
     }
 
+    public static void vcxMockSetAgencyResponse(int messageIndex) {
+        logger.debug("vcxMockSetAgencyResponse() called");
+        LibVcx.api.vcx_set_next_agency_response(messageIndex);
+    }
 }
```
Artemkaaas: Do we really need this mocking functionality for wrappers? 
Isn't enough testing of state transition within Libvcx library?

glowkey: It's very useful in unit tests for applications. It enables quick and easy state testing within the app without setting up all of the libindy/libvcx infrastructure.

PR #1613, opened 2019-05-08
Send message with ref msg id 

ryMarsh44:
adds ref_msg_id to vcx_connection_send_message for responding to a message

Signed-off-by: Ryan Marsh <ryan.marsh44@gmail.com>


jovfer:
@ryMarsh44 @glowkey it seems like breaking changes from API point of view. But as far as send message is experimental API we can merge it to VCX codebase, isn't it?

glowkey:
Yes, that was our opinion. It is experimental and could be merged.

PR #1614, opened 2019-05-09
credential attribute tagging policy

sklump:

Credential attribute tagging policy as per https://jira.hyperledger.org/browse/IS-1258.

*Done:* indy-sdk implementation, tests in indy-sdk and python wrapper. Java wrapper anoncreds preparation.

*To Do:* Java unit tests and all wrappers but python and rust. I think it would be best if someone more conversant in Java, dot.net, javascript etc. instrumented these to look like the python.

*Quibble:* I dislike the fact that in libindy/src/domain/anoncreds/credential_attr_tag_policy, function `canon()` copies the helper `attr_common_view()` from libindy/src/services/anoncreds/helpers.rs. Is there a place to put this where both can have access? Having domain import from the module that imports from domain appears to cause problems in cargo build, or the build step of cargo test. I'm still pretty green with rust/cargo and could use correction here.


sklump:
I'm going to back out the Java stuff that breaks the build, and tweak the python unit tests so the credential attr tag policy tests don't corrupt the wallet for the other searches. I can probably update early next week (2019-05-12/13).

sklump:
Will resubmit from new branch. Master has drifted enough to warrant rebuild from scratch.

PR #1621, opened 2019-05-11
libvcx use AtomicUsize::new(0)

AxelNennker:


AxelNennker:
Actually.... Please don't merge yet. I want to keep 0 (zero) as INVALID and start/init with one (1).

PR #1630, opened 2019-05-16
IS-1255 Ability of specifying configuration file for Indy CLI

Artemkaaas:


cli/src/command_executor.rs
```diff
@@ -165,6 +166,7 @@ impl CommandContext {
             uint_values: RefCell::new(HashMap::new()),
             string_values: RefCell::new(HashMap::new()),
             plugins: RefCell::new(HashMap::new()),
+            taa_acceptance_mechanism: RefCell::new("Click Agreement".to_owned()),
```
jovfer: should be empty by default

PR #1632, opened 2019-05-17
Actualized iOS readme

Artemkaaas:


wrappers/ios/README.md
```diff
@@ -27,38 +27,35 @@ Add pod to target:
...
+   brew openssl 1.0.2q (can be any fresh version)
    ```
 1. Setup environment variables:
    
    ```
    export PKG_CONFIG_ALLOW_CROSS=1
    export CARGO_INCREMENTAL=1
    ```
-1. Edit script build-libindy-core-ios.sh: set the following variables to fit your environment:
+1. Edit script ci/ios-build.sh: set the following variables to fit your environment:
```
jovfer: please add examples and parameters description

kdenhartog:
Is this work for building an iOS framework locally, or it for publishing it to the spec directory? My understanding when I went through this originally was that it's for publishing and that only the CI/CD pipeline would be doing that. Is that the case for this or does it also support local building?

PR #1636, opened 2019-05-20
Fixed wrongly named variable in Java how-to doc for credential issuance

dunff2k:
Corrected the identifier used to an reference the array of attributes supplied to the issuer function that creates a credential. 

A copy and paste of this how-to code otherwise would fail to compile since the currently used identifier is not declared anywhere else in the code.


jovfer:
Hi @dunff2k , thank you for catching that problem!
To accept the PR it should pass all required check. Please take a look on DCO one it fix it:
https://github.com/apps/dco
https://github.com/hyperledger/indy-sdk/pull/1636/checks?check_run_id=132578141

PR #1640, opened 2019-05-21
 IS-1265: AUTH_RULES transaction builder.

Artemkaaas:


libindy/src/domain/ledger/constants.rs
```diff
@@ -23,6 +23,7 @@ pub const GET_REVOC_REG_DELTA: &str = "117";
 pub const GET_VALIDATOR_INFO: &str = "119";
 pub const AUTH_RULE: &str = "120";
 pub const GET_AUTH_RULE: &str = "121";
+pub const AUTH_RULES: &str = "122";
```
kdenhartog: What's the difference between AUTH_RULE 120 and AUTH_RULES 122?

kdenhartog:
This looks like good work. One question that I'd like to understand better is what the difference between txn '"120" AUTH_RULE' and txn '"122" AUTH_RULES' is. Additionally, what calls would an authorized trustee make in order to enable auth_rules work on ledger? Is the configuration of the AUTH_RULE handled by the config ledger? For example, whether it's enabled and what the requirements are for it.

libindy/src/domain/ledger/constants.rs
Artemkaaas: AUTH_RULE send one specific rule
AUTH_RULES accepts and send vector of auth rules

PR #1642, opened 2019-05-22
IS-1262 Rename TRUST_ANCHOR to ENDORSER

Artemkaaas:


libindy/src/services/ledger/mod.rs
```diff
@@ -67,7 +67,7 @@ impl LedgerService {
                 operation["role"] = Value::String(match r {
                     "STEWARD" => STEWARD,
                     "TRUSTEE" => TRUSTEE,
-                    "TRUST_ANCHOR" => TRUST_ANCHOR,
+                    "ENDORSER" => ENDORSER,
```
jovfer: For first iteration we should support both (old and new variants here)

PR #1657, opened 2019-06-04
ST-535: Get request fee

Artemkaaas:
…version


docs/request-info-options.md
```diff
@@ -0,0 +1,179 @@
+Approaches:
```
jovfer: I suggest to point out the chosen approach in the next PR related to this topic

PR #1661, opened 2019-06-05
credential attribute tagging policy implementation

sklump:
Credential attribute tagging policy. Still needs wrappers & wrapper tests except python, rust.


libindy/src/commands/anoncreds/prover.rs
```diff
@@ -302,6 +323,59 @@ impl ProverCommandExecutor {
...
+        }
+
+        debug!("set_credential_attr_tag_policy <<< res: {:?}", res);
+
+        Ok(res)
+    }
+
+    fn get_credential_attr_tag_policy(&self,
+                                      wallet_handle: WalletHandle,
+                                      cred_def_id: &str) -> IndyResult<String> {
```
jovfer: @sklump how do you think about next option:
* change this function to return `IndyResult<Option<String>>` 
* change API appropriately (allow null pointer in result)
* simplify matches at lines 366 and 403

jovfer: I'm not sure is it better or not, may be you already have considered with option

sklump: I wrote as above because no other prover command implementations return Indy<Option<...>>, and "null" is a JSON string denoting the absence of such a policy.

Please feel free to change it to how you prefer, but note that the anoncreds API call would also have to change to turn the null option into a JSON "null" string in any case.

jovfer: We (me and Stephen) have discussed pro and cons of different approaches here and have decided to keep current implementation.
there is some drawback now: set API consume null pointer to enable default policy, get API return "null" string in similar case. If it will be inconvenient for users set API can be update to accept "null" string as well

PR #1611, opened 2019-05-07
Support Swift version 5.0.

lauravuo-techlab:
Update to Xcode 10.2 and Swift version 5.0

PR #1664, opened 2019-06-06
[iOS Wrapper] Fixed build issues

anastasia-tarasova:
- Replaced `OpenSSL` with `OpenSSL-Universal`
- Removed `CoreBitcoin` pod
- Added logic for Base58 data conversion.
- Added support of Swift classes in `Indy-demoTest`


edynox:
Nice! It would be great to include support for Swift 5 here (https://github.com/hyperledger/indy-sdk/pull/1611).

kdenhartog:
It looks like this is failing to pass because the CI pipeline is still running XCode 9.2. We would need to update to 10.2 in order to compile this Swift 5 code, and have the CI pipeline pass. @esplinr or @stevengubler do either of you know if someone is able to do this work so we can merge this PR? (I'm not familiar enough with CI pipelines to do the work)

PR #1669, opened 2019-06-10
Unidiomatic extern crate

AxelNennker:
Removing some more extern crate

Sorry this comes in chunks. Removing extern crate is not as easy and straight forward as it seems

### DIFF
--- a/libindy/src/services/pool/networker.rs
+++ b/libindy/src/services/pool/networker.rs
@@ -1,5 +1,3 @@
-extern crate zmq;
-
 use std::cell::RefCell;
 use std::collections::{BTreeMap, HashMap, HashSet};
 
@@ -14,8 +12,8 @@ use utils::sequence;
 
 use super::time::Duration;
 
-use self::zmq::PollItem;
-use self::zmq::Socket as ZSocket;
+use super::zmq::PollItem;
+use super::zmq::Socket as ZSocket;
 
 pub trait Networker {
     fn new(active_timeout: i64, conn_limit: usize, preordered_nodes: Vec<String>) -> Self;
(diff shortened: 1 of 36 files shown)


libindy/src/services/pool/networker.rs
```diff
@@ -14,8 +12,8 @@ use utils::sequence;
 
 use super::time::Duration;
 
-use self::zmq::PollItem;
-use self::zmq::Socket as ZSocket;
+use super::zmq::PollItem;
```
jovfer: may it be used without `super::` here?

AxelNennker: Does not compile without the super::
I think this is because this is a module inside a module.

This PR as it is now removes all extern crates from all rust files exept when
- it is a lib.rs
- it is a mod.rs
- the rust file is 'included' in tests using 'path = "../../src/something"

If you find a way without the super:: I would be interested to learn how.

jovfer: ok, got it. I don't understand this compiler behavior but I can see the difference now (nested module in one file), thanks for explanation.

AxelNennker:
It is unclear, to me, why the build failed. Erratic CI problem?

jovfer:
>  Erratic CI problem?

Kind off...
After #1518 has been merged we can see problems because of unisolated build: if some PR is failed on Windows worker it may produce non-cleaned entities like a wallet or pool configuration. Next build of any PR will fail with WallerAlreadyExists

AxelNennker:
> > Erratic CI problem?
> 
> Kind off...
> After #1518 has been merged we can see problems because of unisolated build: if some PR is failed on Windows worker it may produce non-cleaned entities like a wallet or pool configuration. Next build of any PR will fail with WallerAlreadyExists

How about recursively and force deleting $HOME/.indy_client and $TMP/indy_client before and after https://github.com/hyperledger/indy-sdk/blob/master/Jenkinsfile.ci#L91 ? Whatever the values are for $HOME and $TMP on windows. %HOME ?

jovfer:
I think it reasonable until we would have isolated builds. But I'm not sure about the correct windows env variable.

PR #1675, opened 2019-06-16
sha2 and sha3 version from 0.7 to 0.8

AxelNennker:
This PR updates the sha2 and sha3 crates from version 0.7 to 0.8

Also I introduced a new method for a node `get_hash` which computes the node's hash.
This replaces two occurrences where the hash was previously computed. Now the dependency on the hash algorithm is in one place which I consider cleaner.

One difference to the original code is that the output of `info!(encoded)` is now removed in one place.
Not sure who might miss this output.


libindy/src/services/pool/state_proof/node.rs
```diff
@@ -138,10 +138,21 @@ impl rlp::Decodable for Node {
     }
 }
 
-type NodeHash = generic_array::GenericArray<u8, <sha3::Sha3_256 as digest::FixedOutput>::OutputSize>;
+type NodeHash = sha3::digest::generic_array::GenericArray<u8, <sha3::Sha3_256 as sha3::digest::FixedOutput>::OutputSize>;
 pub type TrieDB<'a> = HashMap<NodeHash, &'a Node>;
 
 impl Node {
+    pub fn get_hash(&self) -> NodeHash{
```
jovfer: Should name of the function contains mentioning of RLP here?

libindy/src/services/pool/state_proof/mod.rs
```diff
@@ -533,6 +526,7 @@ fn _verify_proof_signature(signature: &str,
 }
 
 fn _parse_reply_for_proof_value(json_msg: &SJsonValue, data: Option<&str>, parsed_data: &SJsonValue, xtype: &str, sp_key: &[u8]) -> Result<Option<String>, String> {
+    use sha2::digest::{FixedOutput, Input};
```
jovfer: Should this `use` be moved to the top of the file?

AxelNennker: Getting this to compile was astonishingly complicated. 

At least when I upgraded sha3 after getting sha2 to compile after its update.

Maybe it is working now because I removed dependencies for generic-array and digest from Cargo.toml. libindy is not using them directly but only through dependencies.
rustc seems to be of different opinion than me which digest (digest, sha2::digest or sha3::digest) and which generic-array (generic-array, sha::generic-array or sha3::generic-array) must be used. The same for Input and FixedOutput.

That is why I have put the use-statements close to where the methods/types/traits are used.

I am traveling today and can't try it quickly.

ps: have you seen the issue I created regarding dependabot? https://jira.hyperledger.org/browse/IS-1296 That would create PRs for new versions of dependencies automatically. Although, as this example sha2/3 shows the PR would not compile, but at least raise awareness of a new version.
pps: BTW rusqlite made a big leap version-wise.

libindy/src/services/pool/state_proof/node.rs
AxelNennker: Maybe, but there are no other hashes so ...

jovfer:
@AxelNennker I've tried to resolve conflicts between this PR and master (after merging #1676 )

PR #1678, opened 2019-06-18
[IS-1291] Correct behavior of offline ledger commands in CLI.

jovfer:
Previously some ledger commands always required opened pool even send
flag was set to false.


jovfer:
@KitHat please review

PR #1679, opened 2019-06-18
call of environment:: in fewer places

AxelNennker:

New utility method to create pool test_pool_create_poolfile simplifies tests code


AxelNennker:
It seems adding a command that deletes this directory before the Windows tests run will help
`C:\\Users\\jenkins\\AppData\\Local\\Temp\\indy_client`
That is probably %TEMP%\\indy_client

libindy/tests/utils/pool.rs
```diff
@@ -26,15 +26,16 @@ struct PoolConfig {
...
-    let txn_file_path = txn_file_path.map_or(
-        environment::tmp_file_path(format!("{}.txn", pool_name).as_str()),
-        |path| path.to_path_buf());
-
-    if !txn_file_path.parent().unwrap().exists() {
-        fs::DirBuilder::new()
-            .recursive(true)
-            .create(txn_file_path.parent().unwrap()).unwrap();
-    }
+    let txn_file_path= match txn_file_path {
```
jovfer: Seems like behavior is changed now for case of specified path and non-existing parent directory.
But it should not be a problem for this test helpers

PR #1680, opened 2019-06-19
Update error.rs

dwjohnston:
Change the error message to make it clear that the Credential Def is on the user's local wallet, not on the public ledger.


jovfer:
@dwjohnston thank you for contribution! We'd like to merge this PR but the required DCO check is failed. Could you fix it, please?
https://github.com/hyperledger/indy-sdk/pull/1680/checks?check_run_id=151222825
https://github.com/apps/dco

PR #1681, opened 2019-06-19
cargo update 2019 June 19

AxelNennker:

cargo update


jovfer:
@AxelNennker lets see will it pass CI. And then let's try to update all Cargo.lock files across the repo to have consistency.

AxelNennker:
I ran the tests on Ubuntu, so libindy compiles and passes tests. 
`cargo update` only replaces the version after the first non-zero number, so no breaking change if maintainers follow the rules.

AxelNennker:
‘jenkinswindows01’ is offline 
so this is not currently building

jovfer:
yes and Sovrin Foundation is trying to fix problem with windows agent

PR #1697, opened 2019-06-30
[libindy] Add missing transaction type for XFER_PUBLIC

tmarkovski:
This PR adds a missing txn type that makes calls to `ledger.build_auth_rule_request` fail if `constraint` contains `auth_type = "10001"`, because the requested txn_type cannot be resolved in the list of known types.
I'm not sure if wrappers are intended to call this function directly, if not, I can abandon this PR.
I came across the bug when writing unit tests for testing fee payments when making XFER_PUBLIC transaction type. The proposed fix addressed my issue and I was able to properly update auth rules and set fees for txn_type = 10001.

### DIFF
--- a/libindy/src/domain/ledger/constants.rs
+++ b/libindy/src/domain/ledger/constants.rs
@@ -25,11 +25,12 @@ pub const AUTH_RULE: &str = "120";
 pub const GET_AUTH_RULE: &str = "121";
 pub const AUTH_RULES: &str = "122";
 pub const GET_DDO: &str = "120";//TODO change number
+pub const XFER_PUBLIC: &str = "10001";
 
-pub const REQUESTS: [&str; 25] = [NODE, NYM, GET_TXN, ATTRIB, SCHEMA, CRED_DEF, GET_ATTR, GET_NYM, GET_SCHEMA,
+pub const REQUESTS: [&str; 26] = [NODE, NYM, GET_TXN, ATTRIB, SCHEMA, CRED_DEF, GET_ATTR, GET_NYM, GET_SCHEMA,
     GET_CRED_DEF, POOL_UPGRADE, POOL_RESTART, POOL_CONFIG, REVOC_REG_DEF, REVOC_REG_ENTRY, GET_REVOC_REG_DEF,
     GET_REVOC_REG, GET_REVOC_REG_DELTA, GET_VALIDATOR_INFO, AUTH_RULE, GET_DDO, TXN_AUTHR_AGRMT, TXN_AUTHR_AGRMT_AML,
-    GET_TXN_AUTHR_AGRMT, GET_TXN_AUTHR_AGRMT_AML];
+    GET_TXN_AUTHR_AGRMT, GET_TXN_AUTHR_AGRMT_AML, XFER_PUBLIC];
 
 pub const TRUSTEE: &str = "0";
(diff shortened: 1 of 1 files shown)


libindy/src/domain/ledger/constants.rs
```diff
@@ -25,11 +25,12 @@ pub const AUTH_RULE: &str = "120";
...
 pub const AUTH_RULES: &str = "122";
 pub const GET_DDO: &str = "120";//TODO change number
+pub const XFER_PUBLIC: &str = "10001";
 
-pub const REQUESTS: [&str; 25] = [NODE, NYM, GET_TXN, ATTRIB, SCHEMA, CRED_DEF, GET_ATTR, GET_NYM, GET_SCHEMA,
+pub const REQUESTS: [&str; 26] = [NODE, NYM, GET_TXN, ATTRIB, SCHEMA, CRED_DEF, GET_ATTR, GET_NYM, GET_SCHEMA,
     GET_CRED_DEF, POOL_UPGRADE, POOL_RESTART, POOL_CONFIG, REVOC_REG_DEF, REVOC_REG_ENTRY, GET_REVOC_REG_DEF,
     GET_REVOC_REG, GET_REVOC_REG_DELTA, GET_VALIDATOR_INFO, AUTH_RULE, GET_DDO, TXN_AUTHR_AGRMT, TXN_AUTHR_AGRMT_AML,
-    GET_TXN_AUTHR_AGRMT, GET_TXN_AUTHR_AGRMT_AML];
+    GET_TXN_AUTHR_AGRMT, GET_TXN_AUTHR_AGRMT_AML, XFER_PUBLIC];
```
Artemkaaas: Libindy doesn't know anything about `XFER_PUBLIC` transaction.
XFER relates to payment plugin and mustn't be listed here.

tmarkovski: I thought this would be the case. How would we solve the problem then whereby I call to `ledger.build_auth_rule_request` with `txn_type = "10001"` fails on this line https://github.com/hyperledger/indy-sdk/blob/612d14a1291e29eba4070b4af91fa8992dcc7221/libindy/src/services/ledger/mod.rs#L397 because it's unable to resolve the code. It seems that the SDK doesn't allow downstream plugins to introduce custom txn types, since it only allows the ones defined here.

tmarkovski: @Artemkaaas It seems I had an older version or libindy or something else was going on. I just tried the master version to run a rule for txn_type = 10001 and it worked fine.

```
08:50:00 Trace src/api/ledger.rs:1901 | indy_build_auth_rule_request: entities >>> submitter_did: "V4SGRU86Z58d6TV7PBUe6f", txn_type: "10001", action: "ADD", field: "*", old_value: Some("*"), new_value: Some("*"), constraint: "{\"constraint_id\":\"ROLE\",\"metadata\":{\"fees\":\"fees_for_xfer\"},\"need_to_be_owner\":false,\"role\":\"*\",\"sig_count\":0}" 
08:50:00 Trace src/api/ledger.rs:1922 | prepare_result: >>> Ok(()) 
08:50:00 Trace src/api/ledger.rs:1924 | indy_build_auth_rule_request: <<< res: Success 
```

Closing this PR. Sorry about the confusion.

PR #1698, opened 2019-07-03
Implement IS-1308

mikelodder7:


jovfer:
@mattraffel @nage thank you for the reviews. BTW this PR is **break API** between libindy and any payment plugins and it should be merged only right before major implementations will be ready to update.

So I have to revert this merge to unblock test process right now.

@mikelodder7 please raise one more PR to revert my revert.

PR #1701, opened 2019-07-03
getting-started container hangs

JanLin:
There is an issue with asyncio which makes notebook to hang. There may be an incompatability with later versions of tornado which causes the problem. By adding new line "tornado==4.5.3" I was able to fix the problem. There may be a later version that will work but did not explore which version causes the break. Below is a link touching on issue with hanging in general.

https://github.com/dask/distributed/issues/2414

### DIFF
--- a/docs/getting-started/getting-started.dockerfile
+++ b/docs/getting-started/getting-started.dockerfile
@@ -20,6 +20,8 @@ RUN pip3 install -U \
 	pip \
 	setuptools \
 	jupyter \
+	tornado==4.5.3 \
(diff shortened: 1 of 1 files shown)


JanLin:
@JanLin signoff

jovfer:
@JanLin thank you for contribution!
As you've noticed the DCO check is failed. you can find a guide how to fix it at the next link https://github.com/hyperledger/indy-sdk/pull/1701/checks?check_run_id=161434057

JanLin:
Newer version has fixed problem. Closed CR.

PR #1704, opened 2019-07-08
IS-1307: Indy CLI persist command history between sessions

Artemkaaas:

### DIFF
--- a/cli/src/utils/history.rs
+++ b/cli/src/utils/history.rs
@@ -5,6 +5,7 @@ use utils::file::{read_lines_from_file, write_file};
 
 
 const HISTORY_SIZE: usize = 100;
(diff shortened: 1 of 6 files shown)


cli/src/utils/history.rs
```diff
@@ -0,0 +1,37 @@
...
+    }
+    Ok(())
+}
+
+pub fn persist<T>(reader: &Reader<T>) -> Result<(), String> where T: Terminal {
+    let content =
+        reader
+            .history()
+            .filter(|record|
+                !SECRET_DATA.iter().any(|secret_word| record.contains(&format!("{}=", secret_word)))
```
jovfer: I suggest to include `=` suffix in the const and simplify this line

PR #1707, opened 2019-07-08
IS-1266: error message for unexpected length of did

Artemkaaas:
…nverted length.

### DIFF
--- a/libindy/src/services/crypto/mod.rs
+++ b/libindy/src/services/crypto/mod.rs
@@ -412,7 +412,8 @@ impl CryptoService {
 
         if did.len() != 16 && did.len() != 32 {
             return Err(err_msg(IndyErrorKind::InvalidStructure,
-                               format!("Trying to use `did` with unexpected length: {}. The length of base58-decoded `did` must be either 16 or 32 bytes.", did.len())));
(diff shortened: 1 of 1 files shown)


libindy/src/services/crypto/mod.rs
```diff
@@ -409,7 +411,8 @@ impl CryptoService {
         let did = base58::decode(did)?;
 
         if did.len() != 16 && did.len() != 32 {
-            return Err(err_msg(IndyErrorKind::InvalidStructure, format!("Trying to use did with unexpected len: {}", did.len())));
+            return Err(err_msg(IndyErrorKind::InvalidStructure,
+                               format!("Trying to use `did` with unexpected length: {}. The length of base58-decoded `did` must be either 16 or 32 bytes.", did.len())));
```
dhh1128: I think there's a problem with the wording here. What you mean, I think, is that the number encoded as base58 must be, before its encoding, a 16- or 32-byte number. But what this string says is that the length of the number after encoding must be 16 or 32 bytes. You could reword as follows: "Trying to use DID with unexpected length: {}. The 16- or 32-byte number upon which a DID is based should be 22/23 or 44/45  bytes when encoded as base58."

PR #1710, opened 2019-07-10
Function to generate nonce for proof request

Artemkaaas:


wrappers/python/indy/anoncreds.py
```diff
@@ -1007,7 +1009,7 @@
         {
             "name": string,
             "version": string,
-            "nonce": string,
+            "nonce": string, - a big number represented as a string (use `generate_nonce` funciton to generate 80-bit number)
```
mikelodder7: funciton -> function

wrappers/python/indy/anoncreds.py
```diff
@@ -1108,7 +1110,7 @@
...
```
mikelodder7: See above

wrappers/python/indy/anoncreds.py
```diff
@@ -1254,7 +1256,7 @@
...
```
mikelodder7: See above

wrappers/python/indy/anoncreds.py
```diff
@@ -1007,7 +1009,7 @@
         {
             "name": string,
             "version": string,
```
mikelodder7: Is there an example of version somewhere

Artemkaaas:
> I'd like to us increase the bit size of the nonces to 128 bits. Would that break existing tests and APIs?

It should not affect

wrappers/python/indy/anoncreds.py
Artemkaaas: This `version` field is for apps. It can be any for now. `Proof Request` (structure) versioning isn't implemented yet.

PR #1712, opened 2019-07-11
IS-1267: Added an additional validation for CLI `ledger nym` command

Artemkaaas:


cli/src/commands/ledger.rs
```diff
@@ -108,6 +108,28 @@ pub mod nym_command {
         let extra = get_opt_str_param("extra", params).map_err(error_err!())?;
         let send = get_opt_bool_param("send", params).map_err(error_err!())?.unwrap_or(SEND_REQUEST);
 
+        if submitter_did == target_did {
```
jovfer: I suggest to remove this first check and keep only the flow below

PR #1721, opened 2019-07-12
Add support for building nodejs wrapper on Windows

SonofNun15:
The previous binding.gyp file in the nodejs wrapper only supported Unix systems. Added custom `link_settings` for Windows machines. Assumes that LD_LIBRARY_PATH is set to the folder that contains indy.dll.lib


SonofNun15:
I plan to update a few of the nodejs ReadMes doing that shortly here.

SonofNun15:
ReadMes updated.

SonofNun15:
If it is not present it just puts in an empty folder which doesn't cause any issues with the library paths. It's hard to assume this would be in any default location, but I suppose we could assume something like `C:\windows\system32`

I do want to verify that `LD_LIBRARY_PATH` can have multiple paths separated by `;` for Windows.

samples/nodejs/README.md
```diff
@@ -17,6 +17,8 @@ You must have `libindy.dylib` at `/usr/local/lib/libindy.dylib` before running n
 ### On Windows
 [Install Libindy](https://github.com/hyperledger/indy-sdk#windows)
 
+Set environment variable `LD_LIBRARY_PATH` to include a path to the lib folder that contains indy.dll.lib
+
```
mattraffel: should this be `contains indy.dll` ?

SonofNun15: It must contain both `indy.dll.lib` AND `indy.dll`

SonofNun15: ...Which the Windows builds include...

SonofNun15:
Verified that paths in `LD_LIBRARY_PATH` separated by `;` are included individually. So putting multiple paths in `LD_LIBRARY_PATH` works fine.

samples/nodejs/README.md
SonofNun15: I mentioned `indy.dll.lib` specifically because this is the file that is explicitly included by node-gyp as it has the static method definitions

wrappers/nodejs/binding.gyp
```diff
@@ -11,11 +11,26 @@
         "src/indy.cc"
       ],
       "link_settings": {
-        "libraries": [
-          "-L<(module_root_dir)",
-          "<!(node -e \"console.log((process.env.LD_LIBRARY_PATH || '').split(':').map(a => '-L' + a.trim()).filter(a => a != '-L').join(' '))\")",
-          "-lindy"
-        ]
+        "conditions": [
+          [ 'OS=="win"',
```
jovfer: I'm not familiar with GYP, could someone else double-check this changes?
btw I thought that it should be like
```
condtitions: [
  [ OS == win, { library_dirs, libraries } ]
  [ OS != win, { libraries } ]
]
```

SonofNun15: I have tested this on a MacBook and it works as intended. The section without the criteria is the default case. I've also seen examples online that use default criteria.

jovfer: Thanks for clarification

PR #1726, opened 2019-07-15
Rusqlite version 0.19

AxelNennker:
updating Rust to 1.36.0 leads to Indy not being able to build.

Upgrading rusqlite and deleting a dependency on libsqlite3-sys solves the problem.
libindy tests succeed


jovfer:
@AxelNennker thanks for catching that and the fix! I've tried to migrate to newer Rust but have no time to migrate to new rusqlite API
@mattraffel it may be interesting for you

PR #1732, opened 2019-07-17
 ST-604 -- Improving state proof verification

KitHat:


libindy/src/services/pool/types.rs
```diff
@@ -428,7 +428,18 @@ pub enum KeyValuesInSP {
...
+    pub verification_type: KeyValueSimpleDataVerificationType
+}
+
+/**
+ Options of common state proof check process
+*/
+#[derive(Serialize, Deserialize, Debug, Eq, PartialEq)]
+pub enum KeyValueSimpleDataVerificationType {
+    Simple(),
+    NumericalSuffixAscendingNoGaps(Option<u64>, Option<u64>, String)
```
jovfer: Have you decided to mix order and continually characteristics

libindy/src/services/pool/state_proof/node.rs
```diff
@@ -165,6 +165,102 @@ impl Node {
...
+            }
+            Node::Hash(ref hash) => {
+                trace!("Node::_get_all_values in Hash");
+                let hash = NodeHash::from_slice(hash.as_slice());
+                if let Some(ref next) = db.get(hash) {
+                    trace!("found value in db");
+                    next._get_all_values(db)
+                } else {
+                    trace!("no value");
+                    Ok(vec![])
```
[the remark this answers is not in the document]
KitHat: Seems like no. It can be empty here -- at least because pool does not return full tree.
But I want to think it over once again.

libindy/src/services/pool/state_proof/mod.rs
```diff
@@ -477,6 +493,95 @@ fn _verify_proof(proofs_rlp: &[u8], root_hash: &[u8], key: &[u8], expected_value
...
+                let no = key.replacen(prefix, "", 1).parse::<u64>();
+                no.ok().map(|a| (a, (key, value)))
+        }).collect();
+        if !vals_for_sort_check.iter().all(|a| a.is_some()) {
+            error!("Some values in state proof are not correlating with state proof rule, aborting.");
+            return false;
+        }
+        let mut vals_for_sort: Vec<(u64, (String, String))> = vals_for_sort_check.into_iter().flat_map(|a| a).collect();
+        // Sort by numerical suffixes in ascending order
+        vals_for_sort.sort_by(|a, b| {
```
jovfer: Can be simplified to `vals_for_sort.sort_by_key(|&(k, _)| k);`

libindy/src/services/pool/state_proof/mod.rs
```diff
@@ -477,6 +493,95 @@ fn _verify_proof(proofs_rlp: &[u8], root_hash: &[u8], key: &[u8], expected_value
...
+            let (a_seq_no, _) = a;
+            let (b_seq_no, _) = b;
+            a_seq_no.cmp(b_seq_no)
+        });
+        trace!("Sorted trie values: {:?}", vals_for_sort);
+        // Shift on the left side by from
+        let vals_with_from = if let Some(from_seqno) = from {
+            if let Some((idx, _)) = vals_for_sort.iter().enumerate().find(|(_, (seq_no, _))| *seq_no as u64 >= from_seqno) {
```
jovfer: it's sorted, can we just use built-in binary_search?

libindy/src/services/pool/state_proof/mod.rs
```diff
@@ -477,6 +493,95 @@ fn _verify_proof(proofs_rlp: &[u8], root_hash: &[u8], key: &[u8], expected_value
...
+                vals_for_sort[idx..].to_vec()
+            } else {
+                vec![]
+            }
+        } else {
+            vals_for_sort
+        };
+        // Verification
+        // Check that all values from response match the trie
+        trace!("Got values from trie: {:?}", vals_with_from);
```
jovfer: Can it be simplified to just 2 arrays comparison?

libindy/src/services/pool/state_proof/node.rs
```diff
@@ -237,6 +238,131 @@ impl Node {
...
+                let mut res = vec![];
+                for nibble in 0..16 {
+                    let next = &node.nodes[nibble];
+                    if let Some(next) = next {
+                        trace!("Checking nibble {}", nibble);
+                        let mut sub_res: Vec<(Vec<u8>, String)> = next._get_all_values(db)?;
+                        trace!("Got some values: {:?}", sub_res);
+                        sub_res = sub_res.into_iter().map(|(mut key, val)| {
+                            let mut key_new: Vec<u8> = vec![nibble as u8];
+                            key_new.append(&mut key);
```
jovfer: If I get it correct the key in final result will be restored in multiply steps.
If it correct statement about current implementation I suggest to track current path from the root and construct key for returned value at the moment of it detection

libindy/src/services/pool/state_proof/mod.rs
KitHat: simplified to slices comparison

PR #1739, opened 2019-07-18
Generator parameter

AxelNennker:
The parameter `generator` was never used but always None and thus the default value.

request handlers should specify the Generator that should be used.

This PR does not change which Generator is used but makes it clearer that the default generator is always used.


AxelNennker:
```
sh: 1: Cannot fork
```
I guess the jenskins server instance ran out of resources while building...

libindy/src/services/pool/pool.rs
```diff
@@ -308,7 +309,8 @@ impl<T: Networker, R: RequestHandler<T>> PoolSM<T, R> {
                     PoolEvent::CatchupRestart(merkle_tree) => {
                         if let Ok((nodes, remotes)) = _get_nodes_and_remotes(&merkle_tree) {
                             state.networker.borrow_mut().process_event(Some(NetworkerEvent::NodesStateUpdated(remotes)));
-                            state.request_handler = R::new(state.networker.clone(), _get_f(nodes.len()), &vec![], &nodes, None, &pool_name, timeout, extended_timeout);
+                            let generator : Generator = Generator::from_bytes(&DEFAULT_GENERATOR.from_base58().unwrap()).unwrap();
```
jovfer: I don't like the idea to duplicate this one-line logic all the time.
I suggest to pick one of the options:
1) return back optional parameter but make name of the parameter be more concrete that in majority of cases (all cases for now) default one will be used. Like "specififc_generator"  - optional generator otherwise default one will be used
2) eliminate this parameter completely and just setup default generator in `RequestSM::new`

AxelNennker: Do you know why the constant has the value it has? Where does that come from?

I have changed the code to use random generators and the tests still pass. So what is the point of this constant. Also it seems the generators not even have to be the same. Strange.

Is something wrong with the signature validation?

jovfer: It should be the same as the generator on the Ledger side. Most probably tests are passed due to fallback path of read requests. To double-check it manually you can run any ledger read request and look for state proof validation status in trace log

AxelNennker: Just pushed code implementing option 2

AxelNennker:
@mikelodder7 Do you know where this constant comes from?
https://github.com/hyperledger/indy-sdk/blob/master/libindy/src/services/pool/request_handler.rs#L66
 
@jovfer How about skipping the base58 decoding for each instance of request handler and using the bytes directly?

```
+    fn test_default_generator() {
+        use ursa::pair::PointG2;
+        let generator_bytes = DEFAULT_GENERATOR.from_base58().unwrap();
+        let bytes: [u8; 128] = [22, 203, 110, 31, 27, 120, 3, 243, 10, 178, 198, 97, 25, 111, 225, 153, 175, 23, 216, 237, 25, 61, 152, 163, 208, 250, 23, 99, 138, 58, 27, 131, 29, 245, 65, 145, 143, 14, 90, 205, 5, 118, 153, 139, 253, 184, 57, 49, 131, 73, 184, 172, 187, 65, 6, 254, 147, 230, 163, 211, 90, 63, 0, 129, 7, 226, 196, 167, 201, 165, 4, 159, 44, 201, 249, 215, 206, 213, 4, 159, 67, 54, 246, 120, 67, 197, 220, 50, 173, 148, 14, 57, 126, 37, 45, 247, 23, 106, 143, 118, 253, 21, 213, 54, 188, 141, 41, 74, 199, 4, 15, 108, 200, 213, 96, 218, 209, 61, 232, 140, 61, 250, 114, 96, 236, 54, 52, 82];
+        assert_eq!(bytes.len(), PointG2::BYTES_REPR_SIZE);
+        assert!(generator_bytes.eq(&bytes.to_vec()));
+    }
```

jovfer:
@AxelNennker I don't think that this optimization worth applying - in base58 it's easier to compare with Node codebase.

AxelNennker:
Good, so the constant in indy-sdk and indy-plenum MUST match?!
https://github.com/hyperledger/indy-plenum/blob/master/crypto/bls/indy_crypto/bls_crypto_indy_crypto.py#L19
https://github.com/hyperledger/indy-sdk/blob/master/libindy/src/services/pool/request_handler.rs#L66

Just for my curiosity?
- Why this value? What are its properties?
- What happens if the generator points are different?

I think this needs documentation e.g. if somebody implements Plenum in nodejs

jovfer:
> What happens if the generator points are different?
BLS verification should fail in this case

lovesh:
> Why this value? What are its properties?

The generator should be considered a protocol level constant hence it must match in all components using the protocol (BLS sig).
The generator has been chosen to be the same as the value AMCL has chosen. But it could have been any point on the curve in the largest prime-order order subgroup. It would have been better to hash some string like "Protocol-Ver-1.1" or something to get a curve point and treat it as a generator. Hence you can change underlying library without worrying about whether you will find the same value as generator in the new library. 
But we should still copy over the generator coordinates in some config file.

AxelNennker:
Could you please give link to this AMCL constant? 

Maybe it can be exported from the [amcl Rust crate](https://crates.io/crates/miracl_amcl) (re-exported from [ursa](https://github.com/hyperledger/ursa))?

https://jira.hyperledger.org/browse/IS-1333

mikelodder7:
I added some notes to the jira ticket

lovesh:
> Could you please give link to this AMCL constant?

Since the curve we use in Indy is BN254 atm. [here](https://github.com/miracl/amcl/blob/master/version3/rust/src/roms/rom_bn254_64.rs#L61) it is, the constants `CURVE_GX` and `CURVE_GY` are the coordinates. 

> Maybe it can be exported from the amcl Rust crate (re-exported from ursa)?

My point is that it should not be exported from AMCL but declared in Indy in some config file, better in config ledger.

mikelodder7:
I put the constant in the JIRA ticket

AxelNennker:
I would like to know how CURVE_GX and CURVE_GY are transformed into the used generator constant.

Who is following up with a ticket to put the constant on the config ledger and plenum and indy reading it from there?

lovesh:
> I would like to know how CURVE_GX and CURVE_GY are transformed into the used generator constant.

They are converted to hex during JSON serialization for G1 point, the base58 must be constructed from that.
I see that you were talking about group G2, so it should be CURVE_PXA, CURVE_PXB, CURVE_PYA, CURVE_PYB. But my point remains the same, not to rely on AMCL providing those.  

> to put the constant on the config ledger and plenum

Don't know if others agree with that or not. We don't have to do it immediately but when we migrate to the new curve (BLS381) we should. Thoughts, @jovfer @mikelodder7 @brentzundel

PR #1753, opened 2019-07-22
correct facts about aad

AxelNennker:

Avoid potential double assignment of Option and remove one mut.
Also some rustfmt


AxelNennker:
Something went wrong

PR #1755, opened 2019-07-22
use vec! to initialize vec with None values

AxelNennker:

### DIFF
--- a/libindy/src/services/pool/networker.rs
+++ b/libindy/src/services/pool/networker.rs
@@ -190,9 +190,7 @@ impl PoolConnection {
             });
         }
 
-        let mut sockets: Vec<Option<ZSocket>> = Vec::new();
-
-        for _ in 0..nodes.len() { sockets.push(None); }
+        let mut sockets: Vec<Option<ZSocket>> = vec![None; nodes.len()];
 
         PoolConnection {
             nodes,


libindy/src/services/pool/networker.rs
```diff
@@ -190,9 +190,7 @@ impl PoolConnection {
             });
         }
 
-        let mut sockets: Vec<Option<ZSocket>> = Vec::new();
-
-        for _ in 0..nodes.len() { sockets.push(None); }
+        let mut sockets: Vec<Option<ZSocket>> = vec![None; nodes.len()];
```
jovfer: it fails compilation on Jenkins https://build.sovrin.org/job/indy-sdk/job/indy-sdk-ci/job/PR-1755/1/execution/node/52/log/

AxelNennker: Sorry, it seems I forgot to follow my usual procedure I do before sending PRs. I'll close this.

jovfer: np. suggested changes was so obvious for me that I was totally surprised to have a compilation error when I wrote this code about year ago. And when you submitted a PR yesterday I was believing that behavior of the rust compiler had been fixed in one of the newer version

AxelNennker: No, it is not fixed.
https://users.rust-lang.org/t/how-to-initialize-vec-option-t-with-none/30580/12

PR #1766, opened 2019-07-26
PoolHandle and CommandHandle typesafety

AxelNennker:
Another go at trying to replace all i32 that are handles with better types.
This PR does not pull through with this because some hashmaps are most of the times used with CommandHandle as the key but sometimes with PoolHandle (call to open pool by name).
This programming shortcut prevents "clean" programming.

You can see the problem when you uncomment the strong types for PoolHandle and CommandHandle in api/mod.rs and comment out the current declarations.


AxelNennker:
This PR replaces many i32 types with PoolHandle and CommandHandle. Not there for full type-safety but an improvement, I think. So, I like this to be merged, because merge-conflicts pop up which are unnecessary work.

PR #1773, opened 2019-07-31
Fix inner types in libindy

KitHat:


AxelNennker:
Would you please consider using a type alias to make changing the type easier in the future and to enable type safety like WalletHandle?
https://github.com/hyperledger/indy-sdk/blob/master/libindy/src/api/mod.rs#L25

PR #1775, opened 2019-07-31
Fix mac installer formula failures

theoturner:
Homebrew install can throw an error if some formulas already are installed. This fixes that issue: installs formulas that are not present and upgrades those that are out of date without throwing errors.

Signed-off-by: Theo Turner <theo_turner@outlook.com>


esplinr:
Related issue:
https://jira.hyperledger.org/browse/IS-1340

PR #1777, opened 2019-08-02
Updated CLI debian package to depends on libncursesw5-dev  library

Artemkaaas:


cli/README.md
```diff
@@ -63,3 +63,12 @@ By default, this file creates in current folder and has the following name:
     ```
     did import <path to the file created on first step>
     ```
+
+
+### Notes
+Indy-CLI depends on `term` rust library that has a system dependency on terminfo database. 
+That why CLI Debian package additionally installs `libncursesw5-dev` library.
```
AxelNennker: s/That why/That is why/

PR #1782, opened 2019-08-02
Cleanup Tests Directories

AxelNennker:
cleanup directories created during tests


AxelNennker:
I think merging the libsodium PR will fix this build failure.

PR #1783, opened 2019-08-03
A struct Digest is confusing and unneeded

AxelNennker:

Indy defined its own struct 'Digest'. This is confusing because there is a RustCrypto trait named Digest as well.
https://github.com/RustCrypto/traits/blob/master/digest/src/digest.rs
Digest and other common cryptographic traits are provides by the crate digest
https://docs.rs/digest/0.8.1/digest/


libindy/src/utils/crypto/hash/openssl.rs
```diff
@@ -2,56 +2,42 @@ extern crate openssl;
 
 use errors::prelude::*;
 use self::openssl::error::ErrorStack;
-use self::openssl::hash::{DigestBytes, hash as openssl_hash, Hasher, MessageDigest};
+use self::openssl::hash::{Hasher, MessageDigest};
 
 pub const HASHBYTES: usize = 32;
 
+pub const EMPTY_HASH_BYTES : [u8; 32] = [227, 176, 196, 66, 152, 252, 28, 20, 154, 251, 244, 200, 153, 111, 185, 36, 39, 174, 65, 228, 100, 155, 147, 76, 164, 149, 153, 27, 120, 82, 184, 85];
```
KitHat: Please, could you make this as a `lazy_static!` block, so it would not look like a magic number?

libindy/src/services/pool/events.rs
```diff
@@ -199,8 +201,13 @@ impl Into<Option<RequestEvent>> for PoolEvent {
                 _parse_msg(&msg).map(|parsed|
                     match parsed {
                         //TODO change mapping for CatchupReq. May be return None
-                        //TODO: REMOVE UNWRAP!!!!!
-                        Message::CatchupReq(_) => RequestEvent::CatchupReq(MerkleTree::from_vec(Vec::new()).unwrap(), 0, vec![]),
+                        Message::CatchupReq(_) => RequestEvent::CatchupReq(
+                            MerkleTree {
```
KitHat: Maybe here would be better to use a `Default` trait and implement it for Merkle tree?

libindy/src/utils/crypto/hash/openssl.rs
AxelNennker: How about adding a comment that these bytes are the same bytes as `openssl_hash(MessageDigest::sha256(), &[])` ?
I think lazy_static are for when you can't use const.

libindy/src/services/pool/events.rs
AxelNennker: There is already a Default impl for MerkleTree in request_handler.rs which is different to that would be needed here. It is used 20 times in the code. It has the empty vector as the hash while this impl has the hash of the empty vector as the hash.

libindy/src/services/pool/events.rs
```diff
@@ -203,8 +205,13 @@ impl Into<Option<RequestEvent>> for PoolEvent {
...
+                                root: Tree::empty(EMPTY_HASH_BYTES.to_vec()),
```
KitHat: Shouldn't we use `MerkleTree::default()` here as well?

AxelNennker: The enum type Tree is different to the type MerkleTree. 
Tree::empty is only used once, so I think implementing default is not needed.

PR #1784, opened 2019-08-04
IS-1334 -- Updated documentation

KitHat:


docs/how-tos/how-to-add-a-new-API-call.md
```diff
@@ -0,0 +1,118 @@
+# How to add a new API call (code walkthrough)
+
+In this guide you will see how to add a new call to Libindy. As an example we will take `indy_create_and_store_my_did` call.
+
+### API Layer
+
+These changes should be made in directory [`libindy/src/api`](libindy/src/api) in file `did.rs`. For your function you should choose which file it will logically fit or create a new one (first option is preferred)
+
+#### Call definition:
+#### Call definition:
```
VladimirWork: Redundant "Call definition"?

docs/how-tos/how-to-add-a-new-API-call.md
```diff
@@ -0,0 +1,118 @@
...
+}
+```
+What you can modify here is function name and params. 
+
+There is only one strict rule for the function name -- it should have prefix `indy_` to avoid collisions with other libraries in system.
+
+Rules for params:
+1) First parameter is almost always `command_handle` -- you will need this param if your call will work asynchronously -- you will need it most of the times.
+2) After the command handle list all of the params that you need to pass. Please, don't use `Option` on the types other than `extern fn`.
+3) The last parameter is almost always `cb` -- it is a callback. His first param should be `command_handle_` (for your app to identify which call this callback came for), then `err` for the error code and then all the values that you call is going to return.
```
AxelNennker: s/His first param/Its first param/

docs/how-tos/how-to-add-a-new-API-call.md
```diff
@@ -0,0 +1,118 @@
...
+4) Please, use the types that were defined in `libindy/src/api/mod.rs` such as `IndyHandle`, `WalletHandle` and so on for the types of your parameters.
```
AxelNennker: I am not a fan of IndyHandle because that one will stay an alias to i32. The other types, like WalletHandle, are in the process of being moved to type safety like WalletHandle.
So I would write 'such as WalletHandle, PoolHandle and so on'

docs/how-tos/how-to-add-a-new-API-call.md
```diff
@@ -0,0 +1,118 @@
...
+```rust
+trace!("indy_create_and_store_my_did: >>> wallet_handle: {:?}, did_json: {:?}", wallet_handle, did_info);
+
+check_useful_json!(did_info, ErrorCode::CommonInvalidParam3, MyDidInfo); // redefine to MyDidInfo if valid
+check_useful_c_callback!(cb, ErrorCode::CommonInvalidParam4);
+
+trace!("indy_create_and_store_my_did: entities >>> wallet_handle: {:?}, did_json: {:?}", wallet_handle, secret!(&did_info));
+```
+
+For most C-types that need to be converted to Rust types Libindy has a set of macroses that will check and convert it if everything is good or throw specified error code otherwise. You can see these macroses in [`libindy\src\utils\ctypes.rs`](libindy\src\utils\ctypes.rs)
```
AxelNennker: s/macroses/macros/g

docs/how-tos/how-to-add-a-new-API-call.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Passing command to Command Thread
+
+#### Adding new command
+
+For that you should go to [`libindy/src/commands`](libindy/src/commands) directory and choose a module that fits your call. In most of the times it matches by the name to the module from the previous section. In our case it is `did.rs`.
```
AxelNennker: s/In most of the times it matches by the name/Most of the times it matches the name/

PR #1785, opened 2019-08-05
IS-1248: Allow user to modify cred_def

Artemkaaas:


Artemkaaas:
Case: Private keys for Cred Def have been compromised.
Here is implemented a similar approach as for Verkey rotation (need to call the combination of 3 functions). 

Issuing of new a credential even doesn't require writing a transaction to the Ledger (cred def without revocation support). So, Issuer even doesn't know about which credentials have been issued.

brentzundel:
Is there documentation being written about this feature where we could add the line, "**Warning, rotating the cred_def keys will result in making all credentials issued under the previous cred_def keys unverifiable**"

dhh1128:
@brentzundel Couldn't verification just compare the date of rotation against the date of credential signing? I am not surprised that we break our current verification algorithm, but I'm surprised if this is a deeper problem that can't be fixed...

brentzundel:
I suppose that when it is possible to request a version of a cred_def from a particular point in the past, then my concerns may be moot.

mikelodder7:
> @brentzundel Couldn't verification just compare the date of rotation against the date of credential signing? I am not surprised that we break our current verification algorithm, but I'm surprised if this is a deeper problem that can't be fixed...

If the issuer includes the signing date as an attribute then this should be easy or they would have to include that metadata themselves. The holder would need a way to communicate the cred_def version to the verifier. Is there a way to query the ledger as of a specific datetime? Or would the holder need to know the transaction id when the cred_def was created so as to state "this credential was signed by the key at transaction X."

PR #1789, opened 2019-08-06
IS-1203: Remove package-lock.json file from vcx nodejs wrapper

Artemkaaas:


AxelNennker:
Maybe preventing package-lock.json being created is a good idea too.
https://www.codementor.io/johnkennedy/get-rid-of-that-npm-package-lock-json-e0bj7ai42

Do the reasons to remove package-lock.json apply the Cargo.lock as well?
https://jira.hyperledger.org/browse/IS-1295

jovfer:
package-lock.json here has a same value as Cargo.lock for Rust wrapper, not libindy or libvcx itself. So we should skip all language-specific lock files except libraries as libindy and libvcx are distributed in binary format.

PR #1793, opened 2019-08-07
use meaningful parameters

AxelNennker:

Check that the parameter is a valid JWE JSON in the API layer. Use strong types in Rust that is not API


wrappers/java/src/test/java/org/hyperledger/indy/sdk/crypto/PackUnpackMessageTest.java
```diff
@@ -77,14 +78,20 @@ public void testPackMessageErrorsInvalidReceivers() throws Exception {
...
-	public void testUnpackMessageErrorsWithInvalidPackedMessageStructure() throws Exception {
+	public void testUnpackMessageErrorsWithInvalidPackedMessageStructure() {
 		String packedMessage = "jibberish";
 
 		thrown.expect(ExecutionException.class);
 		thrown.expectCause(isA(InvalidStructureException.class));
 
-		Crypto.unpackMessage(wallet, packedMessage.getBytes()).get();
-
+		try {
```
jovfer: Should this test just expect InvalidParameterException instead of invalid structure few lines above and keep old test body?

libindy/src/api/crypto.rs
```diff
@@ -748,8 +749,14 @@ pub extern fn indy_unpack_message(
         jwe_len
     );
 
+    //serialize JWE to struct
+    let jwe_struct: JWE = match serde_json::from_slice(jwe_data.as_slice()) {
+        Ok(x) => x,
+        Err(_) => return ErrorCode::CommonInvalidParam2
```
jovfer: @Artemkaaas do we have miscounted parameters here?

PR #1794, opened 2019-08-07
IS-1336 Implement sign/verify with payment address in libvcx

KitHat:


vcx/wrappers/node/src/api/wallet.ts
```diff
@@ -76,6 +77,17 @@ export interface IPaymentAddressSeed {
   seed?: string
 }
 
+function voidPtrToUint8Array (origPtr: any, length: number): Buffer {
```
Artemkaaas: import the same function from another module

vcx/libvcx/src/utils/libindy/payments.rs
```diff
@@ -554,6 +570,21 @@ pub mod tests {
...
+        assert_eq!(res, vec![1, 2, 3])
+    }
+
+    #[test]
+    fn test_verify_with_address() {
+        init!("true");
+        let res = verify_with_address("test", &[1, 2, 3], &[1, 2, 3]).unwrap();
+        assert_eq!(res, true)
+    }
+
```
Artemkaaas: please add sign/verify tests without usage of `test mode`

PR #1804, opened 2019-08-12
use openssl hash because it is twice as fast as sha2::sha256

AxelNennker:


AxelNennker:
> we need to avoid the usages of `unwrap`

Yes, unwrap _should_ be avoided.

I think we need to rethink `parse_key_from_request_for_builtin_sp`. It should return `IndyResult<Option<Vec<u8>>>`. But currently on errors it just returns None which I find unusual.

Not sure how to proceed.
1) return `IndyResult<Option<Vec<u8>>>` and handle the errors at the caller. That is work.
2) live with unwrap because how should sha256 hashing go wrong...

Artemkaaas:
The returning None value is expected behavior for that function. From my point of view the usage of `Result` there is redundant.
you can use `ok()?` combination instead of `unwrap`.

PR #1811, opened 2019-08-13
Deserialize request into serde Value to avoid loosing of fields

Artemkaaas:


AxelNennker:
This seems to be a step in the wrong direction to me.
Why are fields lost? Should those fields be added to Request instead?
With this PR we are loosing type safety. 
I would not do this.

While I am at it, define a type Operation as well?

Artemkaaas:
@AxelNennker  We faced this issue when `payment plugins` appended field into request. This request was used later by Libindy function, but Libindy didn't aware of this field.

AxelNennker:
Intermediate question, I think that `serde_json::from_str` fails if there are extra fields that are not member of Request, right?! There is no extension point in Request and so the payment plugin was wrong to add fields.

I think that in large projects like Indy there is value in having strong types.
https://github.com/serde-rs/json#parsing-json-as-strongly-typed-data-structures exists for a reason.

There might be security implications as well, when arbitrary fields are piped through.
Might be a good thing that extra fields are 'lost'
https://jira.hyperledger.org/browse/IS-1351

Can the extra payment fields be made official by adding them to the type definition of Request?

Artemkaaas:
`Intermediate question, I think that serde_json::from_str fails if there are extra fields that are not member of Request, right?!`  -  wrong. 
serde ignors unknown fields by default. You need to use `deny_unknown_fields` attribute to fail in this case.
https://serde.rs/container-attrs.html#deny_unknown_fields

AxelNennker:
> You need to use `deny_unknown_fields` attribute to fail in this case.
New to me, thank you.

Would be interesting to add this to Request and see what the unit test (for payment) yield.

AxelNennker:
taaAcceptance vs taa_acceptance
Are there several of those? Payments uses taaAcceptance, libindy uses the other.
Locks like an argument for strong types

libindy/src/commands/ledger.rs
```diff
@@ -1162,10 +1161,10 @@ impl LedgerCommandExecutor {
         debug!("append_txn_author_agreement_acceptance_to_request >>> request_json: {:?}, text: {:?}, version: {:?}, taa_digest: {:?}, acc_mech_type: {:?}, time: {:?}",
                request_json, text, version, taa_digest, acc_mech_type, time);
 
-        let mut request: Request<serde_json::Value> = serde_json::from_str(request_json)
+        let mut request: serde_json::Value = serde_json::from_str(request_json)
             .map_err(|err| IndyError::from_msg(IndyErrorKind::InvalidStructure, format!("Cannot deserialize request: {:?}", err)))?;
 
-        request.taa_acceptance = Some(self.ledger_service.prepare_acceptance_data(text, version, taa_digest, acc_mech_type, time)?);
+        request["taaAcceptance"] = json!(Some(self.ledger_service.prepare_acceptance_data(text, version, taa_digest, acc_mech_type, time)?));
```
jovfer:
```suggestion
        request["taaAcceptance"] = json!(self.ledger_service.prepare_acceptance_data(text, version, taa_digest, acc_mech_type, time)?);
```

AxelNennker:
Value vs strong type is a performance thing also. To access a field in Value the code has to search a HashMap, with strong types the access is just BaseAddress+FieldOffset

PR #1813, opened 2019-08-13
new macro prepare_result_CString

AxelNennker:

new macro `prepare_result_CString` to handle the case where libindy returns one CString which is empty in the error case. The macro avoids creating an empty Rust string that is then converted into a CString. Overall fewer code lines.


AxelNennker:
This PR is superseded by the introduction of boxed_callback_string

PR #1817, opened 2019-08-14
Split fu

AxelNennker:


libindy/src/utils/crypto/verkey_builder.rs
```diff
@@ -32,3 +33,21 @@ pub fn build_full_verkey(dest: &str, verkey: Option<&str>) -> Result<String, Ind
...
+pub fn split_verkey(verkey: &str) -> (&str, &str) {
+    if verkey.contains(':') {
+        let splits: Vec<&str> = verkey.split(':').collect();
+        (splits[0], splits[1])
+    } else {
+        (verkey, DEFAULT_CRYPTO_TYPE)
+    }
+}
+
+pub fn verkey_get_cryptoname(verkey: &str) -> &str {
```
Artemkaaas: you can use `split_verkey` to split but return type only.

AxelNennker: True, but I hope that someone with higher Rust-fu can optimize them separately.
Or maybe refactor the verkey handling overall by parsing the verkey-str as soon as possible.
Another thing, I am  not sure about the cryptoname being None vs DEFAULT_CRYPTO_TYPE thingy. How about using None here. I think it rest of the code treats None like DEFAULT_CRYPTO_TYPE anyway.

PR #1818, opened 2019-08-14
push_str is fastest, also with_capacity

AxelNennker:


libindy/src/services/wallet/storage/default/query.rs
```diff
@@ -10,12 +10,17 @@ pub fn wql_to_sql<'a>(class: &'a Vec<u8>, op: &'a Operator, _options: Option<&st
...
     arguments.push(class);
     let clause_string = operator_to_sql(op, &mut arguments)?;
-    let mut query_string = "SELECT i.id, i.name, i.value, i.key, i.type FROM items as i WHERE i.type = ?".to_string();
+    const BASE: &str = "SELECT i.id, i.name, i.value, i.key, i.type FROM items as i WHERE i.type = ?";
     if !clause_string.is_empty() {
+        let mut query_string = String::with_capacity(BASE.len() + 5 + clause_string.len());
+        query_string.push_str(BASE);
         query_string.push_str(" AND ");
         query_string.push_str(&clause_string);
+        Ok((query_string, arguments))
```
Artemkaaas: can we just use `format!("{} AND {}", BASE, clause_string)` ?

AxelNennker: We could, but push_str is twice as fast.
https://users.rust-lang.org/t/fast-string-concatenation/4425/6

Maybe it is a question of "fast enough", developer convenience, shorter code, ... ???

Being twice as fast probably does not merit to replace all the format! in libindy (there are tons of it, not counting the error message formatting). But maybe if the code serves millions of requests per whatever then it makes a difference.

```
ignisvulpis@namenlos:~/development/rust_str_concat_bench$ cargo +nightly bench
    Finished release [optimized] target(s) in 0.67s
     Running target/release/deps/rust_str_concat_bench-d577764b29d6413b

running 4 tests
test format_url_concat ... bench:          82 ns/iter (+/- 3)
test format_url_macro  ... bench:          77 ns/iter (+/- 2)
test format_url_push   ... bench:          30 ns/iter (+/- 2)
test format_url_write  ... bench:          42 ns/iter (+/- 1)
```
 format_url_macro 77 ns vs format_url_push 30

AxelNennker: On a mobile phone accessing the wallet storage the difference does not really matter, I assume. For an agent processing many requests the cycles might add up.

PR #1829, opened 2019-08-17
trait objects without an explicit dyn are deprecated

AxelNennker:


jovfer:
@AxelNennker FYI there was some problems in performance of CLI testing on windows couple months ago. And the reason was changed backtrace version. In this PR i can see some changes similar to problematic, so we have to check time execution for tests before merging the PR

AxelNennker:
I reverted the updates to the Cargo.lock files in this PR so that the 'trait objects without dyn deprecation' patches can be merged. (Assuming that not that is causing the Windows OS performance degradation)

Any idea which crate's update makes Windows slower? We should work with the authors to address this. Staying behind by not upgrading will bite us later e.g. when a security update is needed or a feature is integrated that we badly want.

The issue with the backtrace crate has an aspect that I find interesting: in the release versions of our binaries (libindy.so, cli, libvcx, ....) we care more about correct backtraces than about performance. If a release version panics we want correct backtraces and hopefully panics are rare in release versions. 
So we have the penalty 'only' in our CI runs where panics happen more often and we actually most of the times don't need the backtrace to be totally correct.

PR #1830, opened 2019-08-18
take shortcut for api methods that do not access ledger, pool or wallet

AxelNennker:

There are methods in libindy API that do not access ledger, pool, wallet... or global resources.
These methods work on their arguments only.

Those methods can be executed without going through the executor, command, service layers.
This also saves a lot of code in those layers removing some complexity.


AxelNennker:
I see that there is no love for this PR. Nevermind.

I think that API calls that just e.g. create requests jsons or similar and that do not contact any pool and wallet and ledger should not travel through the CommandExecutor and Service layer of libindy.

Currently there is no benefit to do that but just a waste of time. Those helper methods should be in the wrappers, I think.

PR #1836, opened 2019-08-20
Liibndy test cleaning

Artemkaaas:


AxelNennker:
Would you mind to use WalletHandle instead of i32 and INVALID_WALLET_HANDLE instead of 0?
Same for PoolHandle etc, please.

PR #1842, opened 2019-08-21
update to rust 1.37

AxelNennker:


AxelNennker:
So, there is a regression in rustc 1.37.0 that this code is triggering. It is fixed in nightly but there won't be a version 1.37.1. 
I guess we have to wait for rustc version 1.38 and skip 1.37 in Indy...

jovfer:
@AxelNennker thank you for tracking this topic and clarification comment!

AxelNennker:
FYI the issue is here
https://github.com/rust-lang/rust/issues/63783 or better https://github.com/rust-lang/rust/issues/62973

Adding or removing some whitespace in our code could prevent this regression in rustc to be triggered.

PR #1853, opened 2019-08-26
Update backtrace to 0.3.35 (supports WASM)

markspanbroek:
Update Backtrace dependency to 0.3.35, which has WASM support.

This should help in porting libindy to WASM.

### DIFF
--- a/libindy/Cargo.lock
+++ b/libindy/Cargo.lock
@@ -57,14 +57,13 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
 name = "backtrace"
-version = "0.3.11"
+version = "0.3.35"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "backtrace-sys 0.1.30 (registry+https://github.com/rust-lang/crates.io-index)",
  "cfg-if 0.1.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.60 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustc-demangle 0.1.16 (registry+https://github.com/rust-lang/crates.io-index)",
- "winapi 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


libindy/Cargo.lock
```diff
@@ -57,14 +57,13 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
...
 name = "backtrace"
-version = "0.3.11"
+version = "0.3.35"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "backtrace-sys 0.1.30 (registry+https://github.com/rust-lang/crates.io-index)",
  "cfg-if 0.1.9 (registry+https://github.com/rust-lang/crates.io-index)",
  "libc 0.2.60 (registry+https://github.com/rust-lang/crates.io-index)",
  "rustc-demangle 0.1.16 (registry+https://github.com/rust-lang/crates.io-index)",
- "winapi 0.3.7 (registry+https://github.com/rust-lang/crates.io-index)",
```
jovfer: This deletion may slow down CI for 30min. I temporary set "changes requested" to avoid accidental merge by other maintainers. But as soon as CI efficiency would be checked I don't mind to cancel my review status by anyone else.

markspanbroek: Thanks @jovfer, the CI needed 2 hours and 30 minutes to check this PR. I find this rather long, but I don't know what the normal duration is. Should I add an explicit dependency on winapi to the Cargo.toml to improve the speed?

jovfer: Total time is not a good benchmark here as it may be affected by queue of other builds. The problematic step was testing of windows artifacts. So we have to analyze `Windows Test` stage (after jenkins worker node allocation).
* Current PR https://build.sovrin.org/job/indy-sdk/job/indy-sdk-ci/view/change-requests/job/PR-1853/1/flowGraphTable/ - windows tasks takes about a hour
* An example of another PR https://build.sovrin.org/job/indy-sdk/job/indy-sdk-ci/view/change-requests/job/PR-1859/1/flowGraphTable/ - windows tasks takes ~40m

jovfer: Most probably just reference winapi from Cargo.toml will not help. Probably `backtrace` should be configured somehow. I suggest to check issue section of `backtrace` crate may be someone else already hit this problem and described the solution.

PR #1668, opened 2019-06-09
WIP Sodiumoxide 0.2.2

AxelNennker:
new version 2.2 of sodiumoxide

PR #1585, opened 2019-04-13
new version of libsodium

AxelNennker:
Signed-off-by: Axel Nennker <axel.nennker@telekom.de>

PR #1854, opened 2019-08-26
Replace OpenSSL with Sodium for SHA256 Hash

markspanbroek:
Removes the dependency of libindy on OpenSSL. Sodium was already used for other cryptographic primitives, and these commits replace the last use of OpenSSL with Sodium.

This should help in porting libindy to WASM as this is one less dependency to worry about.


AxelNennker:
openssl sha256 hash is twice as fast as the other implementations.
Please see [crypto-bench](https://github.com/briansmith/crypto-bench)

I would not replace openssl now.

There is also already a PR to update to sodiumoxide 0.2.2
https://github.com/hyperledger/indy-sdk/pull/1668
which struggles with libsodium and static-only builds.
If your PR builds could you please factor out the 0.2.2 update and push it separately? I would then close my sodiumoxide 0.2.2 PR (would be interested to know the differences then)

markspanbroek:
Thanks @AxelNennker, my update to sodiumoxide 0.2.2 also exhibits the linking problem that you mentioned; the CI fails because of it.

Thanks for pointing out that the OpenSSL hash is faster than the Sodium one. I don't know enough about Indy to judge whether hash speed is more important than portability.

AxelNennker:
The linking problem goes away with this PR https://github.com/hyperledger/indy-sdk/pull/1585
If your local machine has a modern libsodium then you don't see it until CI fails.

AxelNennker:
@markspanbroek would you close this, please. Or do you want to keep the PR open?

markspanbroek:
Hi @AxelNennker, I would personally like to see the OpenSSL dependency removed, because it makes porting to several platforms painful (WASM, iOS). However, if no-one is willing to merge this pull request, then I'll close it.

AxelNennker:
@dhh1128 I would like to get rid of openssl and I had a patch in the past for that as well.
But I think that hashing is at the heart of every merkletree implementation and being twice as fast is important for servers using libindy. For mobile applications and WASM clients speed is probably not so important. 

If openssl is ditched than the question is with which digest implementation to replace it.
Ring has a good reputation as well. https://briansmith.org/rustdoc/ring/digest/index.html

AxelNennker:
I ran [crypto-bench](https://github.com/briansmith/crypto-bench) again: `cargo +nightly bench sha256::_2000`
```
openssl  test digest::sha256::_2000     4,347 ns/iter (+/- 250) = 460 MB/s
ring        test digest::sha256::_2000     4,767 ns/iter (+/- 309) = 419 MB/s
sodium   test digest::sha256::_2000       8,311 ns/iter (+/- 1,633) = 240 MB/s
```

AxelNennker:
Conditional compilation is possible
https://doc.rust-lang.org/reference/conditional-compilation.html
The OS dependency is one of the examples.

The platform I ran crypto-bench on: from rust to hw

https://github.com/AxelNennker/crypto-bench/tree/rustc1.36
```
ignisvulpis@namenlos:~$ rustc --version
rustc 1.37.0 (eae3437df 2019-08-13)
ignisvulpis@namenlos:~$ uname -a
Linux namenlos 4.15.0-64-generic #73~16.04.1-Ubuntu SMP Fri Sep 13 09:56:18 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux
```
8 x 6.94.3 "Intel(R) Core(TM) i7-6920HQ CPU @ 2.90GHz"

mikelodder7:
OpenSSL is used for its BigNum component for ZKPs. Since we are using Ursa, if you set the features to portable, then ursa removes OpenSSL entirely.  I’m unfamiliar with where else OpenSSL is being used here but I’m in favor of removing it.

Also, `ring` won’t give you the WASM properties you need. There are raw rust versions of hashing like `sha2` that use assembly if the system supports it but rust if not that does compile to WASM. Also, ursa exposes this. If indy used ursa where it can then many of these decisions have already been made

PR #1855, opened 2019-08-26
new method pool_create_pair_of_sockets

AxelNennker:

New method pool_create_pair_of_sockets


libindy/src/services/pool/mod.rs
```diff
@@ -474,8 +459,8 @@ mod tests {
         }
 
         #[test]
-        fn pool_send_tx_works_for_closed_socket() {
-            test::cleanup_storage("pool_send_tx_works_for_closed_socket");
+        fn pool_send_tx_fails_for_closed_socket() {
```
jovfer:
```suggestion
        fn pool_send_tx_works_for_closed_socket() {
```

jovfer: current name conversion for tests assumes "works" for non-happy path as well

jovfer: @AxelNennker if you will try to use github UI to "commit suggestion" do not forget manually line about DCO in commit message details

PR #1866, opened 2019-08-29
[IS-1353] Improve arch overview docs

jovfer:


docs/architecture/libindy-components.puml
```diff
@@ -0,0 +1,44 @@
+@startuml
```
Artemkaaas: please add relevant .png files

docs/architecture/libindy_layers.md
```diff
@@ -0,0 +1,65 @@
...
+
+* Library interface layer
+* Basic validation
+* Conversion of C types to Rust types
+* Propagation of execution to commands layer
+
+### Commands Executors layer
+
+* Working threads management
+* JSONs conversion to internal types and corresponded validation
```
Artemkaaas: we mostly does it at API layer

docs/architecture/libindy_layers.md
```diff
@@ -0,0 +1,65 @@
...
+* Implements operations business logic and complex validation
+* Management of sockets polling threads
+
+## Layers interaction
+### API -> API
+There is no dependencies or calls between different API modules.
+
+### API -> Commands Executor
+Root Commands Executor contains "main" thread of the library. It accepts Commands from API via standard Rust mechanism `std::sync::mpsc::channel`.
+A command contains user callback wrapped to Rust lambda. It will be called with a result of the processing.
```
Artemkaaas: maybe Rust `closure`?

docs/architecture/libindy_layers.md
```diff
@@ -0,0 +1,65 @@
+There are 3 layers in libindy code organization:
+* API layer
+* Commands Executors layer
+* Services layer
+
+Each of them contains multiply modules devoted to particular part of the flow, like "wallet", "pool", "crypto", etc. 
+
+## Layers purposes
```
ashcherbakov: 1) It would be great to emphasize the main thing about the Commands and Service layers: that Commands layer splits complex operation to atomic ones and delegate execution of atomic operation to Services where Services doesn't communicate with each other.
2) It would be great to emphasize the benefits of such encapsulation
3) It would be great to describe how secure enclave works in libindy

jovfer: 1. fixed
2. not sure, that arch overview is a place to for that. We don't have any specific benefits in libindy, just well-known advantages like testability, easier code support, etc
3. already present in design folder

PR #1873, opened 2019-09-02
IS-1358 and IS-1366 Fully-qualified DID support

KitHat:
Support of creation of DID and ledger requests.


libindy/src/services/anoncreds/verifier.rs
```diff
@@ -519,7 +485,7 @@ impl Verifier {
     fn _is_attr_operator(key: &str) -> bool { key.starts_with("attr::") && key.ends_with("::marker") }
 }
 
-#[cfg(test)]
+/*#[cfg(test)]
```
KitHat: Why is this test commented?

libindy/src/utils/qualifier.rs
```diff
@@ -0,0 +1,41 @@
+use regex::Regex;
+
+pub const DEFAULT_PREFIX: &'static str = "did:sov:";
```
jovfer: Most probably it should be configurable

jovfer:
replaced by #1896

PR #1892, opened 2019-09-20
Adds failing unit test for attr::<attr_name>::value restriction

nrempel:
Signed-off-by: Nicholas Rempel <nbrempel@gmail.com>


swcurran:
@Artemkaaas - a question about this.  When a group of claims have a single restriction, is there any checking that the claims from the group came from the same credential?  I'm guessing not (and I haven't looked at the code :-( ), but just wanted to ask you.

Thanks!

Artemkaaas:
Hello, @swcurran 
Not sure if I understood your question correctly.

Suggest that Proof Request contains requested attributes:
```
"attr1_referent": {
    "name":"name",
    "revealed": "true",
    "restrictions:" {
        "attr::name::value": "Artem"
    },
},
"attr2_referent": {
    "name":"age",
    "revealed": "true",
    "restrictions:" {
        "attr::age::value": "25"
    },
}
```
Prover can't understand if he must use the same credentials for both attributes (based on the ProofRequest structure).
But if he knows he can create a Proof that will contain:
```
"requested_proof": {
    "revealed_attrs": {
        "attr1_referent": {sub_proof_index: index_1, raw: string, encoded: string},  
        "attr2_referent": {sub_proof_index: index_1, raw: string, encoded: string},            
    }
}
```
Verifier can check `sub_proof_index` correspondent to `attr1_referent` is equal to `sub_proof_index` for `attr2_referent`. It means that they match to the same credential.

But we can't do this check on Libindy level (we don't have enough bases) without doing ProofRequest restrictions more powerfull.
We thought regarding adding an additional restriction (like `reference`) to ProofRequest restrictions instead of `attrs` field. But it required much more work. So, we did the faster one and kept `restriction improving` as a future improvement.
It also covers the case when we need a predicate that corresponds to the same credential.

Artemkaaas:
If we are talking about `names` attribute the fact that they refers to one referent is enough to ensure that single credential was used. We don't need any additional check.

swcurran:
Thanks for the answers. Sorry for the slow reply.

Yes, it is the names attribute that is was referencing. I just found the documentation on it here and I see that it does guarantee it comes from the same credential - https://github.com/hyperledger/indy-sdk/blob/f708f496dd23ada341303db7a4e6a0a7e29a549f/docs/migration-guides/migration-guide-1.12.0-1.13.0.md

Nice!  That's what we are looking for.

PR #1896, opened 2019-09-24
Fully qualified DID support

Artemkaaas:


libindy/src/domain/crypto/did.rs
```diff
@@ -1,43 +1,144 @@
...
     pub seed: Option<String>,
     pub crypto_type: Option<String>,
-    pub cid: Option<bool>
+    pub cid: Option<bool>,
+    pub method_name: Option<String>,
+}
+
+impl Validatable for MyDidInfo {
+    fn validate(&self) -> Result<(), String> {
+        if let Some(ref did) = self.did {
```
mikelodder7: Could you use `did_doc` from crates.io?

jovfer: libindy `MyDidInfo` is significantly diverse from W3C DIDDoc

PR #1905, opened 2019-10-01
Ursa 0.2.0

AxelNennker:
update ursa to new version

This might solve linking issues caused by different versions of openssl-sys being used by ursa and libindy

Also updated Cargo.lock


jovfer:
@AxelNennker could you work with @mikelodder7 and make sure that releases or tags exists in the Ursa repo and they are synced to crates.io releases. I think it would be helpful to have ability to check the differences between Ursa versions. As soon as it's not formal releases I don't think special changelog should exist, but at least tags should be useful from my point of view.

AxelNennker:
There are releases in ursa but libindy uses the dev version.
So think it is already as you suggest.

I think there is an issue with the dev-1 version that uses a version of openssl-sys that is different to the version libindy is using and because of it we have link problems on Ubuntu18.
I don't see the link currently in my PRs anymore...

Not sure how to proceed regarding openssl in ursa and libindy. Reexporting everything (openssl, libsodium, ...) from Ursa is not unproblematic in itself because libursa gets too fat and projects dependent on libursa might be able to follow all libursa decisions.
Currently libindy is using openssl for sha256 and libursa is using openssl for bignumber stuff.

https://github.com/hyperledger/ursa/pull/61

jovfer:
@AxelNennker @mikelodder7 taking into account few latest comments in https://github.com/hyperledger/ursa/pull/61 I'm in doubt about this PR...

Should ursa be republished without Cargo.lock to crates.io and then we can update this PR and merge it?

AxelNennker:
ursa dev-2 has the same openssl-sys as libindy and I assume that is good for libindy's Ubuntu18 builds where we saw some link errors

In general, I think, libindy should not use the ursa dev crates but the _released_ versions.
And the ursa crate should not have Cargo.lock which led to the situation that they used different versions of openssl. Neither should libindy use a binary libursa.dylib/so/a but a ursa rust crate with some 0.1.* version number.

Regarding ursa's publishing dev versions to crates.io, I suggest not to do that. If a rust program or library wants to use a pre-release version it can use cargo's git feature and get the pre-release version from there.

mikelodder7:
Ursa-0.2.0-dev-6 is up. Its the same as dev-5 except it includes a CHANGELOG and LICENSE

PR #1909, opened 2019-10-01
[IS-1382] Makes pyzmq version 17.0.0 dependency explicit

svdo:
Building the docker image for running the test pool on localhost (https://github.com/hyperledger/indy-sdk#1-starting-the-test-pool-on-localhost) raised an error: python3-pyzmq version 18.x.x was available, but 17.0.0 needed. This PR makes the dependency on version 17.0.0 explicit.

Solves https://jira.hyperledger.org/browse/IS-1382.


svdo:
That's fast, great job @jovfer, thanks!

PR #1914, opened 2019-10-03
Switch to Rust 2018 edition

jovfer:


jovfer:
We've decided to move other components to new edition in separate PR

PR #1918, opened 2019-10-04
step2.py

dorry:
Fixed issue.  parsed genesis_file_path to a string since json.dumps() doesn't expect to get a PosixPath because it's not JSON serializable.


dhh1128:
@dorry: Thanks for this contribution. It's great, and we want to merge it. However, we need you to comply with Hyperledger's DCO policy by signing your commit. This tells the world that you agree to contribute under the terms of the Apache license. If you have deep git kung-fu, you can do that with `git rebase -i`, altering the commit message to add the appropriate `Signed-off-by: <you>` statement, and force-pushing the update. If that sounds ugly, just close this PR and start a new one  based on a commit that has the identical content, but that is created with `git commit -s` so it's signed.

dorry:
Thank you for your help!

PR #1920, opened 2019-10-04
Fixed bugs in issue-credential by adding the issuer and prover wallet…

dorry:
Fixed bugs in issue-credential by adding the issuer and prover wallet creation and it's exception handling since it was missing and added the wallet module followed by a dot operator since open_wallet() is a function of the wallet module. Also parsed genesis_file_path to string in save_schema_and_cerd_def since PosixPath isn't JSON serializable

Signed-off-by: dorry <eldorry3@gmail.com>


dorry:
@brentzundel  Are you going to merge my pull request or is there something else that I should do? excuse my ignorance  :")

PR #1921, opened 2019-10-05
unneeded return statement

AxelNennker:


AxelNennker:
I see no way this PR can cause this test anoncreds_revocation_interaction_test_issuance_by_demand to fail.

PR #1923, opened 2019-10-07
Node.js Sample: renamed function utils.pathToIndyClientHome to utils.getPathToIndyCli…

dunff2k:
…entHome. This fixes a crash that occurs when running anoncredsRevocation.js


dunff2k:
I tried to fix the DCO as specified by the link  above - but the check still fails

PR #1925, opened 2019-10-07
Update vcx agent profile before create agent connection

Patrik-Stas:
I suppose things might work differently in Evernym proprietary agency, so maybe rather an adjustment on dummy cloud agency is needed. But if this is not breaking Evernym stuff, it would fix issue with dummy cloud agency - an Agent's first created connection is due to this order not injected with agent configuration such as `logoUrl` or `institutionName`.


Artemkaaas:
@ryMarsh44  @nsivraj  Could you look at this?

PR #1958, opened 2019-11-24
IS-1381 -- adding new field "names" to proof request

KitHat:
- [x] Update API docs


libindy/src/domain/anoncreds/proof_request.rs
```diff
@@ -178,7 +181,8 @@ impl Validatable for ProofRequest {
         }
 
         for (_, requested_attribute) in value.requested_attributes.iter() {
-            if requested_attribute.name.is_empty() {
+            if (requested_attribute.name == None || requested_attribute.name == Some("".to_string())) &&
```
jovfer:
```suggestion
            if (requested_attribute.map(str::is_empty).unwrap_or(true)) &&
```

jovfer: and the similar for vector below

libindy/src/services/anoncreds/prover.rs
```diff
@@ -413,13 +436,23 @@ impl Prover {
 
     pub fn extend_proof_request_restrictions(&self,
                                              version: &ProofRequestsVersion,
-                                             name: &str,
+                                             name: &Option<String>,
+                                             names: &Option<Vec<String>>,
```
jovfer: Are allocated string and vector required here and match the code on the level above?

libindy/src/services/anoncreds/prover.rs
```diff
@@ -413,13 +436,23 @@ impl Prover {
...
                                              referent: &str,
                                              restrictions: &Option<Query>,
                                              extra_query: &Option<&ProofRequestExtraQuery>) -> IndyResult<Query> {
-        let mut queries: Vec<Query> = vec![
-            Query::Eq(format!("attr::{}::marker", &attr_common_view(name)), ATTRIBUTE_EXISTENCE_MARKER.to_string())
-        ];
+        info!("name: {:?}, names: {:?}", name, names);
+        let mut queries: Vec<Query> = if let Some(name) = name {
```
jovfer: Have we considered the option to merge `name` and `names` parameters for this function? It may be more straightforward.

libindy/src/domain/anoncreds/proof.rs
```diff
@@ -45,6 +47,17 @@ pub struct RevealedAttributeInfo {
     pub encoded: String
 }
 
+#[derive(Debug, Deserialize, Serialize)]
+pub struct RevealedAttributeGroupInfo {
+    pub sub_proof_index: u32,
+    pub values: HashMap<String, AttributeValue>,
```
jovfer:
```suggestion
    pub values: HashMap<String /* attribute name */, AttributeValue>,
```

wrappers/python/tests/anoncreds/test_verify_proof.py
```diff
@@ -56,6 +56,7 @@
                 "sub_proof_index": 0, "raw": "Alex", "encoded": "1139481716457488690172217916278103335"
             }
         },
+        "revealed_attr_groups": {},
```
jovfer: As far as I understand `revealed_attr_groups` is optional, why we've added it here?

wrappers/java/src/test/java/org/hyperledger/indy/sdk/anoncreds/VerifierVerifyProofTest.java
```diff
@@ -53,7 +53,7 @@
 			"       }" +
 			"   }," +
 			"   \"requested_proof\":{" +
-			"       \"revealed_attrs\":{\"attr1_referent\":{\"sub_proof_index\":0,\"raw\":\"Alex\",\"encoded\":\"1139481716457488690172217916278103335\"}},\"self_attested_attrs\":{},\"unrevealed_attrs\":{},\"predicates\":{}},\"identifiers\":[{\"schema_id\": \"\",\"cred_def_id\": \"\",\"rev_reg_id\":null,\"timestamp\":null}" +
+			"       \"revealed_attrs\":{\"attr1_referent\":{\"sub_proof_index\":0,\"raw\":\"Alex\",\"encoded\":\"1139481716457488690172217916278103335\"}},\"revealed_attr_groups\":{},\"self_attested_attrs\":{},\"unrevealed_attrs\":{},\"predicates\":{}},\"identifiers\":[{\"schema_id\": \"\",\"cred_def_id\": \"\",\"rev_reg_id\":null,\"timestamp\":null}" +
```
jovfer: As far as I understand `revealed_attr_groups` is optional, why we've added it here?

libindy/tests/utils/constants.rs
```diff
@@ -23,9 +24,12 @@ pub const INVALID_IDENTIFIER: &'static str = "invalid_base58_identifier";
 pub const DEST: &'static str = "FYmoFw55GeQH7SRFa37dkx1d2dZ3zUF8ckg7wmL7ofN4";
 pub const DEST_V1: &'static str = "did:sov:FYmoFw55GeQH7SRFa37dkx1d2dZ3zUF8ckg7wmL7ofN4";
 pub const GVT_SCHEMA_NAME: &'static str = "gvt";
+pub const GVT_SUB_SCHEMA_NAME: &'static str = "gvtsub";
```
jovfer: What is the reason to introduce new schema?

KitHat: To have a credential with partial matching fields and try to cheat on it.

wrappers/java/src/test/java/org/hyperledger/indy/sdk/anoncreds/VerifierVerifyProofTest.java
KitHat: All of our params are optional, but we are adding them nevertheless.
I have just followed the old pattern.

libindy/src/services/anoncreds/prover.rs
KitHat: What do you mean by match the code?

libindy/src/domain/anoncreds/proof_request.rs
KitHat: This is funny, how me, @Artemkaaas and you have three different points of view on this code.
I don't think that `map` and `unwrap_or` is more clear than current variant.
It needs to be thought about whilst equality checks are clear right from the start.

libindy/src/services/anoncreds/prover.rs
KitHat: Good catch, found a cute way to do it

libindy/src/domain/anoncreds/proof_request.rs
jovfer: I don't mind to to have separate statements for 2 cases, but we shouldn't allocate temporary empty string just to compare with the input

libindy/src/services/anoncreds/prover.rs
jovfer: usability on the level above

libindy/src/services/anoncreds/verifier.rs
```diff
@@ -485,13 +561,20 @@ impl Verifier {
         }
     }
 
-    fn _is_attr_internal_tag(key: &str, attr: &str) -> bool {
-        key == format!("attr::{}::value", attr) || key == format!("attr::{}::marker", attr)
+    fn _is_attr_internal_tag(key: &str, attr_value_map: &HashMap<String, Option<&str>>) -> bool {
+        INTERNAL_TAG_MATCHER.captures(key).map( |caps|
```
jovfer: can be simplified by and_then

PR #1961, opened 2019-11-27
IS-1395: LibVCX Aries: HTTP server to run Aries Test Suite against VCX

jovfer:


jovfer:
would be merged as a part of #1960

PR #1977, opened 2019-12-09
Tests  for empty attribute in libindy

KitHat:


jovfer:
@KitHat @Artemkaaas please make sure that we have high-level tests at the top of libVCX API

vcx/libvcx/src/issuer_credential.rs
```diff
@@ -999,6 +999,17 @@ pub mod tests {
         assert!(!string.is_empty());
     }
 
+    #[test]
+    fn test_empty_val_succeeds() {
```
Artemkaaas: This test actually doesn't check anything.

KitHat: This test checks that empty parameter value does not cause validation failure with empty param

PR #1987, opened 2019-12-16
In pgsql plugin deprecate min_idle_time, supersede by min_idle_count

Patrik-Stas:
The option `min_idle_time` was being used in r2d2's builder `min_idle(u32)` method which is described in r2d2 documentation as:
```
    /// Sets the minimum idle connection count maintained by the pool.
    ///
    /// If set, the pool will try to maintain at least this many idle
    /// connections at all times, while respecting the value of `max_size`.
    ///
    /// Defaults to `None` (equivalent to the value of `max_size`).
    pub fn min_idle(mut self, min_idle: Option<u32>) -> Builder<M> {
        self.min_idle = min_idle;
        self
    }
```
so to prevent confusion, I renamed min_idle_time plugin configuration option to more accurate `min_idle_count`. The `min_idle_time` still works but prints warning in logger.


Patrik-Stas:
Hi @ianco , can you take a look at this as well?

PR #1992, opened 2019-12-17
docs(how-tos): fixes typo in comment

petermetz:


petermetz:
Hmm, I'm guessing the CI is just broken or flaky right now, correct? 
Making the assumption because a literally changed a single character in documentation only and the CI is red anyway.

PR #2026, opened 2020-01-10
IS-1431 -- Add a sourcelist entry after the manual installation of ar…

KitHat:
…tifacts


jovfer:
I think we should try to find a way to avoid duplication in this changes

PR #2027, opened 2020-01-10
Add pgsql storage strategy sharing connection pool between wallets

Patrik-Stas:
- Strategy `MultiWalletSingleTable` is creating many connection pools, one connection pool per wallet
- This adds `MultiWalletSingleTableSharedPool` strategy which shares postgre connections across wallets
- This also happens to remove several `unsafe` sections in the plugin

### DIFF
--- a/experimental/plugins/postgres_storage/README.md
+++ b/experimental/plugins/postgres_storage/README.md
@@ -75,19 +75,28 @@ cd indy-sdk/experimental/plugins/postgres_storage
 RUST_BACKTRACE=1 cargo test -- --nocapture --test-threads=1
 ```
 
-## Database-per-wallet vs Multi-wallet database
+## Wallet management modes 
+The plug-in supports three modes 
+- `DatabasePerWallet` - each wallet has its own database
+- `MultiWalletSingleTable` - all wallets are stored in single table in single database. Each wallet has its own 
+connection pool.
+- `MultiWalletSingleTableSharedPool` - all wallets are stored in single table in single database. The plugin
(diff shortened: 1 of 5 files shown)


Patrik-Stas:
Hi @ianco can you please have a look?

Patrik-Stas:
@ianco Thanks for review! The problem was I've introduced a breaking change where not specifying strategy did not pick `DatabasePerWallet` as the default strategy. I've fixed the problem and made a few additional cleanups.

Patrik-Stas:
Hi @ianco , I pushed some few more commits but I think I am done with this. Let me know if there's something more to adjust or fix.

Patrik-Stas:
Hi @ianco don't want to rush you, just following up in case you forgot about this PR. Please review when you have some time, thanks a ton!

Patrik-Stas:
@ianco Thank you for review. I was previously running tests, but did not realize that since the strategy is not explicitly added to unit test setups, it will used default strategy if not recognized. 

So this is rebased on recent master with few new commit fixing the issues you pointed out. There was actually 2:
- Tests were failing because changes I made `PostgresStorageType::init_storage` forbid to call `init_storage` twice. It seems that this was previously allowed. The tests in `lib.rs` were not cleaning up by themselves (`init_storage` was called with some in one test and then again in another test without resetting the state of `postgress_storage.rs`)
- Another was implementation issue in the new strategy where opening a wallet did not perform a check whether the wallet exists. Checking now added)

In addition
- Sometimes I was getting random failures in `lib.rs` and this was addressed in separate PR https://github.com/hyperledger/indy-sdk/pull/2067

PR #2033, opened 2020-01-13
Fix intermittent pgsql wallet unittest issues, add logs

Patrik-Stas:
- Where functions signatures specified arg of type `*const c_char`, the data is
  sent here everywhere as `CString`, while `options_json` was passed as `*const i8`,
  causing that sometime, a random character were added to end of this string, making
  it invalid JSON and eventually throwing `serde_json::Error` then converted into
  `WalletStorageError::ConfigError`.
- Sometimes when running tests with `DatabasePerWallet` strategy, one test expects
  that a wallet with name `walle1` does not exist before it's supposedly destroyed
  at cleanup, but when the test attempts to create this wallet, it fails on
  `WalletAlreadyExistsError`. Adding `1`ms delay at the end of testing function `_cleanup()`
 solved this issue.
- Add some error logging on failures before a specific error is mapped into
  a more generic error such as `WalletStorageError`.


Patrik-Stas:
Hi @ianco , as you pointed out in my other PR failing unit tests, I started fixing that issue and I found that many unit tests were failing intermittently on my system (on the v1.14.1 release).

So this should fix it, getting all 180 tests passed on both strategies with these changes. On master branch I typically get randomly 5 to 10 failing unit tests.

PR #2035, opened 2020-01-14
IS-1166: DID API should allow pairwise key rotation

Artemkaaas:


wrappers/ios/libindy-pod/Indy/Wrapper/IndyDid.h
```diff
@@ -93,6 +94,10 @@
  @endcode
  
  @param identityJSON Identity information as json. See example above.
+    {
+       "did": string, (required)
+       "verkey": string (optional, can be avoided if did is cryptonym: did == verkey),
```
kdenhartog: Is this saying a string is not needed if this is being first the first time the DID is being stored, but if a DID is being updated, then the verkey would be needed? Additionally, Is the verkey required to be in the wallet already to perform this update?

kdenhartog: Also, do we need to do a check on the format of the key (e.g. base58btc)

Artemkaaas: `Is this saying a string is not needed if this is being first the first time the DID is being stored, but if a DID is being updated, then the verkey would be needed?` - make sense. Thanks

`Additionally, Is the verkey required to be in the wallet already to perform this update?` - verkey is already stored together with DID in the wallet. 
verkey is optional only if cryptonym DID was passed. In this case Verkey will be the same as DID.
If DID is short and verkey isn't passed the validation error will be thrown.

Artemkaaas: Actualy, there is no any changes in the function definition. I just noticed that `identityJSON` definition was missed in some wrappers.

PR #2037, opened 2020-01-15
IS-1425: Mark Indy SDK Getting Started docs as deprecated

Artemkaaas:


README.md
```diff
@@ -15,7 +15,8 @@ If you have just started learning about self-sovereign identity, here are some r
 
 * This extended tutorial introduces Indy, explains how the whole ecosystem works, and how the
 functions in the SDK can be used to construct rich clients: [Indy-SDK Getting-Started Guide](docs/getting-started/indy-walkthrough.md)
-
+**Please take note** that this tutorial doesn't contain a part related to peer-to-peer interaction.
```
dhh1128: I find this comment extremely confusing. Alice and Faber and Thrift are all peers. How can we say that our tutorial doesn't cover peer-to-peer interactions, then? What do we mean? We need probably 2-3 sentences more of explanation that contrasts what the tutorial does cover with what it does not cover. Perhaps something like this: "This tutorial focuses on an interaction between an individual and several institutions. Institutions register DIDs on a ledger so they can issue credentials. We won't cover how Alice could connect to another private individual. That involves the use of peer DIDs [link to peer did spec] and is explored in greater detail in Aries."

docs/getting-started/indy-walkthrough.md
```diff
@@ -35,9 +35,11 @@ If the concepts of cryptography and blockchain details feel mysterious, fear not
 
 Our goal is to introduce you to many of the concepts of Indy and give you some idea of what happens behind the scenes to make it all work.
 
-We're going to frame the exploration with a story. Alice, a graduate of the fictional Faber College, wants to apply for a job at the fictional company Acme Corp. As soon as she has the job, she wants to apply for a loan in Thrift Bank so she can buy a car. She would like to use her college transcript as proof of her education on the job application and once hired, Alice would like to use the fact of employment as evidence of her creditworthiness for the loan.
+**Please take note** that we are not going to cover peer-to-peer interaction as part of this tutorial.
```
dhh1128: See my previous comment. Saying we're not covering peer-to-peer makes no sense to me.

README.md
Artemkaaas: Thank you. I agree. it is really confusing. You didn't understand what I wanted to say and suggested a variant from different area (I think).
What I wanted to say: 
   The guide included a suggestion on how sides can exchange DIDs, encode and exchange messages. 
  We deleted this parts in favor of Aries project. 

I updated the phrase. Could you look at a new variant?

```
**Please take note** that we are not going to cover how sides set up connection and interact with each other as part of this tutorial.
We assume that already there is established some communication channel between each pair and this channel can be used for message exchange.
Instead of message sending we will just use construction like:
```
    faber['alice_reply'] = alice['reply']
    alice['faber_response'] = faber['response']
```

How this communication channel can be built you can find at [Aries](https://github.com/hyperledger/aries) project which describes it in great details.
```

dhh1128: Yes, the new update makes a lot more sense. Thanks.

PR #2039, opened 2020-01-16
Fix for IS-1470 improve query performance for proof request

ianco:
Signed-off-by: Ian Costanzo <ian@anon-solutions.ca>

Improve performance of queries generated via anoncreds when requesting a proof request, significant improvement for large wallets (> 300k credentials)

See JIRA IS-1470


jovfer:
@ianco thank you for researching the low performance and suggesting the fix. After reading the IS-1470 description I can see the situation as the root problem is in WQL to SQL translator or database indexes, but not in the scope of this PR changes.
I'm assuming that my understanding is wrong due to lack of details, so could you share more details in the ticket to highlight that the problem is actually in initial WQL, not in the WQL-> SQL translator or db structure.
FYI @sklump @esplinr @Artemkaaas

ianco:
Hi @jovfer thanks for reviewing this PR.  I've added some information to the ticket IS-1470 including  a couple of WQL/SQL examples generated for our proof request.

I tried optimizing the database indexes, with limited success.  I was able to get minor performance improvements but the query still degraded badly with large data volumes.

I don't think there's any optimization possible within the plug-in, based on the information available to the plug-in.  (The plug-in gets low-level queries and only deals with encrypted values.  The SQL is generated from the WQL and follows the same structure as the WQL passed in.)

The database structure itself is very simple, the only tables involved in this query are item and tag_encrypted.  Based on how the plug-in is used within the indy-sdk I don't know how this structure can be tuned for any specific type of query.

The change in anoncreds is very minor and results in about a 100x performance improvement, and doesn't suffer the same performance degradation as the data volume increases.

PR #2043, opened 2020-01-18
use version 0.2.5 of sodiumoxide instead of 0.0.19

AxelNennker:


AxelNennker:
I think the sodium_static feature can and should be removed from Cargo.toml.
The libsodium build process has enable-shared=yes and enable-static=yes as a default so both libraries are created, which is good because that uses -fPIC.
sodium_static was introduced for linux package building which needs the static library but who cares if the static library is created anyways

PR #2049, opened 2020-01-24
Update libc version in rust wrapper, indy-sys and vcx agency

Patrik-Stas:
Necessary to migrate to vcx agency to use `actix-web 2.0` and `actix 0.9.0` (using `std::future`) as they transitively require `libc > 0.2.65`.  

I''l take this opportunity to also ask you before I get into it, @Artemkaaas do you approve this effort? New versions of actix, hence migration to `std::future` and `async`. Thank you.


Artemkaaas:
I like the idea of updating `dummy` ti use `async/await`. It must become much easier.

PR #2052, opened 2020-01-24
Use 2.0 agency protocol by default

Patrik-Stas:
- Someone contacted me that Node VCX doesn't work. Turned it's because Node VCX demo, just like main Python demo are using Aries protocol as default. The default sample agency configuration was however using `1.0` by default. So out of the box, the demos fail due to incompatible defaults.
- Also fixed some formatting python postgresql demo.


Patrik-Stas:
Thanks for guidance @Artemkaaas though...any idea what am I doing wrong right now? It would appear that the `sed` step is somewhat ignored after my latest commit. 
Looking at an arbitrary few days old pipeline run https://build.sovrin.org/blue/organizations/jenkins/indy-sdk%2Findy-sdk-cd/detail/master/1440/pipeline/1096 i see here it was running on CentOS, but my failed CI/CD now was running on RHEL. Could it be that sed is not installed or something? I can't see what I am doing wrong.

Artemkaaas:
> Thanks for guidance @Artemkaaas though...any idea what am I doing wrong right now? It would appear that the `sed` step is somewhat ignored after my latest commit.
> Looking at an arbitrary few days old pipeline run https://build.sovrin.org/blue/organizations/jenkins/indy-sdk%2Findy-sdk-cd/detail/master/1440/pipeline/1096 i see here it was running on CentOS, but my failed CI/CD now was running on RHEL. Could it be that these 2 execution environments are slightly different? Missing sed or something? I can't see what I am doing wrong.

Jenkins takes pipeline from the master branch in case the author doesn't have write permissions in the repository. I re-ran your PR with your changes.

Jenkinsfile.ci
```diff
@@ -554,11 +555,10 @@ def linuxVcxBuild(file, env_name, network_name) {
         testEnv.inside("--network=${network_name}") {
             sh '''
                 cd vcx/dummy-cloud-agent
-                sed -i 's/\\("protocol_type": "\\)1.0/\\12.0/' config/sample-config.json
                 LIBRARY_PATH=./ LD_LIBRARY_PATH=./:${LD_LIBRARY_PATH} RUST_BACKTRACE=1 RUST_LOG=indy=info cargo run config/sample-config.json &
```
Artemkaaas: You either need to move running 2.0 tests before 1.0 or update config/sample-config.json again to return 1.0 protocol version. `testEnv.inside` keeps folder changes here.

Patrik-Stas: Seems like you've pushed the fix for me, anything on my side I can do? By the way, seems like jenkins pipeline is no longer public - when I click `details` on CI run, it displays Jenkins login page.

PR #2054, opened 2020-01-25
replace i32 by proper types

AxelNennker:


AxelNennker:
I don't see what is wrong here.
A wallet is expected to be there but it isn't.
Is this because some wallets in the tests are still named "wallet_1"?

I think that each test should have a unique wallet name because that enables tests to run in parallel.

```
axelnennker@Axels-MBP libindy-pod (replace_i32_by_proper_types) $ fgrep -r wallet_1 .
./Indy-demoTests/Test Utils/TestUtils.m:    return @"wallet_1";
./Indy-demoTests/Test Utils/TestUtils.m:    return @"{\"id\":\"wallet_1\"}";
./Indy-demoTests/Case Tests/Wallet/WalletHighCases.m:    NSString *walletConfig = @"{\"id\":\"wallet_1\", \"storage_type\":\"unknown_type\"}";
axelnennker@Axels-MBP libindy-pod (replace_i32_by_proper_types) $ 
```

I am not an iOS programmer, so won't attempt to change this.

PR #2057, opened 2020-01-27
Aries-396 Base64url_safe without padding

Artemkaaas:


libindy/indy-utils/src/crypto/base64/rust_base64.rs
```diff
@@ -13,11 +13,11 @@ pub fn decode(doc: &str) -> Result<Vec<u8>, IndyError> {
 }
 
 pub fn encode_urlsafe(doc: &[u8]) -> String {
-    base64::encode_config(doc, base64::URL_SAFE)
+    base64::encode_config(doc, base64::URL_SAFE_NO_PAD)
```
jovfer:
```suggestion
    base64::encode_config(doc, base64::URL_SAFE)
```

jovfer: I suggest to do it in separate PR, as soon as all community implementation in Aries would be able to parse both options

PR #2064, opened 2020-01-28
update zeroize to 1.1.0

AxelNennker:

update zeroize to version 1.1.0

To reduce the number of crates that are build in several versions.
See `cargo tree -d` output


AxelNennker:
I assume that `signal: 4, SIGILL: illegal instruction` is unrelated to this PR...

PR #2071, opened 2020-02-04
Feature/is 1467 Move crypto functions to indy-utils sub crate

Artemkaaas:


libindy/indy-utils/src/crypto/mod.rs
```diff
@@ -38,7 +47,315 @@ pub mod xsalsa20;
...
+pub mod verkey_builder;
+
+use indy_api_types::errors::prelude::*;
+use verkey_builder::{DEFAULT_CRYPTO_TYPE, clear_verkey};
+
+pub fn create_key(key_info: &KeyInfo) -> IndyResult<Key> {
+    trace!("create_key >>> key_info: {:?}", secret!(key_info));
+
+    match key_info.crypto_type.as_ref().map(String::as_str) {
+        Some(DEFAULT_CRYPTO_TYPE) | None => {}
```
AxelNennker: Using DEFAULT_CRYPTO_TYPE is only correct as long as the default is ed25519.

Artemkaaas: I added this `match` to keep old behavior. else we get other error in this case.

PR #2075, opened 2020-02-06
Fix/pgsql default strategy

Patrik-Stas:


Patrik-Stas:
Hi @ianco 
- As you've pointed out on chat that `init_storage` is optional but not required, I did modifications to make it so...so now:
- Default strategy without calling `init_storage` is `DatabasePerWalletStrategy`
- It is allowed to call `init_storage` multiple times switching between strategies
- As a result I removed `Option<>` wrapper around the selected strategy, as now some strategy is always set
- Reverted changes I previously did in `_cleanup()`methods in unit tests. Originally `_cleanup()` was not modifying strategy and neither it's now.

PR #2077, opened 2020-02-07
[IS-1491] Fix validation in case of leading zeros in encoded attr value.

jovfer:


libindy/src/services/anoncreds/verifier.rs
```diff
@@ -360,6 +360,7 @@ impl Verifier {
                                         proof: &Proof,
                                         attr_info: &RevealedAttributeInfo) -> IndyResult<()> {
         let reveal_attr_encoded = attr_info.encoded.to_string();
+        let reveal_attr_encoded = Regex::new("^0*").unwrap().replace_all(&reveal_attr_encoded, "").to_owned();
```
sklump: DANGER!

This changes "0" into "", which then causes indy to reject the proof. Zero is a perfectly valid number. Arguably also valid is stuff like "-012345".

sklump: ... the ultimate perverse case of -0 could be checked separately.

Easier: convert to int and convert back to string?

jovfer: @sklump thanks for the finding! Let us double-check from math point of view is "0" valid encoding or not at all... If it's valid, we will fix validation here, if not - we will add more validation at creation.

sklump: Of course "0" is a valid encoding: (stringified) 32-bit ints encode to themselves, so zero produces an encoding of "0".

I've had a test case for over a year with an attribute value of zero, and only just now has indy-sdk started rejecting a proof on it - otherwise I wouldn't have found it.

jovfer: @sklump I mean "valid" from the ZKP math point of view. In other words, I would like to double-check the absence of side-effects of using zero as an encoded value. Like in some math it may introduce more correlation, so in such cases, the validation for encoded should be smth like "positive integer (>0)". But as far as I remember it's not our case - CL-signature requires non-negative integers as encodings.

jovfer: >Note that in some cases we want leading zeroes preserved in revealed (stringified) int32-valued attributes: e.g., credit card month, 4-digit PIN, anything that fits a zero-padded template.

you are correct here, and all these use cases can be covered by zero-padded raw and trimmed encoded values. You may find some concerns for a bit different use case in https://jira.hyperledger.org/browse/IS-1492. Fill free to add you comment there if you can see more specific use cases,

brentzundel: From the ZKP math point of view, we should not be allowing zero as an encoded value.
When we reached out to Jan Camenisch for feedback on our system, he was adamant that we check for zeros in the attribute values and blinding factors. This is why the integer encoding method for Rich Schemas zero-centers integer values at 2<sup>255</sup>, to allow for zero and negative integer values as credential values.

PR #2085, opened 2020-02-12
Fix/vcx node wrapper demo, tests libloading

Patrik-Stas:
- VCX Node unit tests were hardcoded to pass on Linux only
- VCX Node Demo was hardcoded to work on OSX only
- Linted demo related code


vcx/wrappers/node/src/vcx.ts
```diff
@@ -35,6 +34,6 @@ export class VCXRuntime {
     const libDir = libPath[platform.toLowerCase()] || libPath['linux']
     const library = `libvcx${postfix}`
     const customPath = process.env.LIBVCX_PATH ? process.env.LIBVCX_PATH + library : undefined
-    return customPath || this._config.basepath || path.resolve(__dirname, `${libDir}${library}`)
+    return customPath || this._config.basepath || `${libDir}${library}`
```
Patrik-Stas: The line `path.resolve(__dirname, `${libDir}${library}`)` was not making sense as the second parameter for `path.resolve` was always absolute, and `path.resolve` will in that case ignore the first argument.
```
> path.resolve("./relative", "/bar/foo")
'/bar/foo'
> path.resolve("/absolute", "/bar/foo")
'/bar/foo'
```

vcx/wrappers/node/test/suite2/ffi.test.ts
```diff
@@ -20,8 +22,15 @@ describe('vcxInit', () => {
 // these tests were created to only test that the ffi could be called with each function
 
 describe('Using the vcx ffi directly', () => {
-  const path = '/usr/lib/libvcx.so'
```
Patrik-Stas: was failing on mac

vcx/wrappers/node/demo/common.js
```diff
@@ -1,13 +1,24 @@
...
+
+function getLibraryPath (libraryName) {
+  const platform = os.platform()
+  const postfix = extension[platform.toLowerCase()] || extension.linux
+  const libDir = libPath[platform.toLowerCase()] || libPath.linux
+  return `${libDir}${libraryName}${postfix}`
+}
 
 async function loadPostgresPlugin (provisionConfig) {
-  const myffi = ffi.Library('/usr/local/lib/libindystrgpostgres.dylib', { postgresstorage_init: ['void', []] })
```
Patrik-Stas: was failing on non-mac

PR #2096, opened 2020-02-17
add maven wrapper to maintain build environment more  easily.

kukgini:
Signed-off-by: kukgini <kukgini@gmail.com>


Artemkaaas:
@kukgini  Could provide a description on how to use it / what is the benefit. None of us are familiar with it.

kukgini:
@Artemkaaas I'm not goot at English. 
But I believe this would be a good explanation: https://www.baeldung.com/maven-wrapper
It's the same idea on graddle wrapper included in vcx java wrapper.
You can use `./mvnw` script instead `mvn`.  the rest is same as original maven.
It is no halmful to who want to use original maven. 
It is useful to who want to use various maven versions between projects.

kukgini:
Furthermore, the development team of this project can control the maven version needed to build this source code.

PR #2100, opened 2020-02-19
IS-1498: name_type crate is now deprecated

Artemkaaas:


libindy/src/domain/anoncreds/credential.rs
```diff
@@ -14,7 +13,7 @@ use super::credential_definition::CredentialDefinitionId;
 use super::revocation_registry_definition::RevocationRegistryId;
 use super::schema::SchemaId;
 
-#[derive(Debug, Deserialize, Serialize, NamedType)]
```
AxelNennker: I suggest to add more tests in addition to the one for WalletRecord. I think it would be good to have tests for some of those structs that don't have the `#derive(NamedType)`anymore.

Artemkaaas: I've thought about it but all these types are defined in libindy/domain (none of them neither moved to indy-api-types yet). And I don't want to expose `short_type_name` function from indy-wallet sub crate.

PR #2104, opened 2020-02-21
adds generic agent information mechanism

ryMarsh44:


vcx/libvcx/src/credential.rs
```diff
@@ -70,12 +66,7 @@ pub struct Credential {
     credential_offer: Option<CredentialOffer>,
     msg_uid: Option<String>,
     // the following 6 are pulled from the connection object
-    agent_did: Option<String>,
```
Artemkaaas: It breaks compatibility with previously serilized objects. I think we should add a new version of object.
Now the sequence of steps: `deserilize` -> `update_state` will fail.

vcx/libvcx/src/connection.rs
```diff
@@ -96,6 +97,99 @@ struct Connection {
     version: Option<settings::ProtocolTypes>
 }
 
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq)]
+pub struct MyAgentInfo {
```
Artemkaaas: Do we really need this complexity of getting `AgentInfo`.
Why can't we just add a single function that returns the whole structure at once?

ryMarsh44: There are two sets of information which are known at different phases.
1. Information known after provision: (agency_did, agency_vk, agent_did, agent_vk) 
2. The rest is information known after a pw connection. 

There are time where someone may want the agent information known only after provisioning and another where they want the entire set after a connection is established.

PR #2112, opened 2020-02-27
Webhook update

mirgee:
Fix `vcx_update_webhook_url` to update the notification webhook settings not only locally, but also on the agency. Update nodejs wrapper implementation to support the new version by making the call asynchronous.


mirgee:
"Everynym guy's" feedback I received via @Artemkaaas which I based the last two commits on:

> UPDATE_CONFIGS is only used for enterprise's 'name' and 'logo url' (even though that api may accept any string as key and value, it won't be utilized for any other purpose
if you want to use UPDATE_COM_METHOD to register http endpoint, then the 'type' should be 2 (for push notification, the type is 1)
```
{
   @type: {name: "UPDATE_COM_METHOD", "version": "1.0"},
   "comMethod": {
      "id": ...,
      "type": 2,
      "value": 
   }
}
```

vcx/libvcx/src/api/vcx.rs
```diff
@@ -308,15 +308,50 @@ pub extern fn vcx_update_institution_info(name: *const c_char, logo_url: *const
...
+/// notification_webhook_url: URL to which the notifications should be sent
+///
+/// cb: Callback that provides error code of the result
+///
+/// #Returns
+/// Error code as u32
 #[no_mangle]
-pub extern fn vcx_update_webhook_url(notification_webhook_url: *const c_char) -> u32 {
-    info!("vcx_update_webhook >>>");
+pub extern fn vcx_update_webhook_url(command_handle: CommandHandle,
```
jovfer, line 324: Please pay attention that this changes are "breaking changes" from API point of view. I think they are reasonable in this particular place as this functionality is pretty new, but we should be careful with breaking changes in general.

mirgee: OK, point taken.

jovfer:
@Patrik-Stas ok, thanks for clarifications. I'm fine with both options

vcx/libvcx/src/api/vcx.rs
```diff
@@ -153,6 +153,19 @@ fn _finish_init(command_handle: CommandHandle, cb: extern fn(xcommand_handle: Co
                 cb(command_handle, e.into());
             }
         }
+
+        match settings::get_config_value(settings::CONFIG_WEBHOOK_URL) {
+            Ok(webhook_url) => match ::messages::agent_utils::update_agent_webhook(&webhook_url) {
+                Ok(()) => debug!("Agent webhook url updated on init"),
```
Patrik-Stas, line 159: - Should call the callback here. 
```
cb(command_handle, error::SUCCESS.code_num);
```
Consequently you also have:
- remove callback from open_wallet success branch
- add return statement to open_wallet error branch

- Also I'd suggest to call `info!` with information about the url which has been registered.

vcx/libvcx/src/api/vcx.rs
```diff
@@ -153,6 +153,19 @@ fn _finish_init(command_handle: CommandHandle, cb: extern fn(xcommand_handle: Co
...
+                Err(e) => {
+                    error!("Error updating agent webhook on init: {}", e);
+                    cb(command_handle, e.into());
+                    return Ok(());
+                }
+            }
+            Err(e) => debug!("Error reading agent webhook from settings: {}", e)
```
Patrik-Stas, line 166: - If `webhook_url` is not present in the `agent provision` passed to `vcx_init`, `get_config_value` will return Error. So we should just in info/debug that webhook url will not be updated since it was not provided. 
- this match branch should also call success callback `cb(command_handle, e.into());`

vcx/libvcx/src/api/vcx.rs
```diff
@@ -144,13 +144,28 @@ fn _finish_init(command_handle: CommandHandle, cb: extern fn(xcommand_handle: Co
...
+                    cb(command_handle, error::SUCCESS.code_num);
+                }
+                Err(e) => {
+                    error!("Error updating agent webhook on init: {}", e);
+                    cb(command_handle, e.into());
+                }
+            }
+            Err(e) => {
+                debug!("webhook_url was not updated in agency: {}", e);
+                cb(command_handle, e.into());
```
Patrik-Stas, line 168: This should be
```
cb(command_handle, error::SUCCESS.code_num);
```
otherwise we'd be throwing errors to anyone who doesn't set webhook_url in agent provision.

vcx/libvcx/src/messages/agent_utils.rs
```diff
@@ -432,6 +462,46 @@ fn update_agent_info_v2(to_did: &str, com_method: ComMethod) -> VcxResult<()> {
     Ok(())
 }
 
+pub fn update_agent_webhook(webhook_url: &str) -> VcxResult<()> {
+    trace!("update_agent_webhook >>> webhook_url: {:?}", webhook_url);
+
+    let to_did = settings::get_config_value(settings::CONFIG_REMOTE_TO_SDK_DID)?;
+
+    let com_method: ComMethod = ComMethod {
+        id: String::from("123"),
```
Patrik-Stas, line 471: What's `id` and why it's `123`?

vcx/libvcx/src/messages/agent_utils.rs
```diff
@@ -97,6 +100,33 @@ pub struct UpdateComMethod {
     com_method: ComMethod,
 }
 
+#[derive(Debug)]
+pub enum ComMethodType {
+    A2A,
+    Webhook
+}
+
+impl Serialize for ComMethodType {
```
Patrik-Stas, line 109: Can you add unit tests for this custom de/serialization?

vcx/libvcx/src/messages/agent_utils.rs
Patrik-Stas, line 471: Was `123` previously used as id of update comm message as well?

Patrik-Stas: Oh okay now I see that in `pub extern fn vcx_agent_update_info` which is also building and sending `ComMethod` structure, it's using `id` passed by user in argument and I haven't found it documented.
@jovfer is it okay to set here such a dummy value like 123? Is `id` only relevant in case of `com_method.e_type == 1`?

jovfer: @ryMarsh44 @Artemkaaas could you take a look on this question?

ryMarsh44: @jovfer I do not know for certain but as far as I can remember, the field `id` is not actually used in any of the agent implementations I've seen. I believe "123" is a default value.

Patrik-Stas:
@jovfer based on Ryan's answer, is this is good to go? Dummy cloud agency is also not utilizing `id`, just reads it and writes it back to response. Are there any other concerns regard this PR?

jovfer:
@Patrik-Stas based on Ryan's answer I would say "no other concerns, fill free to merge the PR"

PR #2113, opened 2020-02-28
rename wallets because test failed

AxelNennker:

the test failed with `wallet with that name already exists` so this PR names issuer and prover wallet. 

I am kind of astonished why this ever worked. Maybe I am confuse. 
Anyway all tests succeed now

### DIFF
--- a/libindy/tests/anoncreds_demos.rs
+++ b/libindy/tests/anoncreds_demos.rs
@@ -500,10 +500,10 @@ mod demos {
         Setup::empty();
 
         //1. Issuer creates wallet, gets wallet handle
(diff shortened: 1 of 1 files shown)


jovfer:
@AxelNennker `create_and_open_default_wallet` helper appends number suffix to the name parameter. The error `wallet with that name already exists` may indicates that previous run was interrupted or failed and hasn't cleaned wallets as a result

AxelNennker:
... or that the cleanup after the tests did not work

Anyway, shouldn't issuer and prover wallet have different names?

jovfer:
In master version they would have names like `default-wallet_id-anoncreds_works_for_revocation_proof_issuance_by_demand-1` and `default-wallet_id-anoncreds_works_for_revocation_proof_issuance_by_demand-2` 
https://github.com/hyperledger/indy-sdk/blob/master/libindy/tests/utils/wallet.rs#L84

PR #2138, opened 2020-03-20
Refactor vcx agency (stop modelling Router and Admin as Actix actors)

Patrik-Stas:
Having Router and Admin modelled as Actors doesn't seem to bring much benefit. It actually makes it cumbersome to add new APIs on these entities as:
1. They have to be modelled as messages
2. The messaging make all APIs asynchronous, even though they could be synchronous (for example registering DID in router)
Furthermore there's only 1 router and admin at all times. 

This is preparation to start working on agency stateless as I will need to adjust Router API and likely add new methods.


vcx/dummy-cloud-agent/src/actors/router.rs
```diff
@@ -25,11 +31,12 @@ use crate::utils::futures::*;
 pub struct Router {
     routes: HashMap<String, Recipient<HandleA2AMsg>>,
     pairwise_routes: HashMap<String, Recipient<HandleA2ConnMsg>>,
-    requester: Addr<Requester>
+    requester: Addr<Requester>,
+    fw_agent_wallet_handle: WalletHandle
```
Artemkaaas: why is wallet_handle need to the router?

Patrik-Stas: Thanks, indeed not needed, forgot to remove this. Pushing commit to remove this now.

PR #2144, opened 2020-03-27
Backmerge Indy 1.15.0

Artemkaaas:


vcx/libvcx/build.rs
```diff
@@ -88,6 +88,8 @@ fn main() {
 
     }else if target.contains("darwin"){
         //OSX specific logic
+        println!("cargo:rustc-link-lib=sodium");
```
jovfer: I'm in doubt about correctness of this changes.

PR #2150, opened 2020-04-03
Add test for IndySDK Node wrapper non-secrets

Patrik-Stas:
Adding missing coverage for non-secrets

Signed-off-by: Patrik <patrik.stas@absa.africa>


Patrik-Stas:
One question about this API, is it possible to store tags as integers and then use `$gte` operator for searches? It didn't seem to work for me - when I tried to save tags such as `{"score": 5}`, I would get back message:
```
 '{"backtrace":"","message":"Error: Invalid structure\\n  Caused by: Invalid Tags json has been passed\\n  Caused by: invalid type: integer `5`, expected a string at line 1 column 11\\n"}'
```
The documentation here https://github.com/hyperledger/indy-sdk/tree/master/docs/design/003-wallet-storage#non-secrets-api seems to distinquish between `<str>` tags and `<int>` tags.
I tried to store numeric tags as string like `{"score": "5"}`, the "$gte" search operator didn't compare values numerically.

jovfer (approved):

PR #2157, opened 2020-04-22
Secure connection to postgres via SSL

mirgee:
If the user specifies that they require / prefer to use SSL secured connection in the storage config, the postgres plugin tries to connect to the postgres server using the certificate provided by the user in the config. If the certificate was not issued by a trusted CA, postgres rejects the connection.

### DIFF
--- a/experimental/plugins/postgres_storage/Cargo.lock
+++ b/experimental/plugins/postgres_storage/Cargo.lock
@@ -2,1287 +2,1774 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.7.6"
+version = "0.7.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8716408b8bc624ed7f65d223ddb9ac2d044c0547b6fa4b0d554f3a9540496ada"
 dependencies = [
- "memchr 2.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.3.3",
 ]
 
 [[package]]
 name = "arrayref"
-version = "0.3.5"
+version = "0.3.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


ianco:
I can't get this to build, I get:

```
   Compiling serde_json v1.0.53
   Compiling named_type_derive v0.1.6
error: failed to run custom build command for `openssl v0.9.24`

Caused by:
  process didn't exit successfully: `/Users/icostanzo/Projects/indy-sdk/experimental/plugins/postgres_storage/target/debug/build/openssl-e04b05c0bd462e35/build-script-build` (exit code: 101)
--- stderr
thread 'main' panicked at 'Unable to detect OpenSSL version', /Users/icostanzo/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-0.9.24/build.rs:16:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.

warning: build failed, waiting for other jobs to finish...
error: build failed
```

What version of OpenSSL do you have installed?  I'm on MacOS Catalina and I hve:

```
openssl version
OpenSSL 1.1.1g  21 Apr 2020
```

mirgee:
@ianco I am running Ubuntu 18.04.4 LTS and my openssl version is
``
openssl version
OpenSSL 1.1.1  11 Sep 2018
``
and Cargo version
``
cargo version
cargo 1.41.0 (626f0f40e 2019-12-03)
``

ianco:
I've had no success in building this update.  I think the root cause is the dependency on postgres `0.15.2`, which depends on an older version of openssl (`^0.9.24`):  https://crates.io/crates/postgres/0.15.2.  As far as I can tell this is not compatible with OpenSSL 1.1.1, and I haven't been able to downgrade OpenSSL to a version that will compile successfully with rust openssl.

I suggest a better approach is to upgrade the postgres crate to the latest `0.17.3`: https://crates.io/crates/postgres/0.17.3.  This updates the rust openssl dependency to `^0.10`.

PR #2160, opened 2020-04-24
Add VCX Java demo

baegjae:
This VCX Java Alice/Faber demo was created with reference to the VCX node demo.

Also, it contains small bug fix in VCX Java wrapper APIs.
- issuerSendCredentialCB - Remove credentialDefId parameter (incurred error)
- vcxProofGetStateCB - Remove proofHandle parameter (returned wrong state)


vcx/wrappers/java/src/main/java/com/evernym/sdk/vcx/issuer/IssuerApi.java
```diff
@@ -155,20 +155,20 @@ public void callback(int commandHandle, int err, int state) {
     }
     private static Callback issuerSendCredentialCB = new Callback() {
         @SuppressWarnings({"unused", "unchecked"})
-        public void callback(int commandHandle, int err, String credentialDefId) {
-            logger.debug("callback() called with: commandHandle = [" + commandHandle + "], err = [" + err + "], credentialDefId = [" + credentialDefId + "]");
-            CompletableFuture<String> future = (CompletableFuture<String>) removeFuture(commandHandle);
+        public void callback(int commandHandle, int err) {
```
Patrik-Stas, line 158: Note for other reviewers: this seems correct, see the callback definition here
https://github.com/hyperledger/indy-sdk/blob/master/vcx/libvcx/src/api/issuer_credential.rs#L196

vcx/wrappers/java/src/main/java/com/evernym/sdk/vcx/proof/DisclosedProofApi.java
```diff
@@ -123,8 +123,8 @@ public void callback(int commandHandle, int err, String proofRequests) {
 
     private static Callback vcxProofGetStateCB = new Callback() {
         @SuppressWarnings({"unused", "unchecked"})
-        public void callback(int commandHandle, int err, int proofHandle, int state) {
-            logger.debug("callback() called with: commandHandle = [" + commandHandle + "], err = [" + err + "], proofHandle = [" + proofHandle + "], state = [" + state + "]");
+        public void callback(int commandHandle, int err, int state) {
```
Patrik-Stas, line 126: Note for other reviewers: this seems correct, see the callback definition here
https://github.com/hyperledger/indy-sdk/blob/master/vcx/libvcx/src/api/disclosed_proof.rs#L438

Patrik-Stas:
Looks good to me 👍 Let's wait for @jovfer 's review as well.

baegjae:
@jovfer Currently, demo is not executed when using maven's `com.evernym:vcx:0.8.0`, so it is written to use the local vcx java wrapper (jar) temporarily. In my opinion, it would be better to merge this PR first. Next, as soon as this PR's bugfix applies to maven, I will register a PR that imports the vcx java wrapper from maven.

Patrik-Stas:
@baegjae I see it's common practice to commit gradlew files cvs, and the .jar file is just 50kb, but for sake of security, can you make binary file checksum as part of jenkins pipeline? 
See: https://docs.gradle.org/current/userguide/gradle_wrapper.html#wrapper_checksum_verification

baegjae:
@Patrik-Stas I applied the your request. If you have any further needs, please let me know.

Patrik-Stas:
@baegjae Thanks for update - however still missing the check part. I'd rather not have the checksum file as part of the repo - instead either in jenkins, or as a github action, I'd like to:
- download the checksum file for this gradlew version from `gradle.org`
- verify the checksum as part of the pipeline

I think actually this will be much easier to add as github action. Extending existing Jenkins pipeline can be quite difficult due to extremely long feedback loop.

kukgini:
Since VCX is forked and will be removed here, I think this PR should be dropped in this repository. https://github.com/hyperledger/indy-sdk/pull/2416

PR #2021, opened 2020-01-02
Update Node.JS VCX wrapper

saholman:
Updates deprecated dependencies for the Node.JS wrapper to work with Node.JS v10 and v12. Also statically links OpenSSL with libvcx to avoid problems when using different Node.JS versions.

PR #2171, opened 2020-05-19
Update VCX Nodejs wrapper node version to 12 

mirgee:
This PR is based on #2021. Bacause of [this issue](https://github.com/nodejs/node/issues/24964), the `openssl` library has be linked statically. Achieved via `vendored` feature.


jovfer:
@mirgee as far as I remember there were 2 blockers to migrate from 8 to 12 NodeJS
1) openssl system library - it should be already fixed in master by linking to the correct version of openssl and the migration were done partially (to v10)
2) ffi-napi issue was last blocker to do a migration from 10 to 12.

Let me check the status of #2021 

Usage of vendored feature may increase CI/CD time

mirgee:
@jovfer 

> openssl system library - it should be already fixed in master by linking to the correct version of openssl and the migration were done partially (to v10)

I was still experiencing the openssl issue with node 12 + ffi-napi, that's why I added the vendored feature, which fixed it...

> Usage of vendored feature may increase CI/CD time

How much would it slow the CI time? On my computer, libvcx build is ~7 s slower with vendored feature.

solidnerd:
Hey @jovfer and @mirgee, 
what is the current state of this pr? I would like to know when it will be merged.

PR #2176, opened 2020-05-24
WIP: Add github actions

Patrik-Stas:
Github actions pipeline prototype. 

- To work, `PACKAGES_ACCESS_TOKEN` secret needs to be set up on Github repo. That's because the pipeline as is is leveraging [Github Packages](https://github.com/features/packages) product for docker image persistence which is free for public Github projects. 
- Image layers are cached per pull request / branch. First commit to new branch will be the slowest, subsequent should be way faster.
- Docker buildkit is used - this makes it possible to use different dockerignore files per different dockerfile. That's important because when you modify libvcx, you wouldn't want to invalidate docker layer caches for libindy image. See https://github.com/moby/moby/issues/12886#issuecomment-533195318

- Job `assure-image-indypool` builds `indy_pool` image (reusing layer in GIthub Packages)
- Job `build-image-ubuntu16-libindy` builds image `ubuntu16-libindy` with compiled libindy (reusing layer in GIthub Packages). 
- Job `build-image-ubuntu16-libvcx` - builds image `ubuntu16-libvcx` with compiled libindy, libnullpay and libvcx
- Job `build-image-ubuntu16-vcxagency` builds image with dummy cloud agency binary.
- Job `test-libindy` runs unit tests for libindy against local pool
- Job `test-libindy-wrapper-rust` runs unit tests for libindy rust wrapper
- Job `test-vcxagency` runs unit tests for dummy cloud agency
- Job `test-libvcx` - runs unit tests for `libvcx` against local pool

- Job `build-libindy-run-unittests` compiles libindy on host Github Actions machine and runs unit tests. This is bit experimental, it might be bettter to run unit tests within image built in `build-image-ubuntu16-libindy`. The upside of this host-approach is that we can stay consistent when testing on MacOS and Windows as Github Actions offers hosts on OSX Catalina and Windows Server 2019 as well.

Missing:
- Python, Java, NodeJS Libindy wrapper testing
- Libvcx wrapper testing
- Builds on operating systems other than Ubuntu 16.04
- Probably some more? 

Also there's surely stuff to refactor, there's quite a bit of repetition in this GA definition right now.
As I realized doing this, minor but cool thing about Github Actions is that people could actually run pipeline in their own forks without doing PR first. Also, it's easier for someone to propose modifications to GA definition.


Patrik-Stas:
@jovfer there's still work left to do here, but can we experimentally merge this to enable collaboration on Github Actions? To have these actions working, secret `PACKAGES_ACCESS_TOKEN` needs to be set for this repo. 
I was thinking we could use   https://github.com/sovbot user to generate him an access token with  `read:packages, write:packages` rights, then use that for that secret. What do you think?

PR #2182, opened 2020-06-03
fix typo

ingee:
- Fix few typos in README
- Fix API url of dummy-cloud-agent's admin-API in README


dhh1128:
@ingee : this looks reasonable, but it will have to pass the [DCO (code-signing) requirement](https://probot.github.io/apps/dco/) before it can be merged.

Issue #2183, opened 2020-06-03
libindy always generate trace logs even if they are not going to be used


jovfer:
In libindy (up to 1.15.0 ver) the internal logging was always set up to trace
https://github.com/hyperledger/indy-sdk/blob/v1.15.0/libindy/src/utils/logger.rs#L121

It results in at least creation of serialized log entry and passing it to FFI for example.

Wrappers (e.g Java one https://github.com/hyperledger/indy-sdk/blob/v1.15.0/wrappers/java/src/main/java/org/hyperledger/indy/sdk/LibIndy.java#L291-L309 ) has logic to avoid final appearance of a log entries at the end, but creating of full logs for large data is extremely inefficient.

This problem is applicable only to custom logger and doesn't affect default logger.

PR #2184, opened 2020-06-03
[gh-2183] Fix libindy and java wrapper.

jovfer:
This PR optimizes libindy and Java wrapper behavior in scope of #2183

### DIFF
--- a/libindy/indy-utils/src/ctypes.rs
+++ b/libindy/indy-utils/src/ctypes.rs
@@ -248,3 +248,14 @@ macro_rules! check_useful_opt_u64 {
         };
     }
 }
+
+#[macro_export]
+macro_rules! check_u32_less_or_eq {
+    ($x:ident, $lim:expr, $e:expr) => {
+        let $x: u32 = if $x <= $lim {
+            $x
+        } else {
+            return err_msg($e.into(), format!("Invalid integer has been passed (should be non-negative and less or equal to {}", $lim)).into()
+        };
+    }
+}
--- a/libindy/src/api/logger.rs
+++ b/libindy/src/api/logger.rs
@@ -5,6 +5,7 @@ use indy_api_types::errors::prelude::*;
 
 use crate::utils::logger::{EnabledCB, LogCB, FlushCB, LibindyLogger, LibindyDefaultLogger, LOGGER_STATE};
 use indy_utils::ctypes;
+use log::LevelFilter;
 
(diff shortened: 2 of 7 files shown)


jovfer:
@dhh1128 thanks for feedback

> Re. #1: A "max" log level doesn't make sense. When you filter logs, you filter the _min_ log level, answering the question, "What is the smallest/most minor/most granular type of event you're willing to include?" You start by suppressing first trace, then debug, then info, then warning, then error, then fatal. This change does it backwards, treating "trace" as max and "error" as min. Surely we don't intend to implement logging such that you can suppress errors but include debug messages?

well, I see you arguments, but this is terminology / naming used by Rust default log crate (https://docs.rs/log/0.4.8/log/#modules ). So I've just re-used it without any criticism. I think the direction of the order is ambiguous. Just re-phrase your question as "what maximum level of details you would like to see?" ("the most detailed" == "the least significant/critical").
Like in Python the 0 is all 10 debug+ and 50 is critical. In rust 0 is none, 1 is error, 5 is trace. Personally I like rust's variant more as it's native to me to threat 0 as "disabled". The consequence of this idea is to have 1 as error and 5 as trace. As a result "max" is setting of the maximum level which will appears in the log.

So in case of rust crate library my opinion is keep default terminology, so no changes to current PR.
As soon as libindy is rather generic one and it provides FFI API, probably we have an option to avoid both "max" and "min" at all. And use just "set_log_level" like sl4j But  I'd rather suggest to keep current one as there is a chance of evolving this codebase into rust crate in scope of Aries evolution.

dhh1128:
Okay. I am convinced that my feedback about "max" is incorrect. Please ignore that and keep the terminology you currently have.

jovfer:
@dhh1128 I didn't mean that I have strong opinion about preferred option. Let me put it in this way: I'm in doubt between `indy_set_log_level` and `indy_set_log_max_lvl`. But if you find some of arguments listed by me as valid ones - ok, lets keep the current naming.

as for 2nd point, I think I'm missing some details, could you elaborate please? 
I'm expecting the following: default log level in tests is `trace`. It would be applied everywhere except new one module `tests/logger_lvl.rs`. And this module has special logger implementation so no trace would be visible in `stdout` or `stderr`. Instead the logger implementation here would gather minimal statistic and check some assertions about these numbers.

Does this description make sense and does it match to what are you observing?

PR #2187, opened 2020-06-08
VCX iOS wrapper updates

airoasis:
Added APIs
* connectionGetState, connectionUpdateState, connectionRelease
* credentialRelease
* proofGetRequests, proofGetState, proofUpdateState, proofRelease

Fixed 
* Fixed the parameter name of vcx_disclosed_proof_get_requests method (offers --> requests)

There are several data types for handles (connectionHandle, credentialHandle, proofHandle), and they are used Interchangeably. For example, for the connectionHandle, three data types (VcxHandle, NSInteger, vcx_connection_handle_t) are used in the code... I believe that there should be a consistent way of implementing it. In this PR, I only use NSInteger for all handles.


Patrik-Stas:
@jovfer all good on our side, I've set you as reviewer as well for sake of cross organizational review practices

PR #2198, opened 2020-06-17
Use correct path for OpenSSL in local Mac build script

tmarkovski:

Update build script for Mac to use correct open ssl path. By default, MacOS comes with preinstalled open ssl at the location `/usr/local/Cellar`, but the brew package previously in this script installs updated open ssl at different location. This causes libindy not to load correctly.

### DIFF
--- a/libindy/mac.build.sh
+++ b/libindy/mac.build.sh
@@ -41,10 +41,7 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
     export CARGO_INCREMENTAL=1
     export RUST_LOG=indy=trace
     export RUST_TEST_THREADS=1
-    for version in `ls -t /usr/local/Cellar/openssl/`; do
-        export OPENSSL_DIR=/usr/local/Cellar/openssl/$version
-        break
-    done
+    export OPENSSL_DIR=/usr/local/opt/`ls /usr/local/opt/ | grep openssl | sort | tail -1`
     cargo build
     export LIBRARY_PATH=$(pwd)/target/debug
     cd ../cli


libindy/mac.build.sh
```diff
@@ -41,8 +41,8 @@ if [[ "$OSTYPE" == "darwin"* ]]; then
     export CARGO_INCREMENTAL=1
     export RUST_LOG=indy=trace
     export RUST_TEST_THREADS=1
-    for version in `ls -t /usr/local/Cellar/openssl/`; do
-        export OPENSSL_DIR=/usr/local/Cellar/openssl/$version
+    for version in `ls -t /usr/local/opt/ | grep openssl | sort`; do
```
Patrik-Stas, line 44: Seems to work fine, but I think could be simplified - how about this?
```
export OPENSSL_DIR =`ls -t /usr/local/opt/ | grep openssl | sort | head -1`
```

Patrik-Stas: Hmh, but your changes (and my sugguested equivalent of your suggested changes) are changing logic a bit.
Previously it was selecting the most recently modified (presumaly latest installed I guess?) openssl. 
Now the `-t` on `ls` could be dropped because anyway it's reshuffled when doing `sort`
```
ls  /usr/local/opt/ | grep openssl | sort
```
So it's select first one alphabetically. In my case I have these:
```
ls  /usr/local/opt/ | grep openssl | sort                                                                                                                                                                          ✔  11847  09:53:30
openssl
openssl@1.1
```
Whereas openssl symlinks to `openssl@1.1`. Maybe it'd be better to select alphabetically the highest, eg:
```
ls  /usr/local/opt/ | grep openssl | sort -r                                                                                                                                                                       ✔  11848  09:53:33
openssl@1.1
openssl
```
Or alternatively just hardcode `/usr/local/opt/openssl` if it's guaranteed to exists and link to some openssl libraries?

tmarkovski: The script already installs the latest by running `brew install openssl`. Using `ls  /usr/local/opt/ | grep openssl | sort | tail -1` seems to select the latests

tmarkovski: I've updated the logic per your suggestion.

PR #2200, opened 2020-06-18
VCX Aries: should not update status of messages on agency

Artemkaaas:
In the current state, LIbvcx stores state objects into in-memory HashMap and return integer handles as identification. 
So applications that use Libvcx have to care about the persistence of state objects.  
Because of that possible situation when VCX processes some message -> update its status on agency -> return result to the app -> app failed before storing new VCX object state. It will make the message unavailable anymore.
Suggestion: VCX should update the status of messages internally. The application above should care about it.


Patrik-Stas:
This does make sense for `vcx_*_update_state_with_message` function, because the caller is working directly with messages and probably desires more control. So yeah, why not to also let him handle state in agency 👍 

However if you remove builtin agency message state updates for `vcx_*_update_state` functions, not only it's breaking change, but seems to me it would be difficult to use  `vcx_*_update_state`  functions. So with these changes, I think it might as well make sense to remove `vcx_*_update_state` altogether and only keep `vcx_*_update_state_with_message`.

Or alternatively I'd suggest that:
-  `vcx_*_update_state` would still update state in agency - comfortable way of progressing libvcx state machines, but less efficient with less control.
- `vcx_*_update_state_with_message` - caller has full control, passes messages and is also responsible to update states in agency (as this PR introduces)

Artemkaaas:
> Or alternatively I'd suggest that:
> 
> * `vcx_*_update_state` would still update state in agency - comfortable way of progressing libvcx state machines, but less efficient with less control.
> * `vcx_*_update_state_with_message` - caller has full control, passes messages and is also responsible to update states in agency (as this PR introduces)

Agree. That option is better.

PR #2210, opened 2020-06-26
Fix c_int usage in python wrapper

andrewwhitehead:
The default value for these parameters was defined as a c_int rather than a c_int64, which may have a different size depending on the platform.


andrewwhitehead:
I believe this may be the cause of some errors we were seeing on the raspberry pi, but that's not confirmed yet.

eorituz:
It caused the issues I had on the raspberry pi.
Namely the 
```
build_get_acceptance_mechanisms_request() => Err(IndyError { inner:
timestamp and version cannot be specified together.
Invalid structure })`

PR #2148, opened 2020-04-02
Add revocation to libvcx V3

mirgee:
Currently, credential revocation is not implemented in VCX V3. This PR implements credential revocation using anoncreds 1.0.

- The implementation preserves previous behaviour. If revoked credential is used to construct proof, the proof can be still progressed such that  `vcx_proof_get_state` will return `Accepted / 4`.
To verify that proof is valid and revoked credentials were not used, `vcx_get_proof` needs to be called and checked for `proof_state`. (`0=Undefined, 1=Verified, 2=Invalid`)
- Extends NodeJS LibVCX example by revocations
- To verify non-happy path in Verifier's state machine, when provided proof is not valid (see test `test_prover_handle_invalid_presentation_message`),  new setting `MOCK_INDY_PROOF_VALIDATION ` was introduced. I am not sure if there's better way to do this?
- the `FinishedState` state of Verifier's state machine wraps revocation-related data in struct `RevocationInfoV1` to make the state machine future proof ( when anoncreds v2 come out)
- Update on libvcx cargo dependencies

PR #2216, opened 2020-07-13
Add revocations to LibVCX Aries

Patrik-Stas:
This is https://github.com/hyperledger/indy-sdk/pull/2148 with some additional changes; made from Absa's fork.

This PR brings the capability to revoke credential when running in V2+Aries / V3 / V4 

- The implementation preserves previous behaviour. If revoked credential is used to construct proof, the proof can be still progressed such that  `vcx_proof_get_state` will return `Accepted / 4`.
To verify that proof is valid and revoked credentials were not used, `vcx_get_proof` needs to be called and checked for `proof_state`. (`0=Undefined, 1=Verified, 2=Invalid`)
- Extends NodeJS LibVCX example by revocations
- To verify non-happy path in Verifier's state machine, when provided proof is not valid (see test `test_prover_handle_invalid_presentation_message`),  new setting `MOCK_INDY_PROOF_VALIDATION ` was introduced. I am not sure if there's better way to do this?
- the `FinishedState` state of Verifier's state machine wraps revocation-related data in struct `RevocationInfoV1` to make the state machine future proof ( when anoncreds v2 come out)
- Update on libvcx cargo dependencies


Patrik-Stas:
Sorry for bit of a mess - let's keep the [original PR over here](https://github.com/hyperledger/indy-sdk/pull/2148)

PR #2218, opened 2020-07-25
Vcx revoke python

krgko:
The revocation feature at PR #2148 has been merged already but, does not implement in Python3 part.

I have modified some of the libraries, updated the demo to support the feature, and tested the flow already.


vimmerru (approved):
Can be merged in case of green pipeline

PR #2220, opened 2020-07-29
VCX - Make a comment field as an optional

airoasis:
According to aries rfcs 
https://github.com/hyperledger/aries-rfcs/tree/master/features/0036-issue-credential
https://github.com/hyperledger/aries-rfcs/tree/master/features/0453-issue-credential-v2
the comment field is an optional field. 
However, libvcx expects this as a mandatary, so if the comment is empty it causes Json parsing error 
```
Caused by: Cannot deserialize A2A message: missing field 'comment'
```
This PR fixes this issue.

### DIFF
--- a/vcx/libvcx/src/v3/messages/issuance/credential.rs
+++ b/vcx/libvcx/src/v3/messages/issuance/credential.rs
@@ -12,7 +12,8 @@ use std::convert::TryInto;
 pub struct Credential {
     #[serde(rename = "@id")]
     pub id: MessageId,
-    pub comment: String,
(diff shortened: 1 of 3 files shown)


dhh1128:
@Artemkaaas , do you agree this should be merged?

Patrik-Stas:
@Artemkaaas 
although this PR is correct, just heads up about something we run into. This break compatibility between libvcx issuer with this PR merged and holder running older libvcx, without this PR merged. Issuer will issue cred offer without comment, but older libvcx clients will obviously expect it and fail to parse the message.
So issuer with new libvcx version **must** set comment in order not to break older libvcx clients.

PR #2222, opened 2020-07-30
Issue: 2221 - adds test and code change to not enforce timestamps if …

ryMarsh44:
…non_revocation_interval={}

Signed-off-by: Ryan Marsh <ryan.marsh44@gmail.com>


libindy/src/services/anoncreds/helpers.rs
```diff
@@ -117,7 +117,7 @@ pub fn get_non_revoc_interval(global_interval: &Option<NonRevocedInterval>, loca
 
     trace!("get_non_revoc_interval <<< interval: {:?}", interval);
 
-    interval
+    interval.filter(|x| x.to.is_some() || x.from.is_some())
```
Artemkaaas, line 120: Could you move `.filter(...` before the logging of resulting?

libindy/src/services/anoncreds/verifier.rs
```diff
@@ -625,6 +625,19 @@ mod tests {
...
+        let rev_reg_defs: HashMap<RevocationRegistryId, RevocationRegistryDefinitionV1> = HashMap::new();
+        let rev_regs: HashMap<RevocationRegistryId, HashMap<u64, RevocationRegistryV1>> = HashMap::new();
+        let rc = Verifier::new().verify(
+            &proof, &req, &schema, &creds, &rev_reg_defs, &rev_regs
+        ).unwrap();
+    }
```
Artemkaaas, line 639: I think it'll be simpler to create a unit-test for `get_non_revoc_interval` function.
And either extend one of existing `anoncreds_demos` integration tests or provide a new one covering that case to avoid hard-coding so big data just for one unit-test.

PR #2223, opened 2020-08-03
implement ability to generate a W3C-compatible presentation from an anoncreds proof

dhh1128:
This is not a complete feature. However, it is a good beginning to a feature. It has a meaningful test that does substantial checking for correctness, but it may not be exhaustive enough. Turning it into a complete feature will require filling out the get_derived_cred_attribs() function in w3c.rs and implementing something on the verification side (either directly verifying a w3c preso, or converting a w3c preso back into an anoncreds proof so it can be verified by our existing features).

The correspondence between an anoncreds proof and a w3c preso can be extrapolated from the checked-in unit test, but also from https://docs.google.com/document/d/1ntLZGMah8iJ_TWQdbrNNW9OVwPbIWkkCMiid7Be1PrA/edit# (and its underlying diagram at https://docs.google.com/drawings/d/1nkbGe-Z51IeAu-QTe33tHY6rLfoJay4ObHzqqooM604/edit).

It is possible that the code will not compile as checked in. If so, it is almost certainly because I had to do a dirty hack of Ursa to make public some private members of the Proof data structure or its children. I don't think we want to make that change in Ursa into a permanent one; I left a comment in the code about how to get the data we need in a more flexible way.


jonathan-rayback:
Please note corrected links to documents similar to those referred to in the main comment. The original documents are no longer available.
Description: https://drive.google.com/file/d/18VqTBs0R2iXR92aKxACAh-D5JI3-PwG-/view?usp=sharing
Diagram: https://docs.google.com/presentation/d/1tusy8R-jpbr-z6b8MVAehY0FG_eUC-_-unrAXjnqZUM/edit?usp=sharing

dhh1128:
Thank you, @jonathan-rayback , for posting the links to the updated docs. They are approximately what was originally linked. I think the diagram has a slightly more evolved version, somewhere in the great bitbucket in the sky, but what we have is enough to communicate the main ideas.

Some further detail on why I said my code included a "dirty hack" against Ursa:

The Ursa data structures that hold the pieces of info that we have to map into a VP are marked `private` in Rust. This means external code isn't supposed to be able to see them. For my proof of concept, I recompiled Ursa with those pieces of data being marked public. But this is not really the right way to get access. We should either write methods on the Ursa data structures that publish just those pieces of data that need to be assembled, or write methods in Ursa that fully transform data structures to something different and public that contains exactly what a VP needs, or we should get the data we need from the serialized JSON version of those data structures (which does contain the pieces of data we want, but would require a lot of dynamic walking of the JSON).

swcurran:
This looks very interesting.  In discussions at Hyperledger Global Forum this week, we talked about making AnonCreds a separate project on its own -- separate from Indy (one of many places on which AnonCreds can be rooted) and separate from Aries (Aries is VC format agnostic).  We are thinking the project will include this:

1. The specification that we've been working on (https://anoncreds-wg.github.io/anoncreds-spec/). Evidently, Hyperledger is more open to including JDF-based spec. work in an HL Project, so the move would not be bad.  Love to get some help on the spec.
2. At least one implementation of a "ledger agnostic" version of AnonCreds -- likely based on the "CredX" implementation in https://github.com/hyperledger/indy-shared-rs.
3. Defining the interface to enable "AnonCreds VDR Methods". These are "methods" in the same sense as DID Methods -- where the registration and resolution of AnonCreds objects is handled as needed for specific ledgers.  This would be the part of the spec. where the reading/writing of the AnonCreds objects is defined.
4. Open source implementations of AnonCreds VDR Methods.  We have now heard about 6 different ways of publishing and retrieving AnonCreds objects by members of the community (Legacy Indy, cheqd, Fabric, Database, Files and "did:indy"). We would want formal implementations of each, and more can be added -- e.g. love to see Besu for example.
5. A way to generate a W3C VC from an AnonCreds proof.
6. Discussions to evolve AnonCreds in a way that we retain all of the features we have today.  We'd start with revocation, since that's the scariest bit, but get going on discussions to consider removing CL-Signatures.

Interested in joining with this effort?

PR #2229, opened 2020-08-25
Update postgres index to use md5

jsaur:
Updates ix_tags_encrypted_value to use md5 so it can support larger values (eg images)
FYI @ianco 


brentzundel:
I recognize that in this particular instance, the cryptographic properties of the hash function used may not be entirely pertinent, but MD5 is terrible in many, many ways and I would hate to see it used here. Would it be a problem to change this to SHA256?

jsaur:
> I recognize that in this particular instance, the cryptographic properties of the hash function used may not be entirely pertinent, but MD5 is terrible in many, many ways and I would hate to see it used here. Would it be a problem to change this to SHA256?

Sure, looks like sha256 works just as well for me

swcurran:
Looks like this has broken some things.  See this issue #2231

PR #2232, opened 2020-09-03
Switch back to md5 to support older postgres dbs

jsaur:
sha256 isn't supported by postgres 10 which bc gov uses, so switching back to md5


ianco:
Tested with postgres 10 (with bc gov apps) with no issues

jovfer (approved):
Compatibility with older version of Postgres is a valid argument.
FYI @brentzundel

PR #2250, opened 2020-10-15
Update README.md

### LINKED ISSUE #2249: listMyDidsWithMeta API specification is different from implementation
As I pointed out in #2238 , listMyDidsWithMeta API returns json which includes tempVerkey field but  https://github.com/hyperledger/indy-sdk/blob/master/wrappers/nodejs/README.md does not say so.

toshirin33:
add tempVerkey field to the return of listMyDidsWithMeta to fix #2249


toshirin33:
I am sorry that I mistook the operation when adding sign-off. Some records in this pull request have disappeared.

PR #2252, opened 2020-10-18
Adding android wrapper library

blazsolar:
Majority of this is just a copy of java wrapper implementation with some changes. Mainly support for api level 21 with `java9.util.concurrent` and making test work on Android emulator.

An example will be contributed in a separate PR.


jovfer:
@blazsolar Thank you for the contribution! 

> Majority of this is just a copy of java wrapper implementation with some changes.

If so, is it possible to merge Android and Java wrappers?

blazsolar:
In theory, yes, but it is not as straightforward as just merging them together. First thing is that the java module at the moment uses maven build system and android uses gradle. I would be conferable migrating java module to gradle but not the other way around as I am not as familiar with maven. 

The second thing is the actual changes. We would have to change the source/bytecode at compile time from `java.util.concurrent` to `java9.util.concurrent` when compiling for Android as it doesn't make sense to use backward compatibility feature for java wrapper. Android gradle plugin has extension points that we could use to achieve something like this but it would require some effort. On the other hand, google is working on backwards compattibility for `java.util.concurrent` out of the box. I can't find the actual bug report for it but i'm certan that there is one. 

Third thing is that I would like to migrate Android implementation from JNA to JNI. The reason for that is we at the moment library ships with `libc++_shared.so` file included. That may cause issues for the consumer of the library in cases where they already add `libc++_shared.so` file themself. With JNI we can avoid including that flile in our library. More on that https://android.googlesource.com/platform/ndk/+/refs/heads/master/docs/user/middleware_vendors.md#for-java-middleware-with-jni-libraries 
This would mean a bit more maintainenc, as bridge files betwen native and java have to be written manually, but not that much. This change could also be included for both android and java library. So this shouldn't be a blocker to merge libraries together in my oppinion.

There are a bit more changes in tests, but nothing that couldn't be made working on both platforms. 

I'm willing to merge the 2 together but it would take some time and I would do it with gradle build system.

blazsolar:
I found the issue about `java.util.concurrent` backporting. https://issuetracker.google.com/issues/153794291 Priority just changed from P3 to P2.

blazsolar:
The `java.util.concurrent` has now been backported.

https://github.com/google/desugar_jdk_libs/blob/master/CHANGELOG.md#version-120-2022-06-21

PR #2254, opened 2020-10-19
iOS build updates

blazsolar:
Few updates to the iOS build process.

Get `OPENSSL_DIR` value directly from brew if not otherwise specified. Instructions dictate that you install OpenSSL through brew so it makes sense to have that as a default value. The developer is still able to overwrite this.

Few minor updates to `README.md`

`universal_framework.sh` was using "iPhone 6" simulator when building wrapper. From Xcode 11 onwards this simulator is no longer preinstalled and compilation would fail if the simulator was not manually installed. The script was updated to use "iPhone 11" simulator as it is preinstalled on XCode 11 and newer.


ianco:
@blazsolar I've tried doing a local iOS build with these updates and am getting errors, looks like it's not finding the arm64 build for openssl?

```
+ cargo lipo --release
[INFO  cargo_lipo::meta] Will build universal library for ["libindy"]
[INFO  cargo_lipo::lipo] Building "libindy" for "aarch64-apple-ios"
   Compiling libindy v1.16.0 (/Users/icostanzo/Projects/indy-sdk/libindy)
error: linking with `cc` failed: exit code: 1
...
  = note: ld: warning: ignoring file /usr/local/Cellar/zeromq/4.3.2/lib/libzmq.dylib, building for iOS-arm64 but attempting to link with file built for macOS-x86_64
          ld: warning: ignoring file /usr/local/opt/openssl@1.1/lib/libcrypto.dylib, building for iOS-arm64 but attempting to link with file built for macOS-x86_64
          ld: warning: ignoring file /usr/local/opt/openssl@1.1/lib/libssl.dylib, building for iOS-arm64 but attempting to link with file built for macOS-x86_64
          ld: warning: ignoring file /usr/local/Cellar/libsodium/1.0.18_1/lib/libsodium.dylib, building for iOS-arm64 but attempting to link with file built for macOS-x86_64
          Undefined symbols for architecture arm64:
            "_BN_new", referenced from:
                openssl::bn::BigNum::new::h0ebc0e3b623105c7 in libopenssl-a056ae0ccb40469d.rlib(openssl-a056ae0ccb40469d.openssl.5uhcjdkl-cgu.4.rcgu.o)
                openssl::bn::BigNum::from_u32::hc5d327ef8b7147f0 in libopenssl-a056ae0ccb40469d.rlib(openssl-a056ae0ccb40469d.openssl.5uhcjdkl-cgu.4.rcgu.o)
... etc and a bunch more errors ...
```

ianco:
... and the CI test is failing on this PR:

```
  --- stderr

  thread 'main' panicked at 'OpenSSL library directory does not exist: /usr/local/Cellar/openssl@1.1/1.1.1g/lib', /Users/jenkins/.cargo/registry/src/github.com-1ecc6299db9ec823/openssl-sys-0.9.53/build/main.rs:52:9

  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

warning: build failed, waiting for other jobs to finish...

error: build failed

[ERROR cargo_lipo] Failed to build "libindy" for "x86_64-apple-ios": Executing "/Users/jenkins/.rustup/toolchains/1.46.0-x86_64-apple-darwin/bin/cargo" "--color" "auto" "build" "-p" "libindy" "--target" "x86_64-apple-ios" "--lib" finished with error status: exit code: 101

script returned exit code 1
```

vimmerru:
@ianco 

>  I've tried doing a local iOS build with these updates and am getting errors, looks like it's not finding the arm64 build for openssl?

Seems you are using Rust 1.46. It forces building of `cdlyb` crate type and it is why you see linker warnings. On iOS we need only `staticlib` that don't need any binary dependencies, but need headers only. It allows to use openssl and zmq from brew.

This behavior will be fixed in Rust 1.49. For now our pipeline uses 1.45. So i believe we don't have problems in this PR and should merge it.

vimmerru (approved):

blazsolar:
@ianco I have done a test build locally with Rust 1.49 and it works fine.

Regarding CI not finding OpenSSL. I have dug a bit deeper into this and it looks like there is a reason OpenSSL was hardcoded to 1.0.2 on CI. There are 2 pods generated by the build process, `libindy` pod that depends on openssl installed via brew and `libindy-objc` pod that depends on other pods to get openssl dependencies in. The OpenSSL pod was not updated for 4 years now and the last version contains OpenSSL 1.0.2. That explains why CI was hardcoded to 1.0.2. When developers are trying to compile this locally, they would usually have newer version of OpenSSL installed through brew if they follow instructions and `libindy-objc` would fail to compile as some symbols would be missing in 1.0.2 that were added in the latest OpenSSL.

I suggest doing the following changes. Migrating from OpenSSL pod to OpenSSL-Universal pod, since first one is not maintained any more and lather one looks like a defacto standard for iOS development.

Update `libindy` pod build process to use prebuild binaries and headers from OpenSSL-Universal pod. They provide prebuilds as part of their build so simple script could be created that would download them and then used them in build process.

Simmilar should probably also be done for `libsodium` and `zeromq` pods. But extra work would be required as pods do not provide prebuild binaries.

Does that sound like a reasonable approach or does anyone have any better ideas how to tackle this?

PR #2255, opened 2020-10-20
VE-2045: add an api call for collecting metrics

Toktar:


libindy/src/api/metrics.rs
```diff
@@ -0,0 +1,43 @@
...
+use indy_api_types::errors::prelude::*;
+use indy_utils::ctypes;
+use indy_api_types::validation::Validatable;
+
+use serde_json;
+use libc::c_char;
+
+/// Generate wallet master key.
+/// Returned key is compatible with "RAW" key derivation method.
+/// It allows to avoid expensive key derivation for use cases when wallet keys can be stored in a secure enclave.
```
skhoroshavin, line 16: I think this needs to be updated

libindy/src/api/metrics.rs
```diff
@@ -0,0 +1,43 @@
...
+/// err: Error code
+///
+/// #Errors
+/// Common*
+/// Wallet*
+#[no_mangle]
+pub extern fn indy_collect_metrics(command_handle: CommandHandle,
+                                       cb: Option<extern fn(command_handle_: CommandHandle,
+                                                            err: ErrorCode,
+                                                            key: *const c_char)>) -> ErrorCode {
```
skhoroshavin, line 28: This should return list of key-value pairs, not a single string

libindy/src/api/metrics.rs
```diff
@@ -0,0 +1,43 @@
...
+    trace!("indy_collect_metrics: >>> command_handle: {:?}, cb: {:?}",
+           command_handle, cb);
+
+    check_useful_c_callback!(cb, ErrorCode::CommonInvalidParam3);
+
+    trace!("indy_generate_wallet_key: params config: {:?}", secret!(config.as_ref()));
```
skhoroshavin, line 34: I suspect these are also left-overs from code that was taken as a template...

libindy/src/commands/metrics.rs
```diff
@@ -0,0 +1,43 @@
...
+
+pub enum MetricsCommand {
+    CollectMetrics(Box<dyn Fn(IndyResult<()>) + Send>)
+}
+
+pub struct MetricsCommandExecutor {
+    close_callbacks: RefCell<HashMap<CommandHandle, Box<dyn Fn(IndyResult<()>)>>>,
+    refresh_callbacks: RefCell<HashMap<CommandHandle, Box<dyn Fn(IndyResult<()>)>>>,
+    // TODO: change  PoolHandle to MetricsHandler
+    open_callbacks: RefCell<HashMap<CommandHandle, Box<dyn Fn(IndyResult<PoolHandle>)>>>,
```
skhoroshavin, line 18: Do we really use/need these callbacks, or they are also leftovers from wallet command executor?

skhoroshavin (changes requested):

libindy/src/commands/metrics.rs
```diff
@@ -0,0 +1,36 @@
...
+    }
+    fn _collect(&self) -> IndyResult<String> {
+
+        trace!("_collect >>>");
+
+        let metrics_map : HashMap<String, f32> = HashMap::new();
+        let res = serde_json::to_string(&metrics_map).unwrap();
+
+        trace!("_collect <<< res: {:?}", res);
+        debug!("collecting metrics from the current thread");
```
skhoroshavin, line 32:
```suggestion
        debug!("collecting metrics from command thread");
```

skhoroshavin (approved):

libindy/src/commands/metrics.rs
```diff
@@ -0,0 +1,36 @@
...
+            }
+
+        };
+    }
+    fn _collect(&self) -> IndyResult<String> {
+
+        trace!("_collect >>>");
+
+        let metrics_map : HashMap<String, f32> = HashMap::new();
+        let res = serde_json::to_string(&metrics_map).unwrap();
```
jovfer, line 29:
```suggestion
        let res = serde_json::to_string(&metrics_map)?;
```

jovfer (changes requested):

jovfer (approved):

PR #2256, opened 2020-10-20
iOS wrapper: annocreds proverDeleteCredential

blazsolar:
Signed-off-by: Blaz Solar <me@blaz.solar>


wrappers/ios/libindy-pod/Indy-demoTests/Test Utils/AnoncredsUtils.m
```diff
@@ -17,17 +17,30 @@ @implementation AnoncredsUtils
...
+            instance = [AnoncredsUtils new];
+            instance.isCommonWalletCreated = false;
+            instance.walletHandle = 0;
+        }
+    }
 
     return instance;
 }
 
++ (void)clearInstance {
```
blazsolar, line 32: This is probably not the best approach but after credentials are removed wallet has to be recreated in order for other tests to work as tests use singleton instance. I'm open to better suggestions.

jovfer (approved):

PR #2260, opened 2020-10-25
updated getting-started.dockerfile to use 'notebook' instead of 'ipyt…

ColinClark:
…hon-notebook'

there doesn't seem to be any current version of 'ipython-notebook' as opposed to just 'notebook'


ColinClark:
i didn't follow the directions, am re-submitting

PR #2266, opened 2020-10-28
Nosodium

AxelNennker:
Use Rust methods to get random bytes and remove dependency to libsodium for that specific task.


vimmerru:
@AxelNennker Hi, what problem are you trying to solve?

FYI In this PR you just added dependency to new crypto library (rand) and as i can see it is used not only for tests. Also i don't see any entropy-based initialization for this random numbers generator.

It may work, but for such changes there should be strong motivation and analysis of rand crate from our crypto engineers.

AxelNennker:
> @AxelNennker Hi, what problem are you trying to solve?
> 
> FYI In this PR you just added dependency to new crypto library (rand) and as i can see it is used not only for tests. Also i don't see any entropy-based initialization for this random numbers generator.
> 
> It may work, but for such changes there should be strong motivation and analysis of rand crate from our crypto engineers.

HI @vimmerru ,
I am trying to get rid of libsodium and replace it by pure Rust.
As a starter I chose this two methods.

libindy is way behind in which version of sodiumoxide it uses. 
sodiumoxide = {version = "0.0.16"}
https://crates.io/crates/sodiumoxide is 0.2.6

Newer versions allow libsodium to be build on Android which is currently done by special build scripts.
My goal is to be able to build libindy for Android by using `cargo build --target=aarch64-linux-android`
Updating sodiumoxide to 0.2.6 has problems because the Debian package seems to need a static build of libsodium.
I gave up on that because general access to Jenkins was revoked. Maybe I try again now.

PR #2267, opened 2020-10-29
Update indentation ubuntu-build.md

znss1989:
The indentation of lines from 73 to 83 are wrong, which makes it confusing to readers. In fact these lines do not belong to the **optional** build of `indy-cli`.

### DIFF
--- a/docs/build-guides/ubuntu-build.md
+++ b/docs/build-guides/ubuntu-build.md
@@ -70,14 +70,15 @@ For manually building this can be achieved by passing `--features sodium_static`
    cd cli/
    RUSTFLAGS=" -L ../libindy/target/debug" cargo build
    ```
-   If you have followed the instructions to build libindy above, the default build type will be `debug`
+   
+If you have followed the instructions to build libindy above, the default build type will be `debug`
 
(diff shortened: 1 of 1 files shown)


brentzundel:
I read those lines as applying directly to the build instructions for indy-cli and don't agree that the indentation here is incorrect.

znss1989:
Thanks for replying, but I guess adding the libindy executable to path should not be an optional action for a potential user.

vimmerru:
imag@znss1989

@znss1989 I believe list numbering and indentation is still broken in your fix. Check the screenshot.

znss1989:
@vimmeru, thanks for checking. But I did not change that part. My point is that lines 73 to 81 should not be indented, giving readers the impression that it is under the Optional subsection, which is not true.

vimmerru:
> @vimmeru, thanks for checking. But I did not change that part. My point is that lines 73 to 81 should not be indented, giving readers the impression that it is under the Optional subsection, which is not true.

the root of the issue you are trying to solve is broken structuring in the whole document. So your fix doesn't help significantly.

PR #2269, opened 2020-10-29
[VE-2045][3]: add command metrics

Toktar:
TODO:

- [x] - add tests
- [x] - fix command metrics displaying
- [x] - find better solution for using &Command instead indexes writing command metrics
- [x] - add constants for command metrics

### DIFF
--- a/libindy/Cargo.toml
+++ b/libindy/Cargo.toml
@@ -56,6 +56,9 @@ indy-api-types = { path = "./indy-api-types"}
 indy-utils = { path = "./indy-utils"}
 indy-wallet = { path = "./indy-wallet"}
 quote = "=1.0.7"
+variant_count = "*"
+num-traits = "0.2"
+num-derive = "0.2"
 
 [dependencies.uuid]
 version = "0.7.4"
--- a/libindy/Cargo.toml
+++ b/libindy/Cargo.toml
@@ -59,6 +59,7 @@ quote = "=1.0.7"
 variant_count = "*"
 num-traits = "0.2"
 num-derive = "0.2"
+convert_case = "0.3.2"
 
 [dependencies.uuid]
 version = "0.7.4"
--- a/libindy/src/commands/metrics.rs
+++ b/libindy/src/commands/metrics.rs
@@ -1,24 +1,8 @@
-use variant_count::VariantCount;
 use indy_api_types::errors::prelude::*;
 use indy_wallet::WalletService;
 use std::rc::Rc;
-use crate::commands::Command;
-use crate::commands::ledger::LedgerCommand;
-use crate::commands::anoncreds::AnoncredsCommand;
-use crate::commands::anoncreds::issuer::IssuerCommand;
-use crate::commands::anoncreds::prover::ProverCommand;
-use crate::commands::anoncreds::verifier::VerifierCommand;
-use crate::commands::blob_storage::BlobStorageCommand;
-use crate::commands::crypto::CryptoCommand;
-use crate::commands::pool::PoolCommand;
-use crate::commands::did::DidCommand;
-use crate::commands::wallet::WalletCommand;
-use crate::commands::pairwise::PairwiseCommand;
-use crate::commands::non_secrets::NonSecretsCommand;
-use crate::commands::payments::PaymentsCommand;
-use crate::commands::cache::CacheCommand;
+use crate::services::metrics::MetricsService;
 use serde_json::{Map, Value};
-use std::fmt;
 
 const THREADPOOL_ACTIVE_COUNT: &str = "threadpool_active_count";
 const THREADPOOL_QUEUED_COUNT: &str = "threadpool_queued_count";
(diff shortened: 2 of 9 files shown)


libindy/src/commands/metrics.rs
```diff
@@ -12,18 +29,585 @@ const OPENED_WALLET_IDS_COUNT: &str = "opened_wallet_ids_count";
...
+    CacheCommandGetSchemaContinue,
+    CacheCommandGetCredDef,
+    CacheCommandGetCredDefContinue,
+    CacheCommandPurgeSchemaCache,
+    CacheCommandPurgeCredDefCache,
+    // MetricsCommand
+    MetricsCommandCollectMetrics,
+    // Exit
+    Exit,
+}
```
skhoroshavin, line 549: I think it would be better to move CommandIndex-related boilerplate to separate file to improve readability of core metrics logic. However this is just a matter of taste...

libindy/src/commands/metrics.rs
```diff
@@ -12,18 +29,585 @@ const OPENED_WALLET_IDS_COUNT: &str = "opened_wallet_ids_count";
...
 }
 
+const COMMANDS_COUNT: usize = MetricsStorage::commands_count();
+
+pub struct MetricsStorage {
+    queue_count: [u64; COMMANDS_COUNT],
+    queue_duration: [u64; COMMANDS_COUNT],
+
+    execute_count: [u64; COMMANDS_COUNT],
+    execute_duration: [u64; COMMANDS_COUNT],
```
skhoroshavin, line 563: I think these names would make more sense:
* queue**d**_commands_count
* queued_commands_duration_ms
* execute**d**_commands_count
* executed_commands_duration_ms

libindy/src/commands/metrics.rs
```diff
@@ -12,18 +29,585 @@ const OPENED_WALLET_IDS_COUNT: &str = "opened_wallet_ids_count";
...
+        MetricsStorage {
+            queue_count: [u64::MIN; COMMANDS_COUNT],
+            queue_duration: [u64::MIN; COMMANDS_COUNT],
+
+            execute_count: [u64::MIN; COMMANDS_COUNT],
+            execute_duration: [u64::MIN; COMMANDS_COUNT],
+        }
+    }
+
+    pub fn cmd_left_queue(&mut self, command_index: usize, duration: u64) {
```
skhoroshavin, line 577: Why not just pass `Command` here, or at least `CommandIndex`? It would be much less error prone.

Btw, if you try to avoid redundant conversions from `Command` to `CommandIndex` to `usize` to get the best performance possible then:
* I'm 99% sure conversion from `CommandIndex` to `usize` is zero-cost
* I'm 90% sure conversion from `Command` to `CommandIndex` is close to zero-cost

libindy/src/commands/metrics.rs
```diff
@@ -12,18 +29,585 @@ const OPENED_WALLET_IDS_COUNT: &str = "opened_wallet_ids_count";
...
+    }
+    const fn commands_count() -> usize {
+        CommandIndexes::VARIANT_COUNT
+    }
+}
+
 pub struct MetricsCommandExecutor {
-    wallet_service: Rc<WalletService>
+    wallet_service: Rc<WalletService>,
+    pub metric_storage: MetricsStorage,
```
skhoroshavin, line 603: I'm not sure about practice of using public structure members...

libindy/src/commands/metrics.rs
```diff
@@ -12,18 +29,585 @@ const OPENED_WALLET_IDS_COUNT: &str = "opened_wallet_ids_count";
...
+    queue_duration: [u64; COMMANDS_COUNT],
+
+    execute_count: [u64; COMMANDS_COUNT],
+    execute_duration: [u64; COMMANDS_COUNT],
+}
+
+impl MetricsStorage {
+    pub fn new() -> MetricsStorage {
+        MetricsStorage {
+            queue_count: [u64::MIN; COMMANDS_COUNT],
```
skhoroshavin, line 569: My Rust knowledge is a little bit rusty, so I might be wrong here, but I thought it should be something like `[u64; COMMANDS_COUNT]`, why `MIN` here? 

FYI @jovfer

libindy/src/commands/metrics.rs
```diff
@@ -35,26 +619,53 @@ impl MetricsCommandExecutor {
...
+        self._append_command_metrics(&mut metrics_map);
         let res = serde_json::to_string(&metrics_map)
-            .to_indy(IndyErrorKind::InvalidState, "Can't serialize a metrics map")?;
+            .to_indy(IndyErrorKind::InvalidState,"Can't serialize a metrics map")?;
 
         trace!("_collect <<< res: {:?}", res);
         debug!("collecting metrics from command thread");
         Ok(res)
     }
+    fn _append_threapool_metrics(&self, metrics_map: &mut Map<String, Value>) {
```
skhoroshavin, line 637: I don't know code conventions of libindy, does it require padding private methods with `_`? FYI @jovfer

libindy/src/commands/metrics.rs
```diff
@@ -35,26 +619,53 @@ impl MetricsCommandExecutor {
...
+                           Value::from(self.wallet_service.get_wallet_ids_count()));
+        metrics_map.insert(String::from(PENDING_FOR_IMPORT_WALLETS_COUNT),
+                           Value::from(self.wallet_service.get_pending_for_import_count()));
+        metrics_map.insert(String::from(PENDING_FOR_OPEN_WALLETS_COUNT),
+                           Value::from(self.wallet_service.get_pending_for_open_count()));
+    }
+    fn _append_command_metrics(&self, metrics_map: &mut Map<String, Value>) {
+        //TODO: fix incorrect keys displaying
+        for index in (0..MetricsStorage::commands_count()).rev() {
+            metrics_map.insert(format!("{:?}_execute_count", MetricsStorage::cmd_name(index)).as_str().to_string(),
```
skhoroshavin, line 661: executed_count?

libindy/src/commands/metrics.rs
```diff
@@ -35,26 +619,53 @@ impl MetricsCommandExecutor {
...
+                               Value::from(self.metric_storage.execute_count[index] as usize));
+            metrics_map.insert(format!("{:?}_execute_duration", MetricsStorage::cmd_name(index)).as_str().to_string(),
```
skhoroshavin, line 663: executed_duration_ms?

libindy/src/commands/metrics.rs
```diff
@@ -35,26 +619,53 @@ impl MetricsCommandExecutor {
...
+                               Value::from(self.metric_storage.execute_duration[index] as usize));
+            metrics_map.insert(format!("{:?}_queue_count", MetricsStorage::cmd_name(index)).as_str().to_string(),
```
skhoroshavin, line 665: queued_count?

libindy/src/commands/metrics.rs
```diff
@@ -35,26 +619,53 @@ impl MetricsCommandExecutor {
...
+                               Value::from(self.metric_storage.queue_count[index] as usize));
+            metrics_map.insert(format!("{:?}_queue_duration", MetricsStorage::cmd_name(index)).as_str().to_string(),
```
skhoroshavin, line 667: queued_duration_ms?

libindy/src/commands/mod.rs
```diff
@@ -80,9 +96,14 @@ pub fn indy_set_runtime_config(config: IndyConfig) {
     }
 }
 
+fn get_cur_time() -> u64 {
+    let since_epoch = SystemTime::now().duration_since(UNIX_EPOCH).expect("Time has gone backwards");
```
skhoroshavin, line 100: If I understand it right this is a potential panic, which is not okay for production code

libindy/src/commands/mod.rs
```diff
@@ -80,9 +96,14 @@ pub fn indy_set_runtime_config(config: IndyConfig) {
...
+    since_epoch.as_secs()
```
skhoroshavin, line 101: Seconds resolution is not enough for these kind of measurements, we need at least milliseconds.

libindy/Cargo.toml
```diff
@@ -56,6 +56,9 @@ indy-api-types = { path = "./indy-api-types"}
 indy-utils = { path = "./indy-utils"}
 indy-wallet = { path = "./indy-wallet"}
 quote = "=1.0.7"
+variant_count = "*"
+num-traits = "0.2"
+num-derive = "0.2"
```
skhoroshavin, line 61: Do we need ALL of these?

libindy/src/commands/metrics.rs
Toktar: No, I am struggling with mutable links issue. But I hope to deal with this problem today. Therefore, there is still TODO.

libindy/src/commands/metrics.rs
Toktar: I think you're right. Would it be better to make the storage getter or wrappers over the storage functions the safest method?

libindy/src/commands/metrics.rs
Toktar: I also doubt my Rust knowledge. But this is creating an array of zeros.

libindy/src/commands/mod.rs
Toktar: I used this way from another production code. I can change error case to returning zero. But in any case it's a critical situation when we can't get a system time, in my opinion.

libindy/Cargo.toml
Toktar: ```variant_count``` - to get CommandIndexes size
```num-traits``` - to get an element of enum by its index
```num-derive`` - to add ```FromPromitive``` derive to CommandIndexes for the previous action.

File: libindy/src/commands/mod.rs [hunk 14]
```diff
@@ -122,75 +143,88 @@ impl CommandExecutor {
...
+                            cmd
+                        }
+                        Err(err) => {
+                            error!("Failed to get command!");
+                            panic!("Failed to get command! {:?}", err)
+                        }
+                    };
+                    let command_index = MetricsStorage::cmd_index(&instrumented_cmd.command);
+                    let start_execution_ts = get_cur_time();
+                    metrics_command_executor.metric_storage.cmd_left_queue(command_index,
```
skhoroshavin, line 160: After discussion with @jovfer it looks like moving that into metrics service could be much more sensible

libindy/src/services/metrics/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+                               Value::from(self.executed_commands_duration_ms.borrow()[index] as usize));
+            metrics_map.insert(format!("{}_{}", cmd_name.as_str(), QUEUED_COMMANDS_COUNT).as_str().to_string(),
+                               Value::from(self.queued_commands_count.borrow()[index] as usize));
+            metrics_map.insert(format!("{}_{}", cmd_name.as_str(), QUEUED_COMMANDS_DURATION_MS).as_str().to_string(),
+                               Value::from(self.queued_commands_duration_ms.borrow()[index] as usize));
+        }
+    }
+}
+
+mod test {
```
skhoroshavin, line 65: I guess you meant
```
#[cfg(test)]
mod test {
...
```

libindy/src/services/metrics/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+    use super::*;
+
+    #[test]
+    fn test_cmd_left_queue() {
+        let metrics_service = MetricsService::new();
+        let index = CommandIndex::IssuerCommandCreateSchema;
+        let durations = [5u128, 2u128, 3u128];
+        let command_count: u128 = durations.len() as u128;
+
+        for duration in durations.iter() {
```
skhoroshavin, line 75: Please, try to avoid loops in tests

libindy/src/services/metrics/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+        }
+
+        assert_eq!(metrics_service.queued_commands_count.borrow()[index as usize], 0);
+        assert_eq!(metrics_service.queued_commands_duration_ms.borrow()[index as usize], 0);
+        assert_eq!(metrics_service.executed_commands_count.borrow()[index as usize], command_count);
+        assert_eq!(metrics_service.executed_commands_duration_ms.borrow()[index as usize], durations.iter().sum::<u128>());
+    }
+    #[test]
+    fn test_append_command_metrics() {
+        let payment_service = MetricsService::new();
```
skhoroshavin, line 102: Name `payment_service` looks a bit misleading in my opinion

libindy/tests/metrics.rs
```diff
@@ -25,13 +20,31 @@ mod collect {
     fn collect_metrics_works() {
```
skhoroshavin, line 20: `collect_metrics_contains_thread_pool_and_wallet_service_statistics`?

And if that name looks too long - probably it is a sign that this need to be split into two separate tests ;)

libindy/src/services/metrics/mod.rs
```diff
@@ -0,0 +1,113 @@
...
+                               Value::from(self.queued_commands_duration_ms.borrow()[index] as usize));
+        }
+    }
+}
+
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_cmd_left_queue() {
```
skhoroshavin, line 69: What about naming this `test_cmd_left_queue_increments_relevant_queued_counters`?

libindy/tests/metrics.rs
```diff
@@ -25,13 +20,31 @@ mod collect {
...
+        assert!(metrics_map.contains_key("threadpool_max_count"));
+        assert!(metrics_map.contains_key("threadpool_panic_count"));
+        assert!(metrics_map.contains_key("opened_wallets_count"));
+        assert!(metrics_map.contains_key("opened_wallet_ids_count"));
+        assert!(metrics_map.contains_key("pending_for_import_wallets_count"));
+        assert!(metrics_map.contains_key("pending_for_open_wallets_count"));
+    }
+
+    #[test]
+    fn collect_metrics_after_works() {
```
skhoroshavin, line 34: `collect_metrics_includes_statistics_for_wallet_command`?

Toktar: Done

libindy/src/services/metrics/mod.rs
Toktar: Done

libindy/tests/metrics.rs
Toktar: Changed to ```collect_metrics_contains_thread_pool_and_wallet_service_statistics```

libindy/src/services/metrics/mod.rs
Toktar: fixed

libindy/src/services/metrics/mod.rs
Toktar: fixed

libindy/src/services/metrics/mod.rs
Toktar: Added

skhoroshavin (approved):

PR #2271, opened 2020-10-30
VE-2045[4]: add metrics collecting to the Java wrapper

Toktar:


wrappers/java/src/main/java/org/hyperledger/indy/sdk/metrics/Metrics.java
```diff
@@ -0,0 +1,56 @@
...
+    /*
+     * STATIC CALLBACKS
+     */
+
+    /**
+     * Callback used when method with string result completes
+     */
+    private static Callback stringCompleteCb = new Callback() {
+
+        @SuppressWarnings({"unused", "unchecked"})
```
skhoroshavin, line 24: Why do we need these?

wrappers/java/src/test/java/org/hyperledger/indy/sdk/metrics/CollectMetricsTest.java
```diff
@@ -0,0 +1,16 @@
...
+import org.junit.Test;
+import static org.junit.Assert.assertNotNull;
+
+
+public class CollectMetricsTest extends IndyIntegrationTest {
+
+	@Test
+	public void testCollectMetricsMethod() throws Exception {
+		String metrics_map = Metrics.collectMetrics().get();
+		assertNotNull(metrics_map);
```
skhoroshavin, line 13: I would also suggest to test that returned string is a valid JSON

wrappers/java/src/main/java/org/hyperledger/indy/sdk/metrics/Metrics.java
Toktar: This annotation is used for all callbacks in libindy. I suggest leaving it.

wrappers/java/src/test/java/org/hyperledger/indy/sdk/metrics/CollectMetricsTest.java
Toktar: Added a few checks for a result JSON.

skhoroshavin (approved):

PR #2282, opened 2020-11-10
[issues-2277] Set universal openssl for Apple devices

adenishchenko:
Signed-off-by: Anton Denishchenko <anton.denishchenko@dsr-corporation.com>


jovfer:
Universal SSL doesn't allow to fix iOS build problem. The temporary solution is in #2296 and the final one would be migration to rust 1.49.0

PR #2286, opened 2020-11-16
Release candidate for indy sdk version 1.16.0

ianco:


ianco:
DCO issue, kill it and start over ...

PR #2294, opened 2020-11-20
[IS-2109] Async libindy internals

adenishchenko:


libindy/src/commands/mod.rs
```diff
@@ -202,31 +212,54 @@ impl CommandExecutor {
...
                             metrics_command_executor.execute(cmd);
                         }
                         Command::Exit => {
                             debug!("Exit command received");
-                            break
                         }
                     }
+
                     metrics_service.cmd_executed(cmd_index,
                                                  get_cur_time() - start_execution_ts);
```
skhoroshavin, line 231: Now executed time includes waiting time as well. This is not a blocker for this PR, but as a next step I would suggest to spend at least some time to look into possibility of tracking executing and waiting time separately, including whether we need this separation.

skhoroshavin (approved):

libindy/src/commands/mod.rs
jovfer: FYI @Toktar

jovfer:
Superseded by #2319

PR #2296, opened 2020-11-23
Downgrade Rust to 1.45.0 for iOS pipelines.

jovfer:


Artemkaaas (approved):

vimmerru:
I suggest to try disable cdylib target. Most probably it is the cause of the issue. In Rust 1.46 they added support of cdylib builds that was skipped before. cdylib requires full linking chain. Related change log item is here https://github.com/rust-lang/rust/blob/master/RELEASES.md#compiler-2

jovfer:
@vimmerru  cdylib is required for our integration tests structure as far as I remember. Please raise separate PR if you find a way to get rid of this target and keep tests compilation/run behavior.

PR #2303, opened 2020-11-25
Fixes for iOS build

ianco:
Fixes iOS build and includes other fixes for 1.16.0 sdk release


ianco:
Note had to amend an older commit to add DCO (https://github.com/hyperledger/indy-sdk/pull/2303/commits/b45a36ae75d31a16d3154ca8b3121725a0074866) and inadvertently added extra "Signed Off" to other commits, not sure if this is kosher or not?

jovfer:
I thought the problematic commit (missed DCO) is already in RC branch (merged by #2289) and now we don't need to do more steps. I will check the commit history and share my thoughts tomorrow

ianco:
@jovfer it was fixed in RC branch but then re-introduced in the PR from master.  Thanks in advance for looking into this.

WadeBarnes:
The extra sig-offs in the commits are messy and inactivate.  They should be fixed.  Some examples:

image

image

ianco:
One option is to cherry-pick commits from master rather than PR'ing the whole branch

ianco:
... weirdly when I originally fixed this in the RC branch it only signed the one offending commit.  Today when I ran through the exact same steps it re-signed everything.  (I tried a few times with the same results.)

ianco:
Superceded ...

PR #2307, opened 2020-11-26
NodeJS wrapper updates

TimoGlastra:
This updates the dependencies for the NodeJS wrapper, thereby it also drops support for Node v8 which reached its LTS end of life in 2019. It works in both v10 and v12. v14 still needs some work.

Because Node v8 is not supported anymore this is a breaking change, however the package version matches the libindy version so I'm not sure how that should be handled


jovfer:
Hi @TimoGlastra  ,
we had to re-write master history to fix DCO there. So now this PR should be updated as well.
Can I force-push to your branch or would you like to do it by yourself or to create a new one?

TimoGlastra:
Hi @jovfer,

The CI is failing anyway so I need to make some changes first. I will create an updated PR once ready.

jovfer:
OK, thanks. Please note, the Jenkins has some security policies and may ignore non-trusted changes in `Jenkinsfile.ci` or `.cd`
Let us know if you will need to run modified CI pipeline and one of the maintainers will assists

PR #2308, opened 2020-11-26
RC 1.16.0

jovfer:


jovfer:
@WadeBarnes @ianco I finally looked at the problem with DCO of commits in `master` and `rc` branches and other related stuff like merge problems. 

Here is my **understanding of some facts**, please check if you find it correct:

1. There was a commit with broken DCO https://github.com/hyperledger/indy-sdk/commit/f84e91f6ef618398f7a593499ea30e83b246ddb6
2. It was merged to master in #2285 by setting this check to pass manually
3. The copy of this commit and probably more alternative history now in `rc` branch. https://github.com/hyperledger/indy-sdk/commit/0bf0748d08ad295044c1d1c916b453ce86dc6bb0
and it was merged in #2289 to RC

As a result of 1-3 we have a divergence of history in `rc` and `master`.

My **thoughts/suggestions**

1. I think we should keep `rc` and `master` branches in the state which will allow us to do merges in both directions
2. As far as I know we should have all commits in every history branch to do 1st point without the overhead and problems like Ian saw while trying to apply iOS fixes in #2302 pr #2303
3. One way is to force-push one of the alternative histories and override `rc` or `master` or both
4. Another option is to just merge the original `master` history with broken DCO to `rc`. So it's the current PR. (DCO is set to pass manually again here)

**Summary**
I don't know other principal ways to sync branches in addition to 3rd or 4th options in the list above. Between 2 of them in a personal repo I would strongly prefer to follow the override option. But in our case (public repo) overriding history may confuse a lot of contributors so the 4th one seems more reasonable.

ianco:
Hi @jovfer, thanks for looking into this!

Just one correction - when I PR'ed the "bad" commit into the `rc` branch I fixed the DCO problem.  However I didn't/don't have the permissions to fix the commit in the `master` branch (and I'm a bit nervous to start messing around with the commit history in `master`).  Right now `rc` has the same updates as in `master` (as in the same set of file changes) - I've cherry-picked commits to bring the `rc` branch up-to-date, but making sure all the commits have properly signed off DCO.

I agree that option 4 is the most straightforward approach, as this keeps all the commits the same in both branches, and simplifies merges that we will need to do in the future.  However it leaves the "bad" commit in place without a proper DCO signoff, and will require a manual override every time we need to merge branches.

If we are going to go with option 3 (would force everyone to update their fork but long-term would solve the DCO problem) then the sooner the better, the more development that happens the deeper the DCO problem becomes.

@ryjones @WadeBarnes

jovfer:
@ianco I'd like to double-check if we are on the same page

> and will require a manual override every time we need to merge branches.

do you mean merges of these branches to some others and overriding the DCO check?

ianco:
@jovfer  yes that's what I meant, overriding the DCO check.

WadeBarnes:
@jovfer, @ianco,  My take on this is we need to amend the offending commit in `master`.  That will require amending the commit and rebasing the remainder of the commits on top of that, which will require a force push due to the resulting change in history.  The sooner we do this the better.  This would need to be done in co-ordination with an admin on the repo as some branch protection settings would need to be temporarily disabled.

The unfortunate consequence is this will mess with developer history and require developers to resync with the new state of `master`.  We can decide what to do with the `rc` branch once the state of `mater` has been repaired.

If we don't do this the offending commit will constantly and continually be an issue moving forward.

cc @ryjones

jovfer:
@ryjones could you approve re-writing history of the master branch and give me enough permissions to do so?

ryjones:
> @ryjones could you approve re-writing history of the master branch and give me enough permissions to do so?

With great power, etc.

jovfer:
Closed as the master was fixed in #2313

PR #2292, opened 2020-11-19
Centos zmq

AxelNennker:
related to https://github.com/hyperledger/indy-sdk/issues/2215
Please also see: https://github.com/zeromq/libzmq/security/advisories

I haven't checked the Debian/Ubuntu versions indy uses
https://software.opensuse.org/download.html?project=network%3Amessaging%3Azeromq%3Arelease-stable&package=libzmq3-dev

PR #2311, opened 2020-11-28
Rusqlite 0.24

AxelNennker:
After I updated to the latest stable rust version the tests failed.

The stacktrace showed the error is related to rusqlite. So, I update rusqlite from 0.20 to 0.24.
This fixes the uninitialized error panic.

After updating to 0.24 the compiler complained about two match statements not covering all cases. 
My choices to fix these errors are open for discussion - of course. :-)


jovfer:
Hi Axel, same question as https://github.com/hyperledger/indy-sdk/pull/2292#issuecomment-735728962

AxelNennker:
Now I am getting this.
```
+ cargo clippy --manifest-path libindy/Cargo.toml -- -W clippy::style -W clippy::correctness -W clippy::complexity -W clippy::perf

    Updating crates.io index

error: checksum for `libsqlite3-sys v0.20.1` changed between lock files


this could be indicative of a few possible errors:


    * the lock file is corrupt

    * a replacement source in use (e.g., a mirror) returned a different checksum

    * the source itself may be corrupt in one way or another


unable to verify that `libsqlite3-sys v0.20.1` is the same as when the lockfile was generated


script returned exit code 101
```
After that I did a `cargo update` and pushed the new Cargo.lock but nothing changed. Strange.

Sorry, for the DCO issue. Not sure how that happened.
Maybe I close this PR and recreate it. Simple enough.
If a maintainer wants to do this simple update, please don't hesitate doing it yourself.

PR #2318, opened 2020-12-02
Tags for metrics

adenishchenko:


libindy/src/services/metrics/mod.rs
```diff
@@ -125,10 +150,8 @@ mod test {
 
         metrics_service.append_command_metrics(&mut metrics_map);
 
-        assert_eq!(metrics_map.len(), COMMANDS_COUNT * 4);
-        assert!(metrics_map.contains_key("issuer_command_create_schema_queued_commands_count"));
-        assert!(metrics_map.contains_key("issuer_command_create_schema_queued_commands_duration_ms"));
-        assert!(metrics_map.contains_key("issuer_command_create_schema_executed_commands_count"));
-        assert!(metrics_map.contains_key("issuer_command_create_schema_executed_commands_duration_ms"));
+        assert!(metrics_map.contains_key("commands_count"));
+        assert!(metrics_map.contains_key("commands_duration"));
```
skhoroshavin, line 154:
```suggestion
        assert!(metrics_map.contains_key("commands_duration_ms"));
```
It is a good practice to indicate unit of measure in metrics name, let's stick to it.

libindy/src/services/metrics/mod.rs
```diff
@@ -125,10 +150,8 @@ mod test {
...
+        assert_eq!(metrics_map.get("commands_count").unwrap().as_array().unwrap().len() + metrics_map.get("commands_duration").unwrap().as_array().unwrap().len(), COMMANDS_COUNT * 4);
```
skhoroshavin, line 155: It would be much better to test number of `command_count ` and `commands_duration_ms` metrics separately (checking that each are `COMMANDS_COUNT * 2`)

Also, this test is definitely not complete now - we should test internal schema as well (i.e. that metrics include array of objects, and each object has value and tags, and for timing metrics tags are objects with fields like command, subcommand and stage, and stage is either queued or executed, etc etc)

libindy/src/commands/metrics.rs
```diff
@@ -55,26 +56,47 @@ impl MetricsCommandExecutor {
...
         let tp_instance = crate::commands::THREADPOOL.lock().unwrap();
-        metrics_map.insert(String::from(THREADPOOL_ACTIVE_COUNT),
-                           Value::from(tp_instance.active_count()));
-        metrics_map.insert(String::from(THREADPOOL_QUEUED_COUNT),
-                           Value::from(tp_instance.queued_count()));
-        metrics_map.insert(String::from(THREADPOOL_MAX_COUNT),
-                           Value::from(tp_instance.max_count()));
-        metrics_map.insert(String::from(THREADPOOL_PANIC_COUNT),
-                           Value::from(tp_instance.panic_count()));
+        let metrics_map = checkMetricsStructure(metrics_map);
```
skhoroshavin, line 63: 1. I would prefer to stick to snake_case for function names to be consistent with the rest of codebase
2. Function that mutates it's arguments shouldn't be named like `check_something`. I would propose `ensure_metrics_contain_<yadayada>`.
3. I checked contents of this function - and it is commented out. How does this code compile at all?
4. This function ensures that there are command-metrics inside of map - is it really needed, and if yes - why is this being done in function that is supposed to add threadpool-related metrics?

libindy/src/commands/metrics.rs
```diff
@@ -55,26 +56,47 @@ impl MetricsCommandExecutor {
...
+        let mut threadpool_threads_count = Vec::new();
+
+        threadpool_threads_count.push(Value::from("value", tp_instance.active_count()));
+        threadpool_threads_count.push(Value::from("tags", MetricsTags{stage: "active".to_string()}));
```
skhoroshavin, line 67: I wouldn't call tag as `stage` in this case. Probably `state` is a better fit here?

Also, do we really need a custom structure here instead? Rust expertise from @jovfer would be welcome here as well.

libindy/src/commands/metrics.rs
```diff
@@ -55,26 +56,47 @@ impl MetricsCommandExecutor {
...
+
+        threadpool_threads_count.push(Value::from("value", tp_instance.queued_count()));
+        threadpool_threads_count.push(Value::from("tags", MetricsTags{stage: "queued".to_string()}));
+
+        threadpool_threads_count.push(Value::from("value", tp_instance.max_count()));
+        threadpool_threads_count.push(Value::from("tags", MetricsTags{stage: "max".to_string()}));
```
skhoroshavin, line 73: `max` state for thread looks like a.. stretch :) Why not a separate metric in this case?

libindy/src/services/metrics/mod.rs
```diff
@@ -40,25 +42,48 @@ impl MetricsService {
...
-                               Value::from(self.executed_commands_duration_ms.borrow()[index] as usize));
-            metrics_map.insert(format!("{}_{}", cmd_name.as_str(), QUEUED_COMMANDS_COUNT).as_str().to_string(),
-                               Value::from(self.queued_commands_count.borrow()[index] as usize));
-            metrics_map.insert(format!("{}_{}", cmd_name.as_str(), QUEUED_COMMANDS_DURATION_MS).as_str().to_string(),
-                               Value::from(self.queued_commands_duration_ms.borrow()[index] as usize));
+            let (command, subcommand) = MetricsService::cmd_name(index);
+            let mut tags = Value::from(MetricsTags{command, subcommand, stage: "executed".to_string() });
+
+            commands_count.push(
+            Value::from(MetricsValue::new(self.executed_commands_count.borrow()[index] as usize, &tags))
```
skhoroshavin, line 70: Please fix formatting here

wrappers/nodejs/test/anoncreds.js
```diff
@@ -184,6 +184,15 @@ test('anoncreds', async function (t) {
...
   t.is(err.indyName, 'AnoncredsProofRejected')
 
+  // Prover deletes credential
+  // await indy.proverDeleteCredential(wh, outCredId)
+  // Make sure it is really gone from the wallet
+  // err = await t.throwsAsync( indy.proverGetCredential(wh, outCredId))
+  // t.is(err.indyName, 'WalletItemNotFound')
+  // Make sure we can't delete it again
+  // err = await indy.proverDeleteCredential(wh, outCredId)
+  // t.is(err.indyName, 'WalletItemNotFound')
```
skhoroshavin, line 194: This doesn't seem to be related to this PR, probably committed by mistake?

libindy/src/services/metrics/mod.rs
```diff
@@ -40,25 +42,48 @@ impl MetricsService {
...
+            commands_duration.push(
+                Value::from(MetricsValue::new(self.executed_commands_duration_ms.borrow()[index] as usize, &tags))
+            );
+
+            commands_duration.push(
+                Value::from(MetricsValue::new(self.queued_commands_duration_ms.borrow()[index] as usize, &tags))
+            );
         }
+        metrics_map.insert("commands_count".to_owned(), serde_json::to_value(commands_count).unwrap());
+        metrics_map.insert("commands_duration".to_owned(), serde_json::to_value(commands_duration).unwrap());
```
skhoroshavin, line 86: Please avoid `unwrap` in production code.

libindy/src/services/metrics/mod.rs
```diff
@@ -40,25 +42,48 @@ impl MetricsService {
...
         self.executed_commands_duration_ms.borrow_mut()[command_index as usize] += duration;
     }
-    pub fn cmd_name(index: usize) -> String {
-        CommandIndex::from(index).to_string().to_case(Case::Snake)
+
+    pub fn cmd_name(index: usize) -> (String, String) {
+        let cmd = CommandIndex::from(index).to_string().find("Command");
+        match cmd {
+            None => (CommandIndex::from(index).to_string().to_owned(), String::from("")),
+            Some(cmd) => (CommandIndex::from(index).to_string()[0..Some(cmd).unwrap()].to_owned(), CommandIndex::from(index).to_string()[Some(cmd).unwrap()+7..].to_owned()) ,
```
skhoroshavin, line 50: I don't think string search and manipulations (with magic `+7` offsets and unwraps all over the place) are good idea here. Why not just modify `CommandIndex` to have functions `command_name()` and `command_subname()`?

skhoroshavin (changes requested):
This is a good start, but it needs more work before merge.

libindy/src/commands/metrics.rs
```diff
@@ -42,39 +46,112 @@ impl MetricsCommandExecutor {
...
-                           Value::from(tp_instance.panic_count()));
-    }
+        let mut threadpool_threads_count = Vec::new();
+        let mut tag = HashMap::<String, String>::new();
+
+        tag.insert(String::from("label"), String::from(THREADPOOL_ACTIVE_COUNT));
+        threadpool_threads_count.push(
+            serde_json::to_value(MetricsValue::new(tp_instance.active_count(), tag.clone()))
+                .to_indy(IndyErrorKind::IOError, "Unable to convert json")?,
+        );
```
skhoroshavin, line 76: Can we simplify this somehow? Like extract common parts of it to common function?

libindy/src/services/metrics/mod.rs
```diff
@@ -31,34 +30,92 @@ impl MetricsService {
         }
     }
 
-    pub fn cmd_left_queue(&self, command_index: CommandIndex, duration: u128) {
+    pub fn cmd_left_queue(&self, command_index: CommandMetric, duration: u128) {
```
skhoroshavin, line 33: I guess we should rename variable here as well?

libindy/src/services/metrics/mod.rs
```diff
@@ -31,34 +30,92 @@ impl MetricsService {
...
-                               Value::from(self.executed_commands_duration_ms.borrow()[index] as usize));
-            metrics_map.insert(format!("{}_{}", cmd_name.as_str(), QUEUED_COMMANDS_COUNT).as_str().to_string(),
-                               Value::from(self.queued_commands_count.borrow()[index] as usize));
-            metrics_map.insert(format!("{}_{}", cmd_name.as_str(), QUEUED_COMMANDS_DURATION_MS).as_str().to_string(),
-                               Value::from(self.queued_commands_duration_ms.borrow()[index] as usize));
+            let command = MetricsService::cmd_name(index);
+            let tags_executed = MetricsService::get_command_tags(
+                command.clone(),
+                String::from("executed"),
+            );
```
skhoroshavin, line 70: Do we really need all that string allocation and copying here?

libindy/src/services/metrics/mod.rs
```diff
@@ -31,34 +30,92 @@ impl MetricsService {
...
-            let cmd_name = MetricsService::cmd_name(index);
-            metrics_map.insert(format!("{}_{}", cmd_name.as_str(), EXECUTED_COMMANDS_COUNT).as_str().to_string(),
-                               Value::from(self.executed_commands_count.borrow()[index] as usize));
-            metrics_map.insert(format!("{}_{}", cmd_name.as_str(), EXECUTED_COMMANDS_DURATION_MS).as_str().to_string(),
-                               Value::from(self.executed_commands_duration_ms.borrow()[index] as usize));
-            metrics_map.insert(format!("{}_{}", cmd_name.as_str(), QUEUED_COMMANDS_COUNT).as_str().to_string(),
-                               Value::from(self.queued_commands_count.borrow()[index] as usize));
-            metrics_map.insert(format!("{}_{}", cmd_name.as_str(), QUEUED_COMMANDS_DURATION_MS).as_str().to_string(),
-                               Value::from(self.queued_commands_duration_ms.borrow()[index] as usize));
+            let command = MetricsService::cmd_name(index);
```
skhoroshavin, line 66: I think `command_name` would be more obvious name of that variable.

libindy/src/services/metrics/mod.rs
```diff
@@ -31,34 +30,92 @@ impl MetricsService {
...
+                String::from("queued"),
+            );
+
+            commands_count.push(
+                serde_json::to_value(MetricsValue::new(
+                    self.executed_commands_count.borrow()[index] as usize,
+                    tags_executed.clone(),
+                ))
+                .to_indy(IndyErrorKind::IOError, "Unable to convert json")?,
+            );
```
skhoroshavin, line 82: Can we simplify this? Extract some common parts to helper function?

libindy/src/services/metrics/mod.rs
```diff
@@ -125,10 +182,58 @@ mod test {
...
+
+        let expected_commands_duration_ms = [
+            json!({"tags":{"command":"payments_command_build_set_txn_fees_req_ack","stage":"executed"},"value":0}),
+            json!({"tags":{"command":"metrics_command_collect_metrics","stage":"queued"},"value":0}),
+            json!({"tags":{"command":"cache_command_purge_cred_def_cache","stage":"executed"},"value":0}),
+            json!({"tags":{"command":"non_secrets_command_fetch_search_next_records","stage":"queued"},"value":0}),
+        ];
+
+        for command in &expected_commands_count {
+            assert!(commands_count.contains(&command));
```
skhoroshavin, line 232: I'd prefer to avoid loops in tests for a lot of reasons, but the most compelling one in this case is that when this assert fails you won't see why exactly until you rerun this under debugger.

skhoroshavin (approved):
In my opinion this looks good enough to be merged, however please address comments in subsequent PR

PR #2319, opened 2020-12-02
Async Rust in libindy

jovfer:


Patrik-Stas:
Hi @jovfer I know this is fairly old PR but this looks pretty cool, wondering whether this would pass on rerun, now that CI/CD stuff has been sorted out.

jovfer:
@Patrik-Stas Evernym has more changes atop it in the internal fork. And we are planning to publish our changes. Hopefully, it would be the same licence and in the nearest future, so it would be possible to merge libindy async from there instead of fixing this PR.
FYI @esplinr

Patrik-Stas:
@jovfer wow that would be really amazing! Honestly I think if you manage to pull off this contribution, it can substantially change the future of libindy - not only because by itself it's a great contribution, but especially because it can drive further community engagement. Imo async libindy is the main factor whether Aries frameworks stay with libindy or migrate to indy-shared-rs and whether community contributions in general continue here or move over to indy-shared-rs).

Patrik-Stas:
I did not realize just how much awesome this PR is before. 

This is not only gonna improve performance by a magnitude, but also creates an opportunity to expose async, Rust native API from libindy crate itself. If the individual methods of executors (`PoolCommandExecutor`, `WalletCommandExecutor`) are changed to `pub`, eg. method `open` in `PoolCommandExecutor`,  you could then look at the `commands` layer as native Rust API, and `libindy` could be published and consumed via crates.io.

This would in particular greatly benefit `aries-vcx` library, as we could call `libindy` directly instead of via C ABI - which would give us lot more flexibility and make maintenance a lot more easy. Definitely incentivizing us to stay with libindy.

It also gives potential for other Rust project to build on top of libindy and therefore find / incentivize new Rust developers. 

For proof of concept, I have:
- took this PR and made a few modifications (exposed more `commands` module from the crate, made some methods `pub`) https://github.com/Patrik-Stas/indy-sdk/pull/2
- Created small project trying to consume that version of libindy as crate dependency. And it works! https://github.com/Patrik-Stas/consume-indy-crate/blob/master/src/libindy_async.rs


Soo now I am really excited and want to get this merged! :-) Happy to help with this.

georgejpadayatti:
Any status on when this PR would get merged? Excited for this PR to solve the blocking nature of the wallet APIs.

jovfer:
There is no plan at Evernym to update this PR. But our own version of IndySDK is available at https://gitlab.com/evernym/verity/vdr-tools under Apache 2.0 Licenses and main branch there is actually the evolved version of this PR/branch. So if Hyperledger community is interesting in such changes there are no license concerns to bring some commits from Evernym repo

PR #2337, opened 2020-12-24
Fix for 'gpg: signing failed: Inappropriate ioctl for device' code signing error

WadeBarnes:
Signed-off-by: Wade Barnes <wade.barnes@shaw.ca>

### DIFF
--- a/ci/deb-build-and-upload.sh
+++ b/ci/deb-build-and-upload.sh
@@ -32,4 +32,9 @@ cargo deb --no-build --deb-version ${version}-${package_type} --variant ${packag
 
 mkdir debs &&  mv target/debian/*.deb ./debs/
 
+# Fix for 'gpg: signing failed: Inappropriate ioctl for device' code signing error
+# ToDo:
+#  - If this fixes the issue, the logic should be implemented in the sovrin-packaging tools and removed from here.
+export GPG_TTY=$(tty)
+
 ./sovrin-packaging/upload_debs.py ./debs $repo $type --distro=$package_type --host $host --ssh-key $key $extra_flags


ianco (approved):

ci/deb-build-and-upload.sh
```diff
@@ -32,4 +32,9 @@ cargo deb --no-build --deb-version ${version}-${package_type} --variant ${packag
 
 mkdir debs &&  mv target/debian/*.deb ./debs/
 
+# Fix for 'gpg: signing failed: Inappropriate ioctl for device' code signing error
+# ToDo:
+#  - If this fixes the issue, the logic should be implemented in the sovrin-packaging tools and removed from here.
+export GPG_TTY=$(tty)
```
skhoroshavin, line 38: This is going to be executed on Jenkins machine, however problem happens on `repo.sovrin.org` machine. So if you want to try this fix you're basically forced to do that in sovrin-packaging tools.

WadeBarnes: That script and the python scripts it calls that do the code signing should all be running in the same context on the same container.

skhoroshavin: Python script it calls indeed runs in the same context, but that python script actually does ssh into `repo.sovrin.org` and calls gpg here, and that's different context.

skhoroshavin: Please see:
https://github.com/sovrin-foundation/sovrin-packaging/blob/master/sovrin_repo.py#L325
https://github.com/sovrin-foundation/sovrin-packaging/blob/master/sovrin_repo.py#L118
https://github.com/sovrin-foundation/sovrin-packaging/blob/master/sovrin_repo.py#L302

WadeBarnes: Seeing as the configuration on `repo.sovrin.org` should not have changed in a very long time, I'm curious as to why the code signing is not working anymore.  Where would you suggest we start looking and implement the fix?

WadeBarnes: Perhaps where the remote environment is setup; https://github.com/sovrin-foundation/sovrin-packaging/blob/master/sovrin_repo.py#L325

WadeBarnes: Looking more closely at the configuration on `repo.sovrin.org` the `.bashrc` file is setting up the environment.
```
####GPG Agent setup ####
GPG_TTY=$(tty)
export GPG_TTY
```
So I guess this is more of a question of why that's not working anymore.

WadeBarnes:
The problem is at a different level.  We'll need to dig deeper into the issue and why the existing configurations are not being picked up.

PR #2338, opened 2020-12-29
Bucket metrics for Libindy

adenishchenko:
Signed-off-by: Anton Denishchenko <anton.denishchenko@dsr-corporation.com>


libindy/src/services/metrics/mod.rs
```diff
@@ -12,32 +12,24 @@ pub mod models;
...
 impl MetricsService {
     pub fn new() -> Self {
         MetricsService {
-            queued_commands_count: RefCell::new([u128::MIN; COMMANDS_COUNT]),
-            queued_commands_duration_ms: RefCell::new([u128::MIN; COMMANDS_COUNT]),
-
-            executed_commands_count: RefCell::new([u128::MIN; COMMANDS_COUNT]),
-            executed_commands_duration_ms: RefCell::new([u128::MIN; COMMANDS_COUNT]),
+            queued_counters: RefCell::new([CommandCounters::new(0,0,[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]); COMMANDS_COUNT]),
+            executed_counters: RefCell::new([CommandCounters::new(0,0,[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]); COMMANDS_COUNT]),
```
Toktar, line 23: I suggest replacing `[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]` with `[0; BUCKET_COUNT]` for both counters.

skhoroshavin: I think I have a better suggestion - remove parameters from `CommandCounters::new` method at all

libindy/src/services/metrics/models.rs
```diff
@@ -12,3 +15,32 @@ impl MetricsValue {
...
+#[derive(Serialize, Deserialize, Copy, Clone, Debug)]
+pub struct CommandCounters {
+    pub count: u128,
+    pub duration_ms_sum: u128,
+    pub duration_ms_bucket: [u128; BUCKET_COUNT],
+}
+
+impl CommandCounters {
+    pub fn new(count: u128, duration_ms_sum: u128, duration_ms_bucket: [u128; BUCKET_COUNT]) -> Self {
+        CommandCounters {count, duration_ms_sum, duration_ms_bucket}
```
skhoroshavin, line 28: I don't think we need parameters in this method, it should be okay to always initialize with zeros

libindy/src/services/metrics/mod.rs
```diff
@@ -127,10 +118,10 @@ mod test {
...
         let metrics_service = MetricsService::new();
         for index in (0..MetricsService::commands_count()).rev() {
-            assert_eq!(metrics_service.queued_commands_count.borrow()[index as usize], 0);
-            assert_eq!(metrics_service.queued_commands_duration_ms.borrow()[index as usize], 0);
-            assert_eq!(metrics_service.executed_commands_count.borrow()[index as usize], 0);
-            assert_eq!(metrics_service.executed_commands_duration_ms.borrow()[index as usize], 0);
+            assert_eq!(metrics_service.queued_counters.borrow()[index as usize].count, 0);
+            assert_eq!(metrics_service.queued_counters.borrow()[index as usize].duration_ms_sum, 0);
+            assert_eq!(metrics_service.executed_counters.borrow()[index as usize].count, 0);
+            assert_eq!(metrics_service.executed_counters.borrow()[index as usize].duration_ms_sum, 0);
```
skhoroshavin, line 124: I would suggest to create a unit test suite for MetricCounter class, so that you don't need to do all those repetitions here. And, btw, in your test you're missing check that buckets are also initialized as zeros.

libindy/tests/metrics.rs
```diff
@@ -16,50 +16,57 @@ mod collect {
...
             json!({"tags":{"label":"opened_wallets_count"},"value":0}),
             json!({"tags":{"label":"opened_wallet_ids_count"},"value":0}),
             json!({"tags":{"label":"pending_for_import_wallets_count"},"value":0}),
             json!({"tags":{"label":"pending_for_open_wallets_count"},"value":0}),
         ];
 
-        for command in &expected_threadpool_threads_count {
-            assert!(threadpool_threads_count.contains(&command));
-        }
+        assert!(wallet_count.contains(&expected_wallet_count[0]));
```
skhoroshavin, line 38: I think removing level of indirection and writing just `assert!(wallet_count.contains( json!({"tags":{"label":"opened"},"value":0}) );` would be much more readable :) Even better - you could restructure your tests into following separate test cases:
* check generic schema (that metrics are a dictionary of arrays of dictionaries, and those inner dictionaries always have `value` and `tags` fields)
* check that you have all expected top-level metrics (`wallet_count`, `threadpool_thread_count`, `commands_count`, `commands_duration_ms`, `commands_duration_ms_bucket`, etc)
* for each top-level metric check that you have test that it has all expected submetrics with corresponding tags (and I don't think you really need to check values at this level)

Also you it looks like there is a tautology - you already have metric called `wallet_count`, so I suggest to reduce labels just to `opened`, `opened_ids`, `pending_for_import`, `pending_for_open`.

libindy/src/commands/metrics.rs
```diff
@@ -66,31 +66,30 @@ impl MetricsCommandExecutor {
...
-        let mut threadpool_threads_count = Vec::new();
-        let mut tag = HashMap::<String, String>::new();
+        let mut threadpool_threads_count: Vec<Value> = Vec::new();
 
-        tag.insert(String::from("label"), String::from(THREADPOOL_ACTIVE_COUNT));
-        threadpool_threads_count.push(
-            serde_json::to_value(MetricsValue::new(tp_instance.active_count(), tag.clone()))
-                .to_indy(IndyErrorKind::IOError, "Unable to convert json")?,
+        threadpool_threads_count.push( self.get_metric_json(
+            String::from(THREADPOOL_ACTIVE_COUNT),
```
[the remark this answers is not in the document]
adenishchenko: It's nice idea. I use `&str` instead `&'static str`. As far as I know `'static` is bad practice for Rust.
Do I really need to use `&'static str` to this place?

skhoroshavin (approved):

libindy/src/commands/metrics.rs
```diff
@@ -154,4 +140,13 @@ impl MetricsCommandExecutor {
 
         Ok(())
     }
-}
+
+    fn get_metric_json(&self, label: String, value: usize) -> IndyResult<Value> {
+        let mut tag = HashMap::<String, String>::new();
```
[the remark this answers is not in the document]
adenishchenko: Sorry, but I faced problems with lifetime in MetricsValue structure. I stayed `&str` for now.

skhoroshavin (approved):

PR #2342, opened 2020-12-30
remove crypto/rust_base64.rs

AxelNennker:

This PR removes crypto/rust_base64.rs and uses the base64 directly. It also harmonizes the versions used in different modules to 0.13.0.


jovfer:
Thanks Axel, for the proposed improvements. I'm in doubt to choose between current code and your option. On one hand the proposed changes simplify logic and reduce amount of code. On the other hand current abstraction simplify usage a bit.
So I'm neutral to this PR. From technical / functional perspective it's correct and can be merged safely as far as I can see.

@mirgee @Patrik-Stas @skhoroshavin @kdenhartog @Artemkaaas thoughts?

kdenhartog (approved):
Seems to simplify the code a bit. The main advantage I see here is the encapsulation of the redundant unsafe blocks to the base64_decode_ptr(). To me that's puts me slightly in favor of the refactored code. Additionally, I do like that we're now clarifying in the variables in decode_config that we're using no padding where as before it was presumed (or incorrect according to the RFC) based on the var names.

Issue #2363, opened 2021-03-03
Postgres storage plugin defaults to the wrong DB on postgres>=12


vongohren:
After a conversation here: https://chat.hyperledger.org/channel/aries-cloudagent-python?msg=JgzXEfAFPQEAc6CKJ

This issue is created.
Following are some screenshots and the explanation
<img width="842" alt="Screen Shot 2021-03-01 at 14 51 05" src="https://user-images.githubusercontent.com/1012055/109822434-3a69f480-7c37-11eb-98f1-cf8f3fa5fe36.png">

<img width="1680" alt="Screen Shot 2021-03-01 at 14 50 49" src="https://user-images.githubusercontent.com/1012055/109822385-2faf5f80-7c37-11eb-89bd-d1fbf15c0e40.png">

Through the aca-py, and following the docs, it seems to be an issue with the indy SDK connecting to the DB when postgres version is atleast equal or bigger than 12.

Config used
```
aca-py start \
-it http '0.0.0.0' 10000 \
-ot http \
--admin '0.0.0.0' 5000 \
-e 'http://localhost:10000/' \
--auto-provision \
--wallet-storage-type postgres_storage \
--wallet-storage-config '{\"url\":\"host.docker.internal:5432\"}' \
--wallet-storage-creds '{\"account\":\"local-cloud-agent\",\"password\":\"testtest\",\"admin_account\":\"admin-cloud-agent\",\"admin_password\":\"testtest\"}' \
--wallet-type indy \
--wallet-name diwala2 \
--wallet-key Diwala.Agent421311 \
--seed diwala9jphVquhjphuVIerbf0XEc74WL \
--genesis-url http://test.bcovrin.vonx.io/genesis \
--label 'Diwala Agent 3' \
--admin-insecure-mode \
--multitenant \
--multitenant-admin \
--jwt-secret very_secret_secret \
--log-level info",
```

Based on the screenshots, it seems to be connecting and trying to make a DB with the name of the admin account.

Which leads to this strange workaround I found which is unclear to me why.

This SDK builds this connection string: postgresql://admin-cloud-agent:testtest@localhost:5432
(ignore the localhost, assume it works)

[...]

PR #2366, opened 2021-03-17
connect to 'postgres' instead of a database whose name is admin's name

toshirin33:
solve the problem #2363

When an admin user connects to a database to create a database for wallets, the database is not given. Then, the database name is the same as the admin user; see https://www.postgresql.org/docs/current/libpq-connect.html 3.11.2. It is required to create a database whose name is the same as the admin user.

By using the database 'postgres' instead of admin user name's database, we can create a database for wallets without creating the database whose name is the same as the admin user.


ianco (approved):

toshirin33:
(ci) test this please

ianco:
@toshirin33  can you check the conflict?  (Probably caused when I merged your other PR)

toshirin33:
@ianco I am sorry that the #2366 contains the following line. I noticed that there is the variable _POSTGRES_DB before issuing this pull request and rewrote, but forgot to drop it in #2368 . 

https://github.com/hyperledger/indy-sdk/blob/dbd89cf94a73e7a62611c4150a874c38b810ff8d/experimental/plugins/postgres_storage/src/postgres_storage.rs#L1048

PR #2402, opened 2021-06-22
Update Rusqlite to version 0.25.3

mirgee:
The current version of `rusqlite` (0.20.0) is transitively dependent on `linked-hash-map v0.5.3`, which contains a bug ([fixed](https://github.com/contain-rs/linked-hash-map/pull/106) in version `0.5.4`). This bug causes runtime panics due to uninitialized reference on wallet access since rust `1.48.0`. See [here](https://github.com/mehcode/config-rs/issues/158), [here](https://github.com/hyperledger/aries-vcx/pull/126), and [here](https://github.com/hyperledger/indy-sdk/pull/2311).

This PR updates version of rusqlite which fixes the bug (replicating the changes in #2311 with green DCO), as the latest version of rusqlite removes dependency on `linked-hash-map`.

Signed-off-by: Miroslav Kovar <miroslavkovar@protonmail.com>

### DIFF
--- a/libindy/indy-wallet/src/storage/default/transaction.rs
+++ b/libindy/indy-wallet/src/storage/default/transaction.rs
@@ -14,6 +14,7 @@ impl<'conn> Transaction<'conn> {
             TransactionBehavior::Deferred => "BEGIN DEFERRED",
             TransactionBehavior::Immediate => "BEGIN IMMEDIATE",
             TransactionBehavior::Exclusive => "BEGIN EXCLUSIVE",
+            _ => ""
         };
         conn.execute_batch(query)
             .map(move |_| {
@@ -75,8 +76,8 @@ impl<'conn> Transaction<'conn> {
             DropBehavior::Commit => self.commit_(),
             DropBehavior::Rollback => self.rollback_(),
             DropBehavior::Ignore => Ok(()),
-            DropBehavior::Panic => {
-                panic!("drop behaviour is set to panic".to_string());
+            _ => {
+                panic!("internal error: unsupported drop behaviour");
             }
         }
     }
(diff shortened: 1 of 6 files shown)


jovfer (approved):

Patrik-Stas (approved):

libindy/indy-wallet/src/storage/default/transaction.rs
```diff
@@ -14,6 +14,7 @@ impl<'conn> Transaction<'conn> {
             TransactionBehavior::Deferred => "BEGIN DEFERRED",
             TransactionBehavior::Immediate => "BEGIN IMMEDIATE",
             TransactionBehavior::Exclusive => "BEGIN EXCLUSIVE",
+            _ => ""
```
Patrik-Stas, line 17: There's no other variants, this should be removed, looks dangerous in case the struct would change in the future, and this match branch would then cover unintended scenarios

libindy/indy-wallet/src/storage/default/transaction.rs
```diff
@@ -75,8 +76,8 @@ impl<'conn> Transaction<'conn> {
             DropBehavior::Commit => self.commit_(),
             DropBehavior::Rollback => self.rollback_(),
             DropBehavior::Ignore => Ok(()),
-            DropBehavior::Panic => {
-                panic!("drop behaviour is set to panic".to_string());
+            _ => {
```
Patrik-Stas, line 79: Should be kept original, the original match covers all scenarios, `_` might be dangerous in case of future updates

Patrik-Stas (changes requested):
Small changes requested

Patrik-Stas (approved):

PR #2432, opened 2021-10-07
fix ordering of attributes being used before being introduced

AxelNennker:

The compiler was complaining about the following:

> warning: derive helper attribute is used before it is introduced
   --> src/services/pool/types.rs:357:3
    |
357 | #[serde(tag = "op")]
    |   ^^^^^
358 | #[derive(Serialize, Deserialize, Debug)]
    |          --------- the attribute is introduced here
    |
    = warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!
    = note: for more information, see issue #79202 <https://github.com/rust-lang/rust/issues/79202>

This PR reorders the two code lines so that the attribute is used after being introduced

### DIFF
--- a/libindy/src/commands/non_secrets.rs
+++ b/libindy/src/commands/non_secrets.rs
@@ -312,8 +312,8 @@ impl NonSecretsCommandExecutor {
     }
 }
 
-#[serde(rename_all = "camelCase")]
 #[derive(Debug, Deserialize, Serialize)]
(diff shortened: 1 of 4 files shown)


WadeBarnes:
@AxelNennker, you are correct, it is not.  A fix for the build issues has been implemented (#2470), which should have also addressed the changes you made here.  Please confirm and close the PR is this is the case.

PR #2442, opened 2021-11-10
Add a podspec for objective-c indy wrapper in non-binary distribution

conanoc:
Signed-off-by: conanoc <conanoc@gmail.com>

Previous podspec libindy-objc.podspec has several issues:
- It is based on pre-built binary and the binaries are not provided since v1.8.2
- Old binaries do not support newer versions of swift
- Binary frameworks cannot support multiple versions of swift

Source based podspec would be more convenient because we do not have to prepare binaries and there is no version dependency for swift.
Some notes about this commit:
- The podspec filename have to be the same as the name of the framework. So, I could not keep the previous name libindy-objc.podspec.
- The lastest version of libindy is v1.16.0, but one source file of the wrapper(wrappers/ios/libindy-pod/Indy/Wrapper/IndyErrors.h) had a [bug](https://github.com/hyperledger/indy-sdk/commit/9cd1d56b6209a7152f3cd983a0d6077430f09deb#diff-a29363c28be6b462a2bb8b4797fc665cbf01c89b7ad7e58dedaaa18e4f365a36)(comma missing) and was fixed later. So, tag v1.16.0 cannot be used for ios wrapper.
- Used OpenSSL-XM(1.0.210.1) pod instead of OpenSSL(1.0.210) pod because OpenSSL is not compatible with the latest version of cocoapods. OpenSSL podspec would be better copied to indy-sdk repo with fixes done in OpenSSL-XM.

### DIFF
--- a/Specs/Indy/1.15.0/Indy.podspec
+++ b/Specs/Indy/1.15.0/Indy.podspec
@@ -1,0 +1,26 @@
+Pod::Spec.new do |spec|
+  spec.name     = 'Indy'
+  spec.version  = '1.15.0'
+  spec.license  =  'Apache License 2.0'
+  spec.homepage = 'https://github.com/hyperledger/indy-sdk'
+  spec.authors  = {'Hyperledger Indy Contributors' => 'hyperledger-indy@lists.hyperledger.org'}
+  spec.summary  = 'libindy objective-C wrapper'
+  spec.source   = {:git => 'https://github.com/hyperledger/indy-sdk.git', :tag => 'v1.15.0'}
+  spec.source_files = 'wrappers/ios/libindy-pod/Indy/**/*'
+  spec.exclude_files = [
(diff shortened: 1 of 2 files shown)


WadeBarnes:
A fix for the build issues has been implemented (https://github.com/hyperledger/indy-sdk/pull/2470).  Please rebase this PR on the latest from the main branch so it is no longer blocked by the related build issues.

conanoc:
@WadeBarnes After the rebase, `build-image-wrapper-base` is failing. Any idea?

WadeBarnes:
> @WadeBarnes After the rebase, `build-image-wrapper-base` is failing. Any idea?

@conanoc, I missed updating the Rust version in a spot; https://github.com/hyperledger/indy-sdk/pull/2475

WadeBarnes:
@mirgee, Is this something you could review?

PR #2451, opened 2021-11-24
Added implementation of MultiWalletMultiTable postgres plugin strategy

JoHenrich:
We saw that the strategy MultiWalletMultiTable was prepared in the code but not yet implemented.

So decided to do that because we experienced some major peformance issues with the implemented strategies so far.

In addition we added another Strategy: MultiWalletSplitDatabaseMultiTable which cluster the tables in different databases depending on the id of the wallet. That lead to a performance boost.

All unit test ran successfull and we tested the strategy in production aswell but we still consider it as experimentall because further tests are required to ensure that there are no undetected misconceptions.

Signed-off-by: j.henrich@esatus.com


ianco:
Hi @JoHenrich  thanks for the PR!  Note that you still have one commit that is missing the DCO signoff

libindy/src/api/anoncreds.rs
```diff
@@ -935,6 +935,7 @@ pub extern fn indy_prover_create_credential_req(command_handle: CommandHandle,
                                                 cred_offer_json: *const c_char,
                                                 cred_def_json: *const c_char,
                                                 master_secret_id: *const c_char,
+                                                device_key: *const c_char,
```
ianco, line 938: Is this change related to the new postgres wallet storage scheme?

If not it should be moved to a separate PR

ianco (changes requested):
Overall looks good but I think the anoncreds changes should be in a separate PR.

JoHenrich:
> Hi @JoHenrich thanks for the PR! Note that you still have one commit that is missing the DCO signoff

Hey, thanks for the fast reply. Yes i saw it too, i don`t manage to sign it off. I have another branch where everything is verified. Should i open a new PR and close this one?

libindy/src/api/anoncreds.rs
JoHenrich: We will remove the anoncreds changes. They are not related to that PR. Thanks for the hint

JoHenrich:
Closing this PR because there is another one pending with all commits signed off

Issue #2468, opened 2022-01-18
Allow java samples to run individually & polish sample docs


tdiesler:

tdiesler:
PR https://github.com/hyperledger/indy-sdk/pull/2469

tdiesler:
Merged

PR #2469, opened 2022-01-18
[#2468] Allow java samples to run individually & polish sample docs

tdiesler:
signed-off-by: Thomas Diesler <tdiesler@redhat.com>

### DIFF
--- a/samples/java/src/main/java/Endorser.java
+++ b/samples/java/src/main/java/Endorser.java
@@ -8,8 +8,6 @@ import org.json.JSONArray;
 import org.json.JSONObject;
 import utils.PoolUtils;
 
-import java.util.Arrays;
-
 import static org.hyperledger.indy.sdk.ledger.Ledger.buildNymRequest;
 import static org.hyperledger.indy.sdk.ledger.Ledger.buildSchemaRequest;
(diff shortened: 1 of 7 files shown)


samples/java/src/main/java/Endorser.java
```diff
@@ -85,7 +88,7 @@ static void demo() throws Exception {
         String schemaAttributes = new JSONArray().put("name").put("age").put("sex").put("height").toString();
         AnoncredsResults.IssuerCreateSchemaResult createSchemaResult =
                 issuerCreateSchema(authorDid, schemaName, schemaVersion, schemaAttributes).get();
-        String schemaId = createSchemaResult.getSchemaId();
+        // String schemaId = createSchemaResult.getSchemaId();
```
WadeBarnes, line 91: Would it be better to remove this variable since it's not used?

WadeBarnes (changes requested):

samples/java/src/main/java/Endorser.java
tdiesler: Done

WadeBarnes (approved):

PR #2505, opened 2022-05-27
allow issuance of multiple cred at same revocation index

Zzocker:
This PR proposes an option for the issuer to specify a cred_rev_id (index in revocation registry) when a credential is issued.This gives more control for an issuer to manage revocation registry, a few examples can be

- Alice issues credentials with same cred_rev_id multiple times to Bob, who wants the same credential to be stored in multiple wallets, In this case Alice can revoke a credential of Bob by revoking a single index, this allows Alice to create a better controller logic for revocation registry
- Alice issues credentials with same cred_rev_id to multiple Holders e.g. Coupons

@DaevMithran

Signed-off-by: Pritam Singh <pkspritam16@gmail.com>

### DIFF
--- a/libindy/include/indy_anoncreds.h
+++ b/libindy/include/indy_anoncreds.h
@@ -80,6 +80,7 @@ extern "C" {
                                                       const char *  cred_req_json,
                                                       const char *  cred_values_json,
                                                       const char *  rev_reg_id,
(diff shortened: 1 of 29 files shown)


DaevMithran:
@andrewwhitehead can you review this ?

